PR #324, opened 2014-11-10
Rework ui resources

hgschmie:
This moves the UI resource away from the root ("/") of the service onto /ui. It also adds a special resource that can be enabled or disabled to restore the old behavior to open the UI when accessing the root of the server.

If the resource is enabled, the root can be replaced with another resource (e.g. custom branding or some other endpoint) while the Singularity UI is still available as /ui.


hgschmie:
reworked to also work with the API docs. This is pretty much indispensable for us to run Singularity as we have to run with `applicationContextPath: "/"` so that we can hook additional resources into the root namespace.

hgschmie:
rebased

hgschmie:
rebased

hgschmie:
added suggestions from @tpetr .

tpetr:
thanks!

PR #328, opened 2014-11-12
Add metrics for all JAX-RS resource endpoints.

hgschmie:
This exposes metrics to the internal metrics registry.


hgschmie:
So I am planning to work on a decorator for that so I will close the pull request, otherwise I will lose track on all the pull requests that I have open.

PR #360, opened 2014-12-09
Change MesosResourcesObject to be resource agnostic.

hgschmie:
Replace hardcoded resources with generic code.


hgschmie:
Yes, prep.

PR #368, opened 2014-12-12
Factor out all the little thread pools.

hgschmie:
Factor thread pool management and control into guice injected, managed pools.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
@@ -135,6 +135,8 @@ public class SingularityConfiguration extends Configuration {
 
   private long zookeeperAsyncTimeout = 5000;
 
+  private int coreThreadpoolSize = 8;
+
   private long threadpoolShutdownDelayInSeconds = 1;
 
   private int listenerThreadpoolSize = 3;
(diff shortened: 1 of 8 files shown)


SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -135,6 +135,10 @@
 
   private long zookeeperAsyncTimeout = 5000;
 
+  private int coreThreadpoolSize = 8;
```
hgschmie: Sharing them allows to factor the thread pool management out of a lot of components which in turn no longer need to be managed (only managed thing was the threadpool shutdown). The "8" is a magic number based on the number of thread pools in existence. We have a metric which tracks the number of threads actually used and I may adjust later.

wsorenson: Could we just have some sort of thread pool provider that is managed?

hgschmie: You mean like a SingularityManagedScheduledExecutorServiceProvider? :-)

wsorenson: I was thinking we should have it be a Provider like a Factory that creates a new Managed Executor Service. That way we don't have to hard code the # of threads, and it also guarantees that each user can not mistakenly over-use threads in the pool and starve another process.

wsorenson:
Addressed in #417

PR #371, opened 2014-12-22
Support all deploy fields on docker deploy form

ssalinas:
@tpetr

### DIFF
--- a/SingularityUI/app/templates/dockerForms/portMap.hbs
+++ b/SingularityUI/app/templates/dockerForms/portMap.hbs
@@ -1,0 +1,35 @@
+<div class="well well-sm docker-port">
+    <h5>Docker Port Mapping</h5>
+    <div class="remove-button"></div>
+
+    <div class="form-group required">
+        <label for="cont-port-type-{{ timestamp }}">Container Port Type</label>
+        <select id="cont-port-type-{{ timestamp }}" class='cont-port-type form-control'>
+            <option value="LITERAL">Literal</option>
+            <option value="FROM_OFFER">From Offer</option>
+        </select>
+    </div>
+
+    <div class="form-group required">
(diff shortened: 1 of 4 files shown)


tpetr:
tested this out in vagrant and it doesn't seem as though the port mappings were persisted into the JSON

ssalinas:
@tpetr I had the portMappings outside docker in the json by accident, fixed now and confirmed that the json persisted on the response

tpetr:
some progress -- now `portMappings` is being set, seemingly only the last item when I add more than one in the UI

ssalinas:
should be good now with ea995d0

SingularityUI/app/templates/dockerForms/portMap.hbs
```diff
@@ -0,0 +1,35 @@
...
+    </div>
+
+    <div class="form-group required">
+        <label for="host-port-{{ timestamp }}">Container Port</label>
+        <input type="text" id="host-port-{{ tiemstamp }}" class="host-port form-control">
+    </div>
+
+    <div class="form-group">
+        <label for="protocal-{{ timestamp }}">Protocol</label>
+        <input type="text" id="protocol-{{ tiemstamp }}" class="protocol form-control" placeholder="default: tcp">
```
tpetr: `tiemstamp` looks like a typo

ssalinas:
fixed my typo

tpetr:
i think this needs a little more work -- i created a new deploy and added 2 port mappings, but when i inspected the POST, it only included the last port mapping

ssalinas:
That should have been fixed with ea995d0 , it works when I run it locally and I can't seem to reproduce. Certain type of request maybe?

tpetr:
sorry, was testing this on the wrong branch! LGTM

PR #374, opened 2014-12-24
ui changes for decommission states

ssalinas:
@tpetr @wsorenson here is a first shot at the ui changes to go along with the changes in decommission states


ssalinas:
There will overall be less, but more uniform, data for the three sections of slaves/racks. The rack/slave history data we could add to a separate more detailed page about the individual rack/slave.

- Before racks would have columns like these (with some extra data for decommissioning/dead):
screen shot 2015-01-05 at 9 08 15 am

- Now it will look like this (for all types, not just active):
screen shot 2015-01-05 at 9 07 21 am


- The slaves overview before was:
screen shot 2015-01-05 at 9 07 52 am
screen shot 2015-01-05 at 9 07 58 am
screen shot 2015-01-05 at 9 08 04 am

- Now, for all types it will just be:
screen shot 2015-01-05 at 9 07 28 am

wsorenson:
- Should we maintain separate sections? 
- Shouldn't the triggered by still be applicable in some cases? (There is an Optional<User> returned for any current state)
- I've introduced a new endpoint for activating a decomissioned slave that will cause it to go active without deleting its history. We should add a way to do this in the UI.

tpetr:
+1 to whitney -- we need the sections because these are concrete states that the slaves can transition between.

we can add an additional icon on the right side of the slave row for re-activating a slave

ssalinas:
I've added the triggered by back in for decommissioning slaves, I'll add another icon for decommissioned ones to re-activate as well

wsorenson:
@ssalinas "triggered by" can now apply to any state (except lost really.) it may need a rename. in the case of an active or completely decomissioned slave, it is the user that activated or started the decomission.

ssalinas:
@wsorenson I've updated it to read as 'Decommissioned By' for the decommission states, and 'Activated By' for the active states, not showing user for Dead/Missing

tpetr:
UI looks okay, but im not sure the logic behind it is 100% right -- when I try to decomission a slave (testing inside our vagrant box), a DELETE request is issued to `/singularity/api/slave/SLAVE_ID?user=tpetr` which is in correct. it should be `POST /singularity/api/slave/SLAVE_ID/decommission?user=tpetr` instead

ssalinas:
Ok, I've fixed the UI to have separate remove and decommission buttons (that was my misunderstanding I thought both were unified now). I also updated the endpoints for /decomission (typo in the endpoint is correct, filed an issue for that)

wsorenson:
Can you rebase this (assuming this is why the build is broken)? Assuming it's done, I'd like to merge it into my PR.

tpetr:
i'm still seeing some UI issues -- double checking after rebase

tpetr:
re-activating a decomissioning slave doesnt seem to work in the UI (the X button does nothing)

ssalinas:
@tpetr I believe I've fixed the activate piece

tpetr:
more open issues:
- both slave decomission / delete buttons bring up the decomission dialog
- i still don't seem to be able to re-activate a decomissioning slave (clicking the X at the right):
screen shot 2015-01-09 at 11 00 20 am

ssalinas:
I've made a few more changes, there were some places where we mistakenly spelled decommission with only 1 m (coming from the api), so the ui will check for both spellings. The appropriate buttons all show up now and the dialogs have been corrected

PR #377, opened 2015-01-05
fix logfetch date range filter for active task logs

ssalinas:
@tpetr this fixes an issue some people were seeing with filtering by date. Now it checks if each of the log files associated with an active task is in the date range as opposed to just checking if the task itself is in that range.

### DIFF
--- a/scripts/logfetch/live_logs.py
+++ b/scripts/logfetch/live_logs.py
@@ -80,7 +80,7 @@ def all_tasks_for_request(args):
   elif len(active_tasks) == 0:
     return historical_tasks
   else:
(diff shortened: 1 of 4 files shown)


scripts/logfetch/live_logs.py
```diff
@@ -80,10 +80,10 @@ def all_tasks_for_request(args):
   elif len(active_tasks) == 0:
     return historical_tasks
   else:
-    return active_tasks + historical_tasks
+      return active_tasks + [h for h in historical_tasks if is_in_date_range(args, int(str(h['updatedAt'])[0:-3]))]
```
tpetr: looks like an indentation issue here

tpetr:
LGTM aside from small nitpick

ssalinas:
@tpetr fixed with one other small item as well, going to merge this

PR #379, opened 2015-01-06
hide icon for run-now when daemon task

ssalinas:
@tpetr fixes the error that some users were getting back when hitting the lightning bolt on a daemon task, now it won't be there

### DIFF
--- a/SingularityUI/app/views/simpleSubview.coffee
+++ b/SingularityUI/app/views/simpleSubview.coffee
@@ -15,7 +15,8 @@ class SimpleSubview extends View
         _.extend super,
             'click [data-action="expandToggle"]': 'expandToggle'
 
-    initialize: ({@template}) ->
+    initialize: (@params) ->
+        { @template } = @params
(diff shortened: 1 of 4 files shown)


SingularityUI/app/views/simpleSubview.coffee
```diff
@@ -20,18 +20,19 @@ class SimpleSubview extends View
...
         @$el.html @template
-            config:   config
-            data:     @data.toJSON()
-            synced:   @data.synced
-            expanded: @expanded
+            config:    config
+            data:      @data.toJSON()
+            synced:    @data.synced
+            expanded:  @expanded
+            modelData: @model?.toJSON()
```
tpetr: it's probably fine, but are we cool with this extra `toJSON()` call for every use of `simpleSubview`?

ssalinas:
@tpetr here is a slightly cleaner solution, thanks @timmfin!

tpetr:
LGTM, thanks

PR #380, opened 2015-01-06
Quartz dialog

ssalinas:
Adds an icon to the requests table when the Schedule column is present:
screen shot 2015-01-06 at 11 42 46 am

clicking the icon produces a vex dialog with info about quartz schedule format:
screen shot 2015-01-06 at 11 43 00 am

Let me know if the dialog is too verbose, maybe we just want the top half with the simple explanation? Or maybe the bottom should have special character explanations instead of the current content?

@tpetr


ssalinas:
putting this to hs_staging instead

PR #382, opened 2015-01-07
add indicator for overdue tasks

ssalinas:
This will add a small badge to tasks that are overdue:

- On the request detail page
screen shot 2015-01-07 at 11 16 22 am

- On the task table page
screen shot 2015-01-07 at 11 16 31 am

@tpetr

### DIFF
--- a/SingularityUI/app/handlebarsHelpers.coffee
+++ b/SingularityUI/app/handlebarsHelpers.coffee
@@ -36,6 +36,15 @@ Handlebars.registerHelper 'timestampFromNow', (timestamp) ->
     timeObject = moment timestamp
     "#{timeObject.fromNow()} (#{ timeObject.format 'lll'})"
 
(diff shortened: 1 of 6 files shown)


tpetr:
overall this looks good, i just have some small nitpicks with the visual style -- i think there's too much padding inside the badge. could we tweak this to make it more look like the `DANGER` label in [bootstrap](http://getbootstrap.com/components/#labels), and maybe remove the `!`?

screen shot 2015-01-07 at 11 21 49 am

@timmfin feel free to lay on your special UI sauce if you have a minute

Eseb:
I agree with @tpetr, I feel that using a button is the wrong choice for this sort of UI element. The badges are much more appropriate, and maybe put it right next to the timestamp rather than having it as its own column.

ssalinas:
@tpetr @Eseb update the badge to look like this using label:
screen shot 2015-01-07 at 2 03 33 pm

tpetr:
megusta

PR #383, opened 2015-01-07
add log tailing to logfetch tool

ssalinas:
@tpetr @wsorenson 
This will add the ability to tail logs for a request, task, or request/deploy when the --tail option is specified with a logfile path (ie. 'logs/access.log'). It will also take the -g grep option and pas that through to the api as well.

If --tail isn't specified the tool will download and grep as before.


scripts/logfetch/log_fetcher.py
```diff
@@ -70,7 +74,8 @@ def entrypoint():
   parser.add_argument("-u", "--singularity-uri-base", help="The base for singularity (eg. http://localhost:8080/singularity/v1)", metavar="URI")
   parser.add_argument("-s", "--start-days", help="Search for logs no older than this many days", type=int, metavar="start_days")
   parser.add_argument("-e", "--end-days", help="Search for logs no new than this many days (defaults to None/today)", type=int, metavar="end_days")
-  parser.add_argument("-g", "--grep", help="Regex to grep for (normal grep syntax) or a full grep command", metavar='grep')
+  parser.add_argument("-g", "--grep", help="Regex to grep for (normal grep syntax) or a full grep command(cannot use full command with --tail)", metavar='grep')
+  parser.add_argument("--tail", help="Logfile name to tail, if this is set, no downloads will happen", metavar="tail")
```
tpetr: can we mimic the `tail` command and use `-t` as well (make people think less)?

scripts/logfetch/tail.py
```diff
@@ -0,0 +1,75 @@
...
+  try:
+    threads = []
+    for task in tasks:
+      thread = LogStreamer(args, task)
+      threads += [thread]
+      thread.start()
+    while True: # main thread needs something to do so it doesn't kill the others
+      time.sleep(1)
+  except KeyboardInterrupt:
+    sys.stdout.write('Stopping tail')
```
tpetr: use `stderr` for these informational messages

scripts/logfetch/tail.py
```diff
@@ -0,0 +1,75 @@
...
+    sys.stderr.write('{0}\n'.format(t))
+  sys.stderr.write('ctrl+c to exit\n')
+  try:
+    threads = []
+    for task in tasks:
+      thread = LogStreamer(args, task)
+      threads += [thread]
+      thread.start()
+    while True: # main thread needs something to do so it doesn't kill the others
+      time.sleep(1)
```
tpetr: better practice is to call `join()` on all threads

scripts/logfetch/tail.py
```diff
@@ -0,0 +1,75 @@
...
+    tasks = [args.taskId]
+  sys.stderr.write('Tailing logs for tasks:\n')
+  for t in tasks:
+    sys.stderr.write('{0}\n'.format(t))
+  sys.stderr.write('ctrl+c to exit\n')
+  try:
+    threads = []
+    for task in tasks:
+      thread = LogStreamer(args, task)
+      threads += [thread]
```
tpetr: this is a huge nitpick i know, but just use `.append()` instead of concatenating the lists

tpetr:
what happens then a request churns through tasks (i.e. OOM kill / misc. failure / bounce / deploy)? right now we won't pick up new tasks -- if we're cool with that, we need to make it very clear in the docs

tpetr:
i'd almost prefer to have a 2nd entrypoint (`logtail`) instead of the `--tail` arg, which completely changes the functionality of the program

ssalinas:
I've made logtail it's own entrypoint and reworked the threading a bit, we were using -t for the taskId already so for right now I made -l the logfile name to tail

scripts/logfetch/entrypoint.py
```diff
@@ -81,4 +85,47 @@ def entrypoint():
...
+  parser = argparse.ArgumentParser(parents=[conf_parser], description="Tail log files from Singularity. One can specify either a TaskId, RequestId and DeployId, or RequestId",
+          prog="log_fetcher")
+
+  parser.set_defaults(**defaults)
+  parser.add_argument("-t", "--taskId", help="TaskId of task to fetch logs for", metavar="taskId")
+  parser.add_argument("-r", "--requestId", help="RequestId of request to fetch logs for", metavar="requestId")
+  parser.add_argument("-d", "--deployId", help="DeployId of task to fetch logs for", metavar="deployId")
+  parser.add_argument("-u", "--singularity-uri-base", help="The base for singularity (eg. http://localhost:8080/singularity/v1)", metavar="URI")
+  parser.add_argument("-s", "--start-days", help="Search for logs no older than this many days", type=int, metavar="start_days")
+  parser.add_argument("-e", "--end-days", help="Search for logs no new than this many days (defaults to None/today)", type=int, metavar="end_days")
```
tpetr: start and end is unnecessary for tail, right?

scripts/logfetch/entrypoint.py
```diff
@@ -81,4 +85,47 @@ def entrypoint():
...
+  parser.add_argument("-g", "--grep", help="String to grep for", metavar='grep')
+  parser.add_argument("-l", "--logfile", help="Logfile path/name to tail (ie 'logs/access.log')", metavar="logfile")
```
tpetr: do we have a good default for this?

tpetr:
re: https://github.com/HubSpot/Singularity/pull/383#issuecomment-69087139 -- totally overlooked `-t` being for task id, right

scripts/logfetch/entrypoint.py
ssalinas: There isn't a default for it currently, right now I have it providing an error message if the log to tail isn't set

scripts/logfetch/entrypoint.py
ssalinas: it was previously used for filtering our really old historical tasks, realizing we won't need those for tailing live logs, I'll remove those and the piece that fetches historical tasks as well

ssalinas:
Tailing should be all set now, I've trimmed the arguments down to just the IDs, logfile, and grep string (along with the base uri/conf file stuff). I left the logfile arg as giving an error message if it is not set.

PR #386, opened 2015-01-08
Fix executor cleanup

tpetr:
SingularityExecutorCleanup was using a deprecated endpoint to load tasks for a specific slave by its hostname. This PR changes the code to use the endpoint that looks up tasks by slave ID.

@wsorenson

### DIFF
--- a/SingularityExecutorCleanup/src/main/java/com/hubspot/singularity/executor/cleanup/SingularityExecutorCleanup.java
+++ b/SingularityExecutorCleanup/src/main/java/com/hubspot/singularity/executor/cleanup/SingularityExecutorCleanup.java
@@ -11,15 +11,20 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
-import com.google.common.base.Throwables;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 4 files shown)


tpetr:
(tested successfully in `mesos_test_3x5` cluster at HubSpot)

SingularityExecutorCleanup/src/main/java/com/hubspot/singularity/executor/cleanup/SingularityExecutorCleanup.java
```diff
@@ -129,16 +138,23 @@ public SingularityExecutorCleanupStatistics clean() {
     return statisticsBldr.build();
   }
 
-  private Collection<SingularityTask> getActiveTasksOnSlave() {
+  private String getLocalSlaveID() {
     try {
-      return singularityClient.getActiveTasks(JavaUtils.getHostAddress());
-    } catch (SocketException e) {
-      throw Throwables.propagate(e);
+      final HttpRequest request = HttpRequest.newBuilder().setUrl(String.format(LOCAL_SLAVE_STATE_URL_FORMAT, JavaUtils.getHostAddress())).build();
```
wsorenson: shouldn't this be part of our mesos client?

PR #387, opened 2015-01-08
update dev doc to match reality

hgschmie:
Replace the webhooks with being a plugin into a general purpose event reporting facility. It currently processes three event types (task, deploy, request) and allows additional listeners to react on these events.

We (Groupon) plug an AMQP sender into this facility and send events onto a message bus topic to allow additional code to listen to what is going on without having to register webhooks and check the state of webhooks all the time.

### DIFF
--- a/Docs/development/vagrant.md
+++ b/Docs/development/vagrant.md
@@ -81,27 +81,31 @@ mysql> show databases;
 type 'exit' to exit mysql client console
 ```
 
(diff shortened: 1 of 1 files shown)


tpetr:
thanks!

tpetr:
@hgschmie just FYI looks like your PR message was a copy of your previous one

PR #388, opened 2015-01-08
Call out decommissioning tasks (not ready for merge yet)

ssalinas:
@tpetr here is step one towards a better visualization of decommissioning tasks
screen shot 2015-01-08 at 3 39 36 pm

Still need to add a filter for tasks running on a decommissioning host in the task table view


ssalinas:
This relies on HubSpot/Singularity#374 before it can be merged (much easier to get slave data after the new endpoints are in place)

This PR will call out tasks that are running on decommissioning slaves in a few places:

- on the task detail page
image

- per task on the active tasks list in the task table
screen shot 2015-01-13 at 3 08 21 pm

- Via a filter button on the tasks page which will bring up all active tasks running on a deommissioning host
screen shot 2015-01-13 at 3 35 14 pm

timmfin:
For the task detail page, I'd move the "decommissioning" label (yeesh that is long, could we abbreviate?) to the right side of the task name. It looks awkward on the left.

tpetr:
@timmfin sorry, steve and i discussed this in an internal ticket -- i raised both of your issues (great minds...) and steve addressed them in the most recent commit

tpetr:
this all looks good, but could you re-create this PR to merge into `decomission_changes `, since it depends on that? merging this PR as-is would automagically merge that PR too, and im not 100% sure it's ready for master yet

PR #389, opened 2015-01-09
add a request type column in the requests table

ssalinas:
Adds the deployable type in a 'Type' column in the requests table
screen shot 2015-01-09 at 12 16 47 pm


wsorenson:
Shouldn't this just be based on https://github.com/HubSpot/Singularity/pull/348

ssalinas:
closing in favor of #413

PR #397, opened 2015-01-14
Remove bad dependency, make build pass again.

hgschmie:


tpetr:
thanks! we're building with `mvn clean deploy` (and i usually test things locally with `mvn clean package`), and it looks like that doesn't run all the goals that travis' `mvn install` does. is there a better goal I should be using? or do i just need to manually run `dependency:analyze` too?

HiJon89:
You can change the phase of the dependency plugin to prepare-package which I think is helpful

PR #400, opened 2015-01-14
Add configuration option for changing task email log tails.

wolfd:
Looking for feedback. Still need to do all kinds of testing before I'm confident this works.


SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
```diff
@@ -160,8 +160,13 @@ private void populateTaskEmailProperties(Map<String, Object> templateProperties,
     Optional<String> directory = taskManager.getDirectory(taskId);
 
     templateProperties.put("singularityTaskLink", getSingularityTaskLink(taskId));
-    templateProperties.put("stdout", getTaskLogFile(taskId, "stdout", task, directory).or(new String[0]));
-    templateProperties.put("stderr", getTaskLogFile(taskId, "stderr", task, directory).or(new String[0]));
+
+    Map<String, String[]> logTails = Maps.newHashMap();
```
tpetr: `Map`s are unordered -- use `new LinkedHashMap<>()` in order to enforce the log file ordering here

SingularityService/src/main/resources/templates/task.jade
```diff
@@ -106,21 +106,13 @@ html
...
-            dd(style="background: #eee; display: block; margin: 0; padding: 10px 14px 0 14px; border-radius: 3px")
-              if stdout
-                pre(style="white-space: pre; margin: 0")
-                  code(style="font-family: Menlo, Monaco, Consolas, Courier, monospace")
-                    for line in stderr
-                      div(style="font-family: Menlo, Monaco, Consolas, Courier, monospace; white-space: pre; margin-bottom: 12px")= line
-              else
-                | No stderr available. The task either did not start, output anything, or the slave hosting the log files was unavailable over http.
\ No newline at end of file
+            for logTail, logName in logTails
```
tpetr: have you verified that you can do this using the `jade4j` library? (i'm honestly not sure -- the answer might be yes)

wolfd: I typoed each vs for, but this should work besides that.

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
```diff
@@ -160,8 +160,13 @@ private void populateTaskEmailProperties(Map<String, Object> templateProperties,
...
+    for (String file : maybeSmtpConfiguration.get().getTaskEmailTailFiles()) {
+      logTails.put(file, getTaskLogFile(taskId, file, task, directory).or(new String[0]));
```
tpetr: one thing i remembered -- in the case where people are using the SingularityExecutor, `service.log` lives inside a folder named after the task ID. could we do a replacement on the `file` variable such that `$MESOS_TASK_ID` becomes the value of `taskId` ?

SingularityService/src/main/resources/templates/task.jade
tpetr: i think you got the key, value ordering flipped here

wolfd: I would think so too, but Jade flips it to maintain conformity an each with a single arg.
http://jade-lang.com/reference/iteration/

wolfd:
I just want to set up SingularityExecutor locally and see if I can successfully tail service.log before this goes in.

SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
```diff
@@ -309,4 +311,11 @@ private static Ranges subtractRanges(Ranges ranges, Ranges toSubtract) {
     return remaining;
   }
 
+  public static Path getTaskDirectoryPath(String taskId) {
+    return Paths.get(getSafeTaskIdForDirectory(taskId)).toAbsolutePath();
```
tpetr: Not sure if we can get away with this implementation as a static method since it's depending on the current directory here -- might make more sense to have this take 2 args (the base path and then the taskId)

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorTaskBuilder.java
```diff
@@ -55,7 +56,7 @@ public SingularityExecutorTaskBuilder(ObjectMapper jsonObjectMapper, JsonObjectF
   }
 
   public Logger buildTaskLogger(String taskId) {
-    Path javaExecutorLogPath = configuration.getTaskDirectoryPath(taskId).resolve(configuration.getExecutorJavaLog());
+    Path javaExecutorLogPath = MesosUtils.getTaskDirectoryPath(taskId).resolve(configuration.getExecutorJavaLog());
```
tpetr: like i said in `MesosUtils`, we shouldn't be implicitly basing all this off the current directory -- if possible, let's build up an absolute path (i know this isn't your fault but let's take a stab at improving this)

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
```diff
@@ -160,8 +162,15 @@ private void populateTaskEmailProperties(Map<String, Object> templateProperties,
     Optional<String> directory = taskManager.getDirectory(taskId);
 
     templateProperties.put("singularityTaskLink", getSingularityTaskLink(taskId));
-    templateProperties.put("stdout", getTaskLogFile(taskId, "stdout", task, directory).or(new String[0]));
-    templateProperties.put("stderr", getTaskLogFile(taskId, "stderr", task, directory).or(new String[0]));
+
+    LinkedHashMap<String, String[]> logTails = new LinkedHashMap<>();
```
tpetr: the `String[]` stuff here is pretty gross -- let's try to improve it

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
```diff
@@ -160,8 +162,15 @@ private void populateTaskEmailProperties(Map<String, Object> templateProperties,
...
+    for (String file : maybeSmtpConfiguration.get().getTaskEmailTailFiles()) {
+      // To enable support for tailing the service.log file, replace instances of $MESOS_TASK_ID.
+      file = file.replaceAll("\\$MESOS_TASK_ID", MesosUtils.getTaskDirectoryPath(taskId.getId()).toString());
```
tpetr: i think we just want `getSafeTaskDirectory()` here

SingularityService/src/main/resources/templates/task.jade
tpetr: i'd be super nice from a readability perspective to strip off the path leading up to where the file is (i.e. the stuff before the relative path specified in the config)

SingularityService/src/main/resources/templates/task.jade
```diff
@@ -106,21 +106,13 @@ html
...
+              dt(style="font-weight: bold; margin: 0 0 9px") #{ logName }
+              dd(style="background: #eee; display: block; margin: 0 0 17px; padding: 10px 14px 0 14px; border-radius: 3px")
+                if logTail
+                  pre(style="white-space: pre; margin: 0")
+                    code(style="font-family: Menlo, Monaco, Consolas, Courier, monospace")
+                      for line in logTail
+                        div(style="font-family: Menlo, Monaco, Consolas, Courier, monospace; white-space: pre; margin-bottom: 12px")= line
+                else
+                  | #{ logName } not available. The task either did not start, output anything, or the slave hosting the log files was unavailable over http.
```
tpetr: We should see if we can discriminate between an empty logfile vs. one that doesn't exist vs. the slave being offline / broken

tpetr:
a bunch of the comments I added are optional or should probably be separate issues, i think the only really essential one is just including the relative file path in the email

SingularityService/src/main/resources/templates/task.jade
tpetr: it'd also be amazing if we were able to directly link to the tail view of the file in the email

wolfd:
image

I fixed the css in my latest commit, should have the same amount of space above and below the text now.

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
```diff
@@ -125,7 +132,7 @@ public void stop() throws Exception {
 
     final String fullPath = String.format("%s/%s", directory.get(), filename);
 
-    final Long logLength = Long.valueOf(maybeSmtpConfiguration.get().getTaskLogLength());
+    final Long logLength = (long)maybeSmtpConfiguration.get().getTaskLogLength();
```
tpetr: why the change here?

SingularityService/src/main/java/com/hubspot/singularity/smtp/JadeHelpers.java
```diff
@@ -0,0 +1,101 @@
...
+   * @param taskId which task to link to.
+   * @return link.
+   */
+  public String getSingularityTaskLink(String taskId) {
+    if (!uiHostnameAndPath.isPresent()) {
+      return "";
+    }
+
+    return String.format(TASK_LINK_FORMAT, uiHostnameAndPath.get(), taskId);
+  }
```
tpetr: all these links will be invalid if `uiHostnameAndPath` is not set, better to explicitly not even generate the link(s) instead of have links with blank `href`s

tpetr:
what's the reasoning behind adding the function calls inside the jade template? i can see it being harder to debug this way, and I think it's generally good practice to have the templates be completely dumb (i.e. just variable substitutions + if statements) -- @timmfin do you agree?

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
wolfd: Mainly because IntelliJ was telling me that it was unnecessarily boxing it, although I just ran some tests and it looks like this is marginally slower than what was there originally.

wolfd:
The reason I put the logic for links into Jade helpers was to make it so I could pass in key value pairs to link to the log emails. That ended up not being worth the effort.

I still think that it would be worth bringing these functions away from the SingularityMailer class, and changing these over from Jade helpers to a pretty generic helper class for SingularityMailer to use would be pretty trivial.

I definitely agree that it is harder to debug.

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -0,0 +1,216 @@
...
+      // To enable support for tailing the service.log file, replace instances of $MESOS_TASK_ID.
+      filepath = filepath.replaceAll("\\$MESOS_TASK_ID", MesosUtils.getSafeTaskIdForDirectory(taskId.getId()));
+
+      logTails.add(
+          ImmutableMap.<String, String>builder()
+              .put("path", filepath)
+              .put("file", getFileName(filepath))
+              .put("link", getSingularityLogLink(filepath, taskId.getId()))
+              .put("log", getTaskLogFile(taskId, filepath, task, directory).or(""))
+              .build());
```
tpetr: instead of a Map<>, this should probably be a `SingularityMailTaskLog` class

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -0,0 +1,216 @@
...
+  public List<Map<String, String>> getJadeTaskHistory(Collection<SingularityTaskHistoryUpdate> taskHistory) {
+    List<Map<String, String>> output = Lists.newArrayListWithCapacity(taskHistory.size());
+
+    for (SingularityTaskHistoryUpdate taskUpdate : taskHistory) {
+      output.add(
+          ImmutableMap.<String, String> builder()
+              .put("date", DateFormatUtils.formatUTC(taskUpdate.getTimestamp(), TASK_DATE_PATTERN))
+              .put("update", WordUtils.capitalize(taskUpdate.getTaskState().getDisplayName()))
+              .put("message", taskUpdate.getStatusMessage().or(""))
+              .build());
```
tpetr: instead of a Map<>, this should probably be a `SingularityMailTaskHistoryUpdate` class

PR #414, opened 2015-01-21
Custom task page links

wolfd:
Adds the ability to create custom links using a task's environment variables.

For example:

```yaml
ui:
  customQuickLinks:
    - name: Example Button
      url: http://{{TASK_HOST}}/{{TASK_REQUEST_ID}}
```

Will generate a link on the task page that looks like
image

In order to support handlebars syntax in the custom links, the full handlebars js is included, not only the runtime. This is needed to do client-side compiling of handlebars text.


SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
```diff
@@ -21,6 +25,9 @@
   @JsonProperty
   private String baseUrl;
 
+  @JsonProperty
+  private List<HashMap<String, String>> customQuickLinks = new LinkedList<>();
```
tpetr: i'd consider renaming this `taskQuickLinks`. Also better to use `Collections.emptyList()` as opposed to allocating an entirely new linked list.

tpetr: it'd also probably be better to make an actual class for the QuickLink data

SingularityService/src/main/java/com/hubspot/singularity/config/TaskQuickLink.java
```diff
@@ -0,0 +1,35 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * TaskQuickLink defines a "quick-link" button to show on the
+ */
+public class TaskQuickLink {
+  /**
+   * Name of the button label in the SingularityUI interface
+   */
+  @JsonProperty
```
tpetr: `@JsonProperty` is unnecessary for immutable classes like this (just need `@JsonCreator` constructor + getters)

SingularityService/src/main/java/com/hubspot/singularity/views/IndexView.java
```diff
@@ -24,7 +27,9 @@
   private final Integer slaveHttpPort;
   private final Integer slaveHttpsPort;
 
-  public IndexView(SingularityConfiguration configuration) {
+  private final String taskQuickLinks;
```
tpetr: i'd be more explicit and call this `taskQuickLinksJSON`

SingularityService/src/main/java/com/hubspot/singularity/config/TaskQuickLink.java
```diff
@@ -0,0 +1,35 @@
...
+  private final String name;
+  /**
+   * Handlebars string to parse to generate a URL for the quick link
+   * (e.g. "http://{{TASK_HOST}}/{{TASK_REQUEST_ID}}")
+   */
+  @JsonProperty
+  private final String url;
+
+  @JsonCreator
+  public TaskQuickLink(@JsonProperty("name") String name, @JsonProperty("url") String url) {
```
tpetr: We talked about this in person, but it'd be nice if we had a way to specify which request type(s) a quick link applies to (i.e. `EnumSet<RequestType> requestTypes`).

The RequestType enum is coming in #348 (not merged yet), so it might make sense to do this in a 2nd PR.

wolfd: I think a second PR makes sense, it doesn't change the structure of how the TaskQuickLinks are specified, just an optional additional field if the quick link isn't supposed to apply to every type.

tpetr:
This is less necessary now that we're finshing up https://github.com/HubSpot/Singularity/pull/515

PR #415, opened 2015-01-21
change cmd line args to a list

wsorenson:


ssalinas:
@wsorenson added the ui update to send an array

PR #370, opened 2014-12-21
Adding more decommission states

wsorenson:
- Addressing #354 
- Cleaning up decommission code
- Adding decommission tests

PR #416, opened 2015-01-23
New status page (WIP)

ssalinas:
also incorporates #370 

Presents more information on the status page, currently looks like:
screen shot 2015-01-23 at 9 28 09 am

most items are clickable and you get a tooltip when mousing over the graph
screen shot 2015-01-23 at 9 38 30 am


ssalinas:
closing in favor of #430

PR #417, opened 2015-01-26
Thread pools

wsorenson:
@hgschmie thoughts?

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedScheduledExecutorServiceFactory.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedScheduledExecutorServiceFactory.java
@@ -1,0 +1,63 @@
+package com.hubspot.singularity;
+
+import static com.google.common.base.Preconditions.checkState;
+import io.dropwizard.lifecycle.Managed;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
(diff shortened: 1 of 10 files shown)


SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedScheduledExecutorServiceFactory.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Override
+  public void start() throws Exception {
+    // Ignored
+  }
+
+  @Override
+  public void stop() throws Exception {
+    if (!stopped.getAndSet(true)) {
+      for (ScheduledExecutorService service : executorPools) {
```
mgreene: At first glance, this seems like it's open to a `ConcurrentModificationException` because the `get(String)` method above modifies `executorPools` and this method has no shared `synchronized` lock with the writer.

SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedScheduledExecutorServiceProvider.java
```diff
@@ -0,0 +1,45 @@
...
+
+  private final long timeoutInSeconds;
+
+  public SingularityManagedScheduledExecutorServiceProvider(final int poolSize, final long timeoutInSeconds, final String name) {
+    this.service = Executors.newScheduledThreadPool(poolSize, new ThreadFactoryBuilder().setNameFormat(name + "-pool-%d").setDaemon(true).build());
+    this.timeoutInSeconds = timeoutInSeconds;
+  }
+
+  @Override
+  public synchronized ScheduledExecutorService get() {
```
mgreene: You can probably drop the `synchronized` lock here given you are reading from a thread safe data structure. Also there's no corresponding lock on the writer via the `stop` method below.

PR #418, opened 2015-01-26
Request Type UI

ssalinas:
@wsorenson 
adds filters for each request type:
screen shot 2015-01-26 at 2 39 39 pm

fixed to submit the request type from the new requests dialog:
screen shot 2015-01-26 at 2 39 52 pm


ssalinas:
found one more spot the request type is used, will push up changes in a second

ssalinas:
fixed the extras now, had to update the user dashboard numbers as well as the header in the request detail page

PR #422, opened 2015-01-28
custom executor documentation

tpetr:
@wolfd check this out, let me know if anything is unclear or doesnt work


Docs/slave_extras.md
```diff
@@ -0,0 +1,164 @@
+## Singularity Extras for Mesos Slaves
+
+*This assumes you've already followed the steps from [install.md](install.md)*
+
```
eherot: Maybe add something here explaining exactly what someone might need to use the Singularity Custom Executor for.

PR #423, opened 2015-01-28
Fix sigKillProcessesAfterMillis not being set correctly from the constructor.

wolfd:


tpetr:
the `ExecutorDataBuilder` constructor is usually called when we want a builder out of a preexisting `ExecutorData` instance, and adding this extra arg should cause compile errors where people are using it -- is nobody actually using this constructor?

wsorenson:
Is definitely a bug, but we may be able to just remove that constructor.

PR #429, opened 2015-01-30
Decom tweaks

ssalinas:
@wsorenson this will fix the decommission badge showing up on a non-decommissioning task and make the message an alert on the task detail page instead of a badge
screen shot 2015-01-30 at 12 50 07 pm


ssalinas:
also fixes the missing padding below the empty table message on racks/slaves pages

PR #435, opened 2015-02-03
Fix appRoot for catchall resource.

wolfd:
When using the catchall resource, explicitly setting "/" the appRoot makes
"http://<stuff>/singularity/" become the appRoot in the index script.

The rest of the UI expects the appRoot to not end in a "/".


tpetr:
nice catch, thanks

PR #436, opened 2015-02-05
more logfetch features (WIP)

ssalinas:


ssalinas:
@tpetr lots of changes in here, but these should be good to go now if you want to take a look

scripts/logfetch/grep.py
```diff
@@ -2,18 +2,21 @@
 import sys
 from termcolor import colored
 
-GREP_COMMAND_FORMAT = 'xargs -n {0} {1} < {2}'
+GREP_COMMAND_FORMAT = 'while read file; do {0} "$file"; done < {1}'
```
tpetr: why the change here?

ssalinas: More complicated custom grep commands weren't playing nicely with xargs. It was more flexible to loop through the file so any grep command that is used will output the way you'd expect it to.

PR #438, opened 2015-02-06
Add chart for tasks resource usage

benrlodge:
Adds a chart for rss vs limit memory in the resources section in tasks, including a couple new handlebars helpers.


tpetr:
:dancer:

PR #439, opened 2015-02-06
Fix calculation on memory resource meter

benrlodge:


SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -16,7 +16,7 @@ Handlebars.registerHelper "ifAll", (conditions..., options)->
     options.fn @
 
 Handlebars.registerHelper 'percentageOf', (v1, v2) ->
-    Math.round(v1/v2)
+    v1/v2 * 100
```
tpetr: might want to wrap the divison in parens just in case

PR #440, opened 2015-02-09
use hostname + web_url in framework info

wsorenson:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityDriver.java
```diff
@@ -31,14 +33,23 @@
...
-        .setUser("")  // let mesos assign
-        .build();
+        .setUser("");  // let mesos assign
+
+    if (singularityConfiguration.getHostname().isPresent()) {
+      frameworkInfoBuilder.setHostname(singularityConfiguration.getHostname().get());
+    }
+
+    if (singularityConfiguration.getUiConfiguration().getBaseUrl().isPresent()) {
+      frameworkInfoBuilder.setWebuiUrl(singularityConfiguration.getUiConfiguration().getBaseUrl().get());
```
tpetr: this will break when people use the setting that hosts the UI at `/ui`

wsorenson: What's the fix?

tpetr:
i want to verify one thing real quick before merging

PR #441, opened 2015-02-11
round humanize file size helper to 2 decimal places

benrlodge:
@tpetr - reworked the handlebars helper a bit

### DIFF
--- a/SingularityUI/app/handlebarsHelpers.coffee
+++ b/SingularityUI/app/handlebarsHelpers.coffee
@@ -76,21 +76,13 @@ Handlebars.registerHelper 'humanizeText', (text) ->
     text
 
 # 2121 => '2 KB'
-Handlebars.registerHelper 'humanizeFileSize', (fileSize) ->
-    kilo = 1024
-    mega = 1024 * 1024
-    giga = 1024 * 1024 * 1024
+Handlebars.registerHelper 'humanizeFileSize', (bytes) ->
+    k = 1024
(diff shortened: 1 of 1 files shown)


SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -76,21 +76,13 @@ Handlebars.registerHelper 'humanizeText', (text) ->
...
-    else if fileSize > kilo
-        return "#{ shorten kilo } KB"
-    else
-        return "#{ fileSize } B"
+Handlebars.registerHelper 'humanizeFileSize', (bytes) ->
+    k = 1024
+    sizes = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+
+    return '0 B' if bytes is 0
+    i = Math.floor(Math.log(bytes) / Math.log(k))
```
tpetr: I know we'd never go beyond YB, but just to be 100% robust we should do `Math.min(Math.floor(Math.log(bytes) / Math.log(k)), sizes.length-1)` to ensure we dont go past the length of sizes

SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -76,21 +76,13 @@ Handlebars.registerHelper 'humanizeText', (text) ->
...
+    return (bytes / Math.pow(k, i)).toPrecision(3) + ' ' + sizes[i]
```
tpetr: i think you want `toFixed(2)` instead

tpetr:
:dancer:

PR #442, opened 2015-02-11
Add dashboad sorting

benrlodge:
@tpetr - I only added sorting to the dashboard - not sure if we wanted searching and filtering capabilities as well, or if we wanted to keep it simple, but I left it out for now.

### DIFF
--- a/SingularityUI/app/templates/dashboard.hbs
+++ b/SingularityUI/app/templates/dashboard.hbs
@@ -67,52 +67,11 @@
         </div>
     {{/if}}
     <div class="row">
-        <div class="col-md-12">
+        <div class="col-md-12 table-staged">
             <div class="page-header">
                 <h2>Starred requests</h2>
             </div>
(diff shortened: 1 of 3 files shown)


Eseb:
I was actually gonna do this myself sometime this month, grumble grumble.

Previously Singularity used to have [sortable](http://github.hubspot.com/sortable/) for all its tables, but I took it out because of a few inconsistencies with what I was trying to do with the Requests & Tasks tables.

The thing is that the tables still have all the little sortable annotations (like marking which columns weren't sortable) so it should be easy to add it back for all the tables, only I wanted to make a theme for it to resolve all the aforementioned inconsistencies.

So if you'd like to check out going down that route, it might be better to integrate sorting for most tables using sortable. I can do it myself, too. I'll leave it up to you guys.

tpetr:
we'll investigate better use of sortable / mixins / etc in a future PR

PR #446, opened 2015-02-17
logfetch fixes

ssalinas:
@tpetr

### DIFF
--- a/scripts/logfetch/entrypoint.py
+++ b/scripts/logfetch/entrypoint.py
@@ -10,6 +10,7 @@ from tail import start_tail
 from grep import grep_files
 from cat import cat_files
 
+VERSION = '0.0.10'
 CONF_READ_ERR_FORMAT = 'Could not load config from {0} due to {1}'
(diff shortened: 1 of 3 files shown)


scripts/logfetch/entrypoint.py
```diff
@@ -10,6 +10,7 @@
 from grep import grep_files
 from cat import cat_files
 
+VERSION = '0.0.10'
```
tpetr: i know the versioning goofup was my mistake, but can we consolidate on `0.X.0` going forward? best to keep all versions in order, and its more understandable to see version #'s go up instead of down

scripts/setup.py
```diff
@@ -10,7 +10,7 @@
 
 setup(
     name='singularity-logfetch',
-    version='0.0.9',
+    version='0.0.10',
```
tpetr: best to set `VERSION="X.Y.Z"` in `logfetch/__init__.py` and import it here, so that we don't have 2 values floating around

tpetr:
LGTM after comments are addressed, thanks

PR #452, opened 2015-02-23
per-uploader credentials

wsorenson:
- s3uploader can have its own credentials
- s3uploader can watch a PID to expire itself


tpetr:
LGTM

PR #453, opened 2015-02-24
visualize CPU usage in task detail

benrlodge:
Normal usage:
screenshot 2015-02-24 08 19 04

And if usage is greater than 110% a badge is added and the meter turns red:
screenshot 2015-02-24 08 21 33

@tpetr

### DIFF
--- a/SingularityUI/app/controllers/TaskDetail.coffee
+++ b/SingularityUI/app/controllers/TaskDetail.coffee
@@ -82,13 +82,6 @@ class TaskDetailController extends Controller
 
         app.showView @view
 
-        #  Refresh once again on page load to calculate initial CPU Usage value
-        @initialPageLoad = true
-        if @initialPageLoad
-            setTimeout (=>
-                @refresh() 
-                @initialPageLoad = false
-            ), 2000
 
     fetchResourceUsage: ->
         @models.resourceUsage?.fetch()
@@ -96,6 +89,10 @@ class TaskDetailController extends Controller
                 # Store current resource usage to compare against future resource usage
                 @models.resourceUsage.setCpuUsage() if @models.resourceUsage.get('previousUsage')              
                 @models.resourceUsage.set('previousUsage', @models.resourceUsage.toJSON())
+                
+                if not @resourcesFetched
+                    setTimeout (=> @fetchResourceUsage() ), 2000
+                    @resourcesFetched = true
 
             .error =>
                 # If this 404s there's nothing to get so don't bother
@@ -103,6 +100,8 @@ class TaskDetailController extends Controller
                 delete @models.resourceUsage
 
     refresh: ->
+        @resourcesFetched = false
+
         @models.task.fetch()
             .done =>
                 @fetchResourceUsage() if @models.task.get('isStillRunning')
--- a/SingularityUI/app/controllers/TaskDetail.coffee
+++ b/SingularityUI/app/controllers/TaskDetail.coffee
@@ -82,23 +82,15 @@ class TaskDetailController extends Controller
 
         app.showView @view
 
-        #  Refresh once again on page load to calculate CPU Usage 
+        #  Refresh once again on page load to calculate initial CPU Usage value
         @initialPageLoad = true
-
         if @initialPageLoad
             setTimeout (=>
                 @refresh() 
                 @initialPageLoad = false
             ), 2000
 
-    refresh: ->
-        @models.task.fetch
-            error: =>
-                # If this 404s the task doesn't exist
-                app.caughtError()
-                app.router.notFound()
-
-
+    fetchResourceUsage: ->
         @models.resourceUsage?.fetch()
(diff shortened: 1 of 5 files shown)


SingularityUI/app/models/TaskResourceUsage.coffee
```diff
@@ -6,4 +6,19 @@ class TaskResourceUsage extends Model
...
 
-module.exports = TaskResourceUsage
+    # Calculate CPU Usage by comparing previous usage to current usage
+    setCpuUsage: ->
+        previous = @get('previousUsage')
+        currentTime = @get('cpusSystemTimeSecs') + @get('cpusUserTimeSecs')
+        previousTime = previous.cpusSystemTimeSecs + previous.cpusUserTimeSecs
+        timestampDiff = @get('timestamp') - previous.timestamp
+        cpus_used_raw = (currentTime - previousTime) / timestampDiff
+        cpus_used = Math.round(cpus_used_raw * 100) / 100
```
tpetr: might be best if we pass the full precision value through, and then use `.toFixed(2)` when rendering

SingularityUI/app/controllers/TaskDetail.coffee
```diff
@@ -89,10 +98,17 @@ class TaskDetailController extends Controller
...
+        @models.resourceUsage?.fetch()
+            .done =>
+                # Store current resource usage to compare against future resource usage
+                @models.resourceUsage.setCpuUsage() if @models.resourceUsage.get('previousUsage')              
+                @models.resourceUsage.set('previousUsage', @models.resourceUsage.toJSON())
+
+            .error =>
+                # If this 404s there's nothing to get so don't bother
+                app.caughtError()
+                delete @models.resourceUsage
```
tpetr: could we be a little more clever here and only fetch resource usage if the task hasn't hit a terminal state (`TASK_KILLED`, `TASK_FINISHED`, `TASK_LOST`, etc.)?

SingularityUI/app/models/TaskResourceUsage.coffee
```diff
@@ -6,4 +6,18 @@ class TaskResourceUsage extends Model
...
+    # Calculate CPU Usage by comparing previous usage to current usage
+    setCpuUsage: ->
+        previous = @get('previousUsage')
+        currentTime = @get('cpusSystemTimeSecs') + @get('cpusUserTimeSecs')
+        previousTime = previous.cpusSystemTimeSecs + previous.cpusUserTimeSecs
+        timestampDiff = @get('timestamp') - previous.timestamp
+        cpus_used = (currentTime - previousTime) / timestampDiff
+        cpuUsageExceeding = (cpus_used / @get('cpusLimit')) > 1.10
+        
+        if cpuUsageExceeding
```
timmfin: Would be useful to have another state between the "green" and red/danger exceeding states... like a yellow "warning" state when  `cpus_used` is >= 80% of the `cpusLimit`?

benrlodge: @tpetr - do you mean what renders in the DOM? Are you referring to something else? The resources section doesn't render if the task is inactive.

tpetr: what happens when the task transitions from running to finished, though? do we display the last valid resource usage, or remove it completely? (i'm literally not sure, need to double check)

need to decide what the proper action would be -- i'm personally leaning towards hiding resource usage

benrlodge: ah got it @tpetr. re: tim's additional states, thoughts? If we do several it may be worth moving the logic out to a handlebars helper.

tpetr: i think the additional states are a good idea, but let's implement them in a separate PR

SingularityUI/app/controllers/TaskDetail.coffee
```diff
@@ -82,17 +82,36 @@ class TaskDetailController extends Controller
 
         app.showView @view
 
+        #  Refresh once again on page load to calculate initial CPU Usage value
+        @initialPageLoad = true
+        if @initialPageLoad
```
tpetr: isn't this `if` statement essentially a no-op, since it's being set to `true` in the line before? @timmfin might have some advice on how best to do the initial refresh for CPU usage data

benrlodge: @tpetr right, thanks for pointing out... will move to call second resource fetch after initial resourced fetched, where then I'll reset the state for future refreshes.

benrlodge:
@tpetr have you thought any more about hiding resource usage if a task is no longer running?

stevenschlansker:
Snazzy!

tpetr:
@benrlodge and I chatted offline and we're going to figure out what to do about hiding / invalidating resource usage in a separate PR.

PR #454, opened 2015-02-24
upgrade to Baragon 0.1.4 + handle INVALID_REQUEST_NOOP

tpetr:
@wsorenson @ssalinas could you help verify this logic is accurate?

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
@@ -383,6 +383,7 @@ public class SingularityCleaner {
       case UNKNOWN:
       case WAITING:
       case SUCCESS:
+      case INVALID_REQUEST_NOOP:
         return true;
       default:
         LOG.trace("Task {} had abnormal LB state {}", taskId, loadBalancerUpdate);
@@ -417,6 +418,7 @@ public class SingularityCleaner {
       case CANCELING:
       case SUCCESS:
       case WAITING:
+      case INVALID_REQUEST_NOOP:
     }
 
     return false;
@@ -459,6 +461,7 @@ public class SingularityCleaner {
 
     switch (lbRemoveUpdate.getLoadBalancerState()) {
       case SUCCESS:
+      case INVALID_REQUEST_NOOP:
         return CheckLBState.DONE;
       case FAILED:
       case CANCELED:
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
```diff
@@ -300,6 +300,7 @@ private CheckTaskState getTaskState(SingularityTask task) {
   private Optional<CheckTaskState> checkLbState(BaragonRequestState lbState) {
     switch (lbState) {
       case SUCCESS:
+      case INVALID_REQUEST_NOOP:
         return Optional.of(CheckTaskState.HEALTHY);
```
wsorenson: Is a task healthy if baragon says the lb state is INVALID?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
```diff
@@ -383,6 +383,7 @@ private boolean shouldRemoveLbState(SingularityTaskId taskId, SingularityLoadBal
       case UNKNOWN:
       case WAITING:
       case SUCCESS:
+      case INVALID_REQUEST_NOOP:
```
wsorenson: In this case, we don't need to remove from Baragon b/c Baragon doesn't know about it, right?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
tpetr: in my original notes, INVAILD_REQUEST_NOOP should be fired when removing a nonexistent upstream, or removing a service from a nonexistent LB -- let's sync up if that state is supposed to mean something else

wsorenson: And if we get it back from an ADD request, I can't see the client handling that as nominal

tpetr: right -- good point

ssalinas:
@tpetr added a condition for `INVALID_REQUEST_NOOP` to go to a `FAILED` deploy state in deploy checker, imitating the behavior of a `FAILED` load balancer request

PR #455, opened 2015-02-24
S3 uploader and logrotate edge cases

wsorenson:


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskDefinition.java
```diff
@@ -84,10 +87,23 @@ public String getTaskId() {
...
+  }
+
+  @JsonIgnore
+  public Optional<Integer> getExecutorPidSafe() {
+    try {
+      return Optional.of(Integer.parseInt(executorPid));
+    } catch (NumberFormatException nfe) {
+      return Optional.<Integer> absent();
+    }
+  }
```
mjball: `return Optional.fromNullable(Ints.tryParse(executorPid))`

PR #456, opened 2015-02-24
Add build recipe to actually compile singularity

eherot:


cookbook/recipes/build.rb
```diff
@@ -0,0 +1,27 @@
...
+  user       node[:singularity][:user]
+  action     :export
+end
+
+execute 'build_singularity' do
+  action  :run
+  # Maven (or rather npm) has issues with
+  # being run as root.
+  user    node[:singularity][:user]
+  environment('HOME' => '/home/singularity')
```
wolfd: Is this supposed to be hardcoded?

eherot: Good catch. Will fix.

tpetr:
LGTM, but I'd only suggest using this installation method (building straight from git) for testing / playing with Singularity.

The preferred installation method for production environments (in my opinion) would be to download a pre-compiled released JAR from Maven (i.e. the `shaded.jar` links on http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.hubspot%22%20AND%20a%3A%22SingularityService%22). Could we add support for this too?

eherot:
I would be totally in favor of adding that. I didn't even know it existed until just now.

PR #457, opened 2015-02-24
Add a recipe to install the Singularity Executor and its config files

eherot:
Depends on build recipe from #456


wolfd:
Also depends on #464

tpetr:
could we fold this into #464 to reduce confusion?

eherot:
Sure. I'll do that right now.

PR #458, opened 2015-02-24
Use the Mesos cookbook to install mesos instead of doing it ourselves

eherot:


cookbook/attributes/default.rb
```diff
@@ -14,21 +14,8 @@
...
-  },
-  :slave => {
-    :containerizers => 'docker,mesos',
-    :switch_user => 'false'
-  },
-  :slave_resources => {},
-  :slave_attributes => {}
-}
+set[:mesos][:type] = 'mesosphere'
+default[:mesos][:master][:zk] = 'http://localhost:2181/singularity'
```
tpetr: the zk connection string here should be `zk://localhost:2181/mesos`, we're also going to need to set a matching value for `default[:mesos][:slave][:master]`

cookbook/attributes/default.rb
```diff
@@ -14,21 +14,8 @@
...
-    :cluster => 'vagrant-singularity',
-  },
-  :slave => {
-    :containerizers => 'docker,mesos',
-    :switch_user => 'false'
-  },
-  :slave_resources => {},
-  :slave_attributes => {}
-}
+set[:mesos][:type] = 'mesosphere'
```
tpetr: we should set `node[:mesos][:mesosphere][:with_zookeeper]` here too

cookbook/attributes/default.rb
tpetr: actually i see it's in `.kitchen.yml` -- what's the difference here? might be best if it's all in one place

cookbook/attributes/default.rb
```diff
@@ -14,21 +14,8 @@
...
-  :package_version => "0.21.0-1.0.ubuntu1404",
-  :common => {
-    :ip => private_ip,
-  },
-  :master => {
-    :cluster => 'vagrant-singularity',
-  },
-  :slave => {
-    :containerizers => 'docker,mesos',
-    :switch_user => 'false'
```
tpetr: we need to make sure these defaults get carried over to the `cookbook-mesos` version. the Vagrantfile might have to be updated too -- it sets some default slave resources

eherot: How are these options actually being used? They don't seem to appear anywhere in the cookbook...

cookbook/attributes/default.rb
eherot: `attributes` is the defaults for actual nodes. `.kitchen.yml` is just used for test-kitchen.

eherot: As such it is probably sane to set a default for `default[:mesos][:slave][:master]` here.

cookbook/attributes/default.rb
eherot: Done.

cookbook/Berksfile
```diff
@@ -1,3 +1,7 @@
 source 'https://supermarket.getchef.com'
 
 metadata
+
+cookbook 'mesos',
+         github: 'everpeace/cookbook-mesos',
+         ref: '85d4a8f5c441402844775d56b165919c99f8a546'
```
tpetr: this appears to conflict with `mesos` in [metadata.rb](https://github.com/evertrue/Singularity/blob/evertrue/eherot/use_mesos_cookbook/cookbook/metadata.rb#L15) for me

tpetr:
i can't seem to get chef to converge -- issues `node[:singularity][:user]` needing a default, and then the mesos recipe not being found. are you seeing this on your end?

ssalinas:
closing this since cookbook has moved

PR #459, opened 2015-02-24
Logfetch v12 (WIP)

ssalinas:
- Add search over all s3 logs when no tasks are found in date range

### DIFF
--- a/scripts/logfetch/entrypoint.py
+++ b/scripts/logfetch/entrypoint.py
@@ -87,7 +87,8 @@ def fetch():
     "chunk_size" : DEFAULT_CHUNK_SIZE,
     "dest" : DEFAULT_DEST,
     "task_count" : DEFAULT_TASK_COUNT,
-    "start_days" : DEFAULT_DAYS
+    "start_days" : DEFAULT_DAYS,
+    "end_days" : 0 #today
   }
 
   try:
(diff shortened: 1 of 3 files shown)


scripts/logfetch/entrypoint.py
```diff
@@ -91,17 +108,16 @@ def fetch():
...
   parser.add_argument("-g", "--grep", dest="grep", help="Regex to grep for (normal grep syntax) or a full grep command")
 
   args = parser.parse_args(remaining_argv)
 
-  if args.deployId and not args.requestId:
-    exit("Must specify request-id (-r) when specifying deploy-id")
-  elif not args.requestId and not args.deployId and not args.taskId:
-    exit('Must specify one of\n -t task-id\n -r request-id and -d deploy-id\n -r request-id')
+  check_args(args)
+  args.start_days = convert_to_days(args.start_days) if args.start_days else args.start_days
```
tpetr: what does the if/else do here? only calls `convert_to_days()` if its a true-ish value?

ssalinas: if you provide a value for args.start_days, call that to make sure it's value that can be used, if not, keep it as is. I can rewrite it to be more clear

ssalinas: On second thought, we provide a default value for that, so we don't need the if to check that it isn't None

tpetr:
LGTM thanks

PR #463, opened 2015-02-26
Use OpenSSL library cookbook to generate a random mysql root password

eherot:


cookbook/recipes/mysql.rb
```diff
@@ -14,10 +28,18 @@
   action :create
 end
 
+file '/etc/mysql-default/root_password' do
```
eherot: It actually might make more sense to do this in `/root/.my.cnf` but I didn't want to create any potential conflicts with existing setups.

ssalinas:
closing this since cookbook has moved

PR #466, opened 2015-02-26
fix bug in original custom executor resources implementation

tpetr:
my original implementation had a bug -- it was correctly carving out task + executor resources from the offer, but then incorrectly setting the task resources to that total amount. the executor amount was being tacked on at task launch time, which caused issues in QA.


tpetr:
(this wasn't caught while in the testing cluster due to the fact that it's not as packed as our QA or PROD clusters, so the generous offers Singularity was receiving from Mesos were satisfying the inflated resource amounts)

PR #467, opened 2015-02-28
tweaks to MesosUtils + Resources

tpetr:
- Return a empty `Ranges` instance instead of `null` in `getRanges()`
- Fix off-by-one error in `getNumRanges()`
- Add convenience methods for creating a port range resource (useful for unittests)
- Add `buildResourcesFromMesosResourceList()`, which converts `List<Resource>` into `Resources` (useful for unittests)
- add `equals()` and `hashCode()` to `Resources`

@wsorenson

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
+++ b/SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
@@ -56,7 +56,7 @@ public final class MesosUtils {
       }
     }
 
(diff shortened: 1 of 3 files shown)


wsorenson:
Fix a bug, how about adding a test?

tpetr:
^ `MesosUtilsTest.assertFound()` was re-implementing `MesosUtils.getNumPorts()`, so I updated the test to use the latter.

PR #471, opened 2015-03-04
Improved task cleaning ui

benrlodge:
This will change buttons, modal prompts, and kill types (sigterm, kill-9) for killing a task depending on whether a task is in cleanup or is cleaning. 
@tpetr

### DIFF
--- a/SingularityUI/app/templates/taskDetail/taskOverview.hbs
+++ b/SingularityUI/app/templates/taskDetail/taskOverview.hbs
@@ -23,6 +23,7 @@
         <div class="alert alert-warning" role="alert">This task is running on a decommissioning host and will be replaced soon</div>
     {{/if}}
 {{/if}}
+
 {{! Part of taskBase }}
 <header class='detail-header'>
     <div class="row">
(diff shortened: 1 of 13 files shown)


stevenschlansker:
Nice, we actually ran into a situation just yesterday where this would have been useful.

wsorenson:
Ben - keep in mind other executors won't behave the same as ours, so I'm not sure what the wording here is , but kill -9 is definitely something specific to Singularity executors.

@stevenschlansker given what I've said above, and you're not using the Singularity Executor, I'm not sure it would have helped you (although there are some cases this PR should improve for everyone.)

SingularityUI/app/templates/taskDetail/taskOverview.hbs
```diff
@@ -1,8 +1,29 @@
+{{#if data.isStillRunning}}
+    {{#ifNotEqual data.killStatus 'kill9'}}
+        {{#if data.isStillRunning}}
+            {{#if data.isInCleanup}} 
+                <div class="alert alert-warning" role="alert">Request is bouncing. This task will be killed once all replacement tasks are healthy. You may still force kill this task.</div>
+            {{/if}}
+            {{#unless data.isInCleanup}}
```
tpetr: can't we use `{{else}}` here?

SingularityUI/app/templates/taskDetail/taskOverview.hbs
```diff
@@ -1,8 +1,29 @@
+{{#if data.isStillRunning}}
+    {{#ifNotEqual data.killStatus 'kill9'}}
+        {{#if data.isStillRunning}}
+            {{#if data.isInCleanup}} 
+                <div class="alert alert-warning" role="alert">Request is bouncing. This task will be killed once all replacement tasks are healthy. You may still force kill this task.</div>
```
tpetr: not sure we can safely always use this message in this state -- probably should be surfacing the message from the `SingularityTaskCleanup` object

SingularityUI/app/templates/vex/taskKillDestroy.hbs
```diff
@@ -0,0 +1,5 @@
+<p>Are you sure you want to destroy this task?</p>
+<pre>{{ id }}</pre>
+<p>
+    This will override cleanup type, immediately initiate a non-graceful kill (-9 PID) and not wait for replacement tasks.
```
tpetr: might want to reword to `This will override any pending cleanup operations and immediately initiate process termination (i.e. kill -9) without waiting for replacement tasks.`

SingularityUI/app/templates/taskDetail/taskOverview.hbs
benrlodge: yep, oversight. there's also an extra data.isStillRunning check that's unnecessary.

SingularityUI/app/templates/taskDetail/taskOverview.hbs
benrlodge: Are you referencing the `data.isInCleanup`? That's being set from a request to `tasks/cleaning` which I thought was the `SingularityTaskCleanup` object. Or are you referencing the `data.isStillRunning`?

SingularityUI/app/templates/vex/taskKillDestroy.hbs
wsorenson: More like "and immediately send a kill request to the executor"

SingularityUI/app/templates/taskDetail/taskOverview.hbs
tpetr: we chatted offline, but i was referring to the "Request is bouncing" part -- the `SingularityTaskCleanup` object has a `message` field that we should just use instead

PR #474, opened 2015-03-09
Fix slaves rendering

benrlodge:
The Slaves view was re-rendering on page load for every model in the collection, causing a delay, this removes that issue. @tpetr


tpetr:
:dancer:

PR #475, opened 2015-03-10
(WIP) automatically go to tailing service.log (or other log file) after clicking 'run now'

benrlodge:
@tpetr - I took an initial run at this, doesn't use local storage to save the filename right now (as I mentioned), but otherwise will redirect to whatever filename is entered.

### DIFF
--- a/SingularityUI/app/models/Request.coffee
+++ b/SingularityUI/app/models/Request.coffee
@@ -121,14 +121,19 @@ class Request extends Model
                 @unpause().done callback
 
     promptRun: (callback) =>
+                  
         vex.dialog.prompt
             input: runTemplate id: @get "id"
             buttons: [
                 $.extend _.clone(vex.dialog.buttons.YES), text: 'Run now'
                 vex.dialog.buttons.NO
             ]
+            afterOpen: -> 
+                $('#filename').val localStorage.getItem('taskRunRedirectFilename')
+
             callback: (data) =>
                 return if data is false
+                localStorage.setItem('taskRunRedirectFilename', data.filename) if data.filename?
                 data.id = @get 'id'
                 @run(data.commandLineInput).done callback(data)
 
--- a/SingularityUI/app/models/Request.coffee
+++ b/SingularityUI/app/models/Request.coffee
@@ -122,14 +122,15 @@ class Request extends Model
 
     promptRun: (callback) =>
         vex.dialog.prompt
-            message: runTemplate id: @get "id"
+            input: runTemplate id: @get "id"
             buttons: [
                 $.extend _.clone(vex.dialog.buttons.YES), text: 'Run now'
                 vex.dialog.buttons.NO
             ]
             callback: (data) =>
                 return if data is false
-                @run(data).done callback
+                data.id = @get 'id'
+                @run(data.commandLineInput).done callback(data)
 
     promptRemove: (callback) =>
         vex.dialog.confirm
--- a/SingularityUI/app/templates/vex/requestRun.hbs
+++ b/SingularityUI/app/templates/vex/requestRun.hbs
@@ -1,3 +1,20 @@
 <p>Are you sure you want to run a task for this request immediately?</p>
 <pre>{{ id }}</pre>
-<p>Additional command line input (optional):</p>
+
+<p>Additional command line input <small>(optional)</small>:</p>
+<div class="vex-dialog-input">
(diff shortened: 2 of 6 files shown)


SingularityUI/app/templates/vex/requestRun.hbs
```diff
@@ -1,3 +1,20 @@
...
+<p>Wait for task to start, then bring me directly to a file <small>(optional)</small>:</p>
+
+<div class="vex-dialog-input">
+	<input id='filename' name="filename" type="text" class="vex-dialog-prompt-input" placeholder="e.g. service.log" value="">
+</div>
+
+<div class="vex-dialog-input checkbox">
+    <label>
+		<input id='storeSetting' name="storeSetting" type="checkbox"> Save this setting in local storage
+    </label>
```
tpetr: can we simplify the UI here by just storing the last value? (i.e. if it was checked last time, default it to checked this time)

benrlodge: Sure make sense, but are you saying still keep the checkbox - you said "if it was checked"? Or could we just drop the checkbox all together - just keep the input field, and then if they fill in a filename, we populate that next time?

benrlodge: @tpetr, pushed up that change mentioned ^ - now looks like: 

screenshot 2015-03-11 09 59 45

SingularityUI/app/models/Request.coffee
```diff
@@ -121,15 +121,21 @@ class Request extends Model
                 @unpause().done callback
 
     promptRun: (callback) =>
+                  
         vex.dialog.prompt
-            message: runTemplate id: @get "id"
+            input: runTemplate id: @get "id"
```
tpetr: does this need to change?

SingularityUI/app/templates/vex/requestRun.hbs
```diff
@@ -1,3 +1,14 @@
 <p>Are you sure you want to run a task for this request immediately?</p>
 <pre>{{ id }}</pre>
-<p>Additional command line input (optional):</p>
+
+<p>Additional command line input <small>(optional)</small>:</p>
+<div class="vex-dialog-input">
+	<input id='commandLineInput' name="commandLineInput" type="text" class="vex-dialog-prompt-input" placeholder="" value="">
+</div>
+<hr style='border-color: #dadada'>
```
tpetr: let's classify this

SingularityUI/app/templates/vex/requestRun.hbs
```diff
@@ -1,3 +1,14 @@
...
+
+<p>Wait for task to start, then bring me directly to a file <small>(optional)</small>:</p>
```
tpetr: lets make this a little more understandable for someone who's never used this before (i.e. leave blank for no action)

SingularityUI/app/views/request.coffee
```diff
@@ -61,11 +61,18 @@ class RequestView extends View
             app.router.navigate 'requests', trigger: true
 
     runRequest: (e) =>
-        @model.promptRun =>
-            @trigger 'refreshrequest'
-            setTimeout =>
+        @model.promptRun (data) =>   
+            # If user wants to redirect to a file after the task starts
+            if data.filename.length > 1
+                @listenTo @activeTasks, 'add', @redirectToFile
```
tpetr: is there an alternative that will un-listen once `@redirectToFile` is called? like a `@listenOnce`

tpetr:
we're going to simplify the logic and redirect the user to the file the want to tail as soon as the task starts -- @benrlodge is going to tweak the file tailing view so that the error message is less drastic, and the page will auto-update for as long as the task is running (this will be in a separate PR, and we'll rebase it in here once it's merged)

SingularityUI/app/models/Request.coffee
benrlodge: Yea, (as far as I can tell) this allows me to specify additional inputs (and the id/order) - since the message object before only loaded the additional command line input

SingularityUI/app/views/request.coffee
benrlodge: Yep we can do `listenToOnce`, but I also already call stopListening in the `stopPollingTask` method (line 116) which stops the event listener in the case that they cancel the redirect (as well as if they redirect). So we can change that, but I'll still be calling the `stopPollingTask` to cancel the interval, so it won't really have any noticeable effect, if I'm thinking about this right.

tpetr:
moving to #504

PR #477, opened 2015-03-11
stop tailing when task has finished

benrlodge:
right now the log tailing view will periodically hit Singularity for more data indefinitely, but when a task completes, it's a pretty safe bet that we're never going to get any more data from it. we should only auto-tail a task if it hasn't hit a terminal task state.

@tpetr

### DIFF
--- a/SingularityUI/app/views/tail.coffee
+++ b/SingularityUI/app/views/tail.coffee
@@ -29,6 +29,12 @@ class TailView extends View
         @listenTo @collection, 'sync', =>
             @$el.removeClass 'fetching-data'
 
+        @listenTo @model, 'change', @setTaskHistory
+
+
+    setTaskHistory: ->
+        @isTaskRunning = @model.get 'isStillRunning'
+
     handleAjaxError: (response) =>
         # ATM we get 404s if we request dirs and 500s if the file doesn't exist
         if response.status in [404, 500]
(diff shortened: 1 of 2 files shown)


SingularityUI/app/views/tail.coffee
```diff
@@ -142,16 +148,20 @@ class TailView extends View
...
         @scrollToBottom()
 
         clearInterval @tailInterval
         @tailInterval = setInterval =>
+            @stopTailing() if not @isTaskRunning
+
             @collection.fetchNext().done =>
                 # Only show the newly tail-ed lines if we are still tailing
                 @scrollToBottom() if @isTailing
+                @model.fetch()
```
tpetr: will `@model.fetch()` happen as part of the global refresh? that should probably be good enough, if so

SingularityUI/app/views/tail.coffee
```diff
@@ -29,6 +29,12 @@ class TailView extends View
         @listenTo @collection, 'sync', =>
             @$el.removeClass 'fetching-data'
 
+        @listenTo @model, 'change', @setTaskHistory
```
tpetr: this is stretching my backbone knowledge, but is it possible to do something like this?

```coffeescript
@listenTo @model, 'change:isStillRunning', => @stopTailing() unless @model.get 'isStillRunning'
```

benrlodge: yep that will work

SingularityUI/app/views/tail.coffee
benrlodge: We can do that, if we use the global refresh, we just still need to check at `startTailing` as that can still be run after the model is changed (after `initialdata` event and a couple other actions).

PR #479, opened 2015-03-13
Add option to persist more logrotated files in S3

wolfd:
Also adds documentation to some files I was reading.
Create the logrotate directory as part of the runner.

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
@@ -2,9 +2,10 @@ package com.hubspot.singularity.executor.config;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
+import java.util.List;
 
 import com.google.common.base.Optional;
+import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
@@ -12,8 +13,6 @@ import com.google.inject.name.Named;
 import com.hubspot.mesos.JavaUtils;
 import com.hubspot.mesos.MesosUtils;
 import com.hubspot.singularity.runner.base.config.SingularityRunnerBaseConfigurationLoader;
-
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 @Singleton
 public class SingularityExecutorConfiguration {
@@ -47,12 +46,12 @@ public class SingularityExecutorConfiguration {
   private final String logrotateDateformat;
 
   private final String logrotateExtrasDateformat;
-  private final String[] logrotateExtrasFiles;
+  private final List<String> logrotateExtrasFiles;
 
   /**
    * Extra files to backup to S3 besides the service log.
    */
-  private final String[] s3FilesToBackup;
+  private final List<String> additionalS3FilesToBackup;
 
   private final Path logMetadataDirectory;
   private final String logMetadataSuffix;
@@ -133,11 +132,7 @@ public class SingularityExecutorConfiguration {
     this.logrotateCount = logrotateCount;
     this.logrotateMaxageDays = logrotateMaxageDays;
     this.logrotateDateformat = logrotateDateformat;
-    if ((s3FilesToBackup != null) && (s3FilesToBackup.trim().length() > 0)) {
-      this.s3FilesToBackup = s3FilesToBackup.split(",");
-    } else {
-      this.s3FilesToBackup = new String[0];
-    }
+    this.additionalS3FilesToBackup = Splitter.on(",").trimResults().omitEmptyStrings().splitToList(s3FilesToBackup);
     this.s3Bucket = s3Bucket;
     this.s3KeyPattern = s3KeyPattern;
     this.s3MetadataSuffix = s3MetadataSuffix;
@@ -145,11 +140,7 @@ public class SingularityExecutorConfiguration {
     this.tailLogLinesToSave = Integer.parseInt(tailLogLinesToSave);
     this.serviceFinishedTailLog = serviceFinishedTailLog;
     this.logrotateExtrasDateformat = logrotateExtrasDateformat;
-    if ((logrotateExtrasFiles != null) && (logrotateExtrasFiles.trim().length() > 0)) {
-      this.logrotateExtrasFiles = logrotateExtrasFiles.split(",");
-    } else {
-      this.logrotateExtrasFiles = new String[0];
-    }
+    this.logrotateExtrasFiles = Splitter.on(",").trimResults().omitEmptyStrings().splitToList(logrotateExtrasFiles);
     this.useLocalDownloadService = Boolean.parseBoolean(useLocalDownloadService);
     this.localDownloadServiceTimeoutMillis = Long.parseLong(localDownloadServiceTimeoutMillis);
 
@@ -196,8 +187,7 @@ public class SingularityExecutorConfiguration {
     return logrotateExtrasDateformat;
   }
 
-  @SuppressFBWarnings("EI_EXPOSE_REP")
-  public String[] getLogrotateExtrasFiles() {
+  public List<String> getLogrotateExtrasFiles() {
     return logrotateExtrasFiles;
   }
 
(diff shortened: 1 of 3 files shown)


SingularityExecutor/src/main/resources/logrotate.conf.hbs
```diff
@@ -16,6 +16,7 @@ notifempty
 {{#if extrasFiles}}
 {{#each extrasFiles}}{{{this}}} {{/each}} {
   dateformat {{{ extrasDateformat }}}
+  olddir {{{ rotateDirectory }}}
```
wolfd: I'm not sure if this is really the right approach. I rotate all extra files into the rotateDirectory, and not only service.log

SingularityExecutor/src/main/resources/runner.sh.hbs
```diff
@@ -32,6 +32,13 @@ if [[ ! -d ./tmp ]]; then
   sudo chown -R {{{ user }}} ./tmp
 fi
 
+# Create log directory for logrotate runs
+if [[ ! -d {{{ logDir }}} ]]; then
+    echo "Creating log directory ({{{ logDir }}})"
+    mkdir {{{ logDir }}}
```
tpetr: might want `mkdir -p`, just in case

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskLogManager.java
```diff
@@ -148,8 +155,17 @@ private boolean writeTailMetadata(boolean finished) {
...
+   * The matching files are caught by the S3 Uploader and pushed to S3.
+   * @return file glob String.
+   */
   private String getS3Glob() {
-    return String.format("%s*.gz*", taskDefinition.getServiceLogOutPath().getFileName());
+    List<String> fileNames = new LinkedList<>();
+    fileNames.add(taskDefinition.getServiceLogOutPath().getFileName().toString());
+    fileNames.addAll(Arrays.asList(configuration.getS3FilesToBackup()));
+
+    return String.format("{%s}*.gz*", StringUtils.join(fileNames, ","));
```
tpetr: is this valid glob syntax? also for joinining, use guava's `Joiner.on(",").join()` instead of `StringUtils.join()` (not because StringUtils is necessarily bad, but it's best to standardize on the guava library here)

wolfd: The glob syntax seems to be fine, as it generates
"fileGlob":"{service.log,executor.bash.log,executor.java.log}*.gz*"
in the json config.

Changed to Joiner in the next commit.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/LogrotateTemplateContext.java
```diff
@@ -17,18 +25,35 @@ public String getRotateDateformat() {
...
+   * Logs are moved into $directory for rotation.
+   * @return directory.
+   */
   public String getRotateDirectory() {
     return configuration.getLogrotateToDirectory();
   }
 
+  /**
+   * Extra files for logrotate to rotate.
+   * @return filenames to rotate.
```
wsorenson: Are these extra <code>@return</code> statements useful?

SingularityService/src/main/java/com/hubspot/singularity/config/S3Configuration.java
```diff
@@ -9,18 +9,36 @@
...
+  /**
+   * How long for SingularityService to wait for generating a link
+   */
   @NotNull
   private int waitForS3LinksSeconds = 1;
 
+  /**
+   * Links to logs will expire after given number of milliseconds.
+   * A new link is generated for every /logs API call.
+   */
```
wsorenson: This is a useful comment.

SingularityService/src/main/java/com/hubspot/singularity/config/S3Configuration.java
```diff
@@ -9,18 +9,36 @@
...
   @NotNull
   private long expireS3LinksAfterMillis = TimeUnit.DAYS.toMillis(1);
 
+  /**
+   * S3 Bucket that SingularityS3Uploader puts logs in.
```
wsorenson: This is probably not a useful comment.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -127,6 +133,11 @@ public SingularityExecutorConfiguration(
     this.logrotateCount = logrotateCount;
     this.logrotateMaxageDays = logrotateMaxageDays;
     this.logrotateDateformat = logrotateDateformat;
+    if ((s3FilesToBackup != null) && (s3FilesToBackup.trim().length() > 0)) {
+      this.s3FilesToBackup = s3FilesToBackup.split(",");
+    } else {
+      this.s3FilesToBackup = new String[0];
+    }
```
tpetr: using a `Splitter` would probably be cleaner here (logrotateExtrasFiles is using this same pattern, maybe we should change that too)

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -49,6 +49,11 @@
   private final String logrotateExtrasDateformat;
   private final String[] logrotateExtrasFiles;
 
+  /**
+   * Extra files to backup to S3 besides the service log.
+   */
+  private final String[] s3FilesToBackup;
```
tpetr: should probably be explicit and call it `additionalS3FilesToBackup`, since we're doing the service.log file by default

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfigurationLoader.java
```diff
@@ -47,6 +47,7 @@
   public static final String TAIL_LOG_LINES_TO_SAVE = "executor.service.log.tail.lines.to.save";
   public static final String TAIL_LOG_FILENAME = "executor.service.log.tail.file.name";
 
+  public static final String S3_FILES_TO_BACKUP = "executor.s3.uploader.extras.files";
```
tpetr: consider `executor.s3.uploader.additional.files`

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskLogManager.java
```diff
@@ -148,8 +151,17 @@ private boolean writeTailMetadata(boolean finished) {
...
+   * The matching files are caught by the S3 Uploader and pushed to S3.
+   * @return file glob String.
+   */
   private String getS3Glob() {
-    return String.format("%s*.gz*", taskDefinition.getServiceLogOutPath().getFileName());
+    List<String> fileNames = new LinkedList<>();
+    fileNames.add(taskDefinition.getServiceLogOutPath().getFileName().toString());
+    fileNames.addAll(Arrays.asList(configuration.getS3FilesToBackup()));
+
+    return String.format("{%s}*.gz*", Joiner.on(",").join(fileNames));
```
tpetr: just thinking out loud here -- i was originally going to suggest doing something like `Joiner.on(",").join(Iterables.concat(Collections.emptyList(taskDefinition.getServiceLogOutPath().getFileName().toString()), Arrays.asList(configuration.getS3FilesToBackup())))` to prevent creating an extra list here, but that's kinda gross / harder to understand

SingularityExecutor/src/main/resources/logrotate.conf.hbs
tpetr: this is unnecessary, and broke logrotate in hs_staging -- turns out for these, the current directory is already the `rotateDirectory`, so setting this makes it attempt .../logs/logs

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfigurationLoader.java
wolfd: I was copying the format of https://github.com/HubSpot/Singularity/blob/persist_more_logs/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfigurationLoader.java#L45 but I agree that additional is a better word. Which would you prefer?

PR #480, opened 2015-03-13
Better tailing error when a file has not been generated yet

benrlodge:
This will replace the more "harsh" error message:

screenshot 2015-03-13 16 02 50

With a "calmer" message: (open to wording tweaks)

screenshot 2015-03-13 15 47 40

Aside from just being less harsh, the additional purpose of this PR is to support a separate PR in the works that will redirect a user to a tailing file after they run a task, when that file may or may not have be generated by the time they arrive...so...this will provide them with a better waiting area if they are waiting on their file to load.

Right now this only handles a bad request (400) and more specifically, when a file/directory is not found. However, I've set things up so we could pass various types of error messages in the future.

How things are setup -
- If a 400 error occurs, we search the returned message to see if it contains "does not have a directory yet", and from there bypass the standard error message. (A more specific code returned from the request may be cleaner, but I'm working with what I have with this current setup, which isn't terrible.)
- In the collection, an event is triggered if we hit this error, with the relevant info passed along with it to the view which is listening, and that info is then passed in to the template to show/hide the message.
- Last an interval is set polling for the initial data to be loaded, at which point the interval and message are removed.

@tpetr

### DIFF
--- a/SingularityUI/app/application.coffee
+++ b/SingularityUI/app/application.coffee
@@ -101,6 +101,20 @@ class Application
             return if @blurred and jqxhr.statusText is 'timeout'
 
             url = settings.url.replace(config.appRoot, '')
+            
+            try
+                serverMessage = JSON.parse(jqxhr.responseText).message or jqxhr.responseText
+            catch
+                serverMessage = jqxhr.responseText
+
+            serverMessage = _.escape serverMessage
+            
+            # Check if the file/directory doesn't exist yet so we
+            # can handle the error with somee style on tailing pages           
+            directoryPatt = new RegExp("does not have a directory yet")
+            hasNoDirectory = directoryPatt.test serverMessage
+            
+            return if jqxhr.status is 400 and hasNoDirectory
 
             if jqxhr.status is 502
                 Messenger().info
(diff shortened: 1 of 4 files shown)


SingularityUI/app/application.coffee
```diff
@@ -101,6 +101,20 @@ class Application
...
+                serverMessage = jqxhr.responseText
+
+            serverMessage = _.escape serverMessage
+            
+            # Check if the file/directory doesn't exist yet so we
+            # can handle the error with somee style on tailing pages           
+            directoryPatt = new RegExp("does not have a directory yet")
+            hasNoDirectory = directoryPatt.test serverMessage
+            
+            return if jqxhr.status is 400 and hasNoDirectory
```
timmfin: Could you use `app.caughtError()` so that you don't need to duplicate this dirrectoryPattern logic in both places (here and in `LogLines.coffee`?)

timmfin:
Nice, much better message :)

SingularityUI/app/application.coffee
benrlodge: cool thanks, didn't see that feature.

tpetr:
closing this in favor of #502

PR #481, opened 2015-03-13
remove invalid auto value that may be the root of issues on qa

benrlodge:
Attempting to fix the padding issue with the action buttons that is occurring on QA by removing the invalid padding auto value.

screenshot 2015-03-13 17 29 09

@tpetr


tpetr:
aha, i was wondering what happened... thanks!

PR #483, opened 2015-03-17
Deploy history JSON on Request page should only show deploy info

benrlodge:
...not entire task info
@tpetr


tpetr:
LGTM, thanks

PR #493, opened 2015-03-30
add LB state and healthchecks to task page

benrlodge:
@tpetr 

preview: 
screenshot 2015-03-30 09 47 55

### DIFF
--- a/SingularityUI/app/templates/taskDetail/taskHealthChecks.hbs
+++ b/SingularityUI/app/templates/taskDetail/taskHealthChecks.hbs
@@ -1,47 +1,49 @@
 {{#if synced}}
-    <div class="page-header">
-        <h2>Health Checks</h2>
-    </div>
-    {{#ifGT data.healthcheckResults.length 0}}
-        <div class="table-container">
-            <table class="table table-striped">
-                <thead>
-                    <tr>
-                        <th>Time Occured</th>
-                        <th>Duration</th>
-                        <th>Status</th>
-                        <th>errorMessage</th>
-                    </tr>
-                </thead>
-                <tbody>
-                    {{#each data.healthcheckResults}}
+    {{#ifGT data.task.taskRequest.deploy.healthcheckUri.length 0}}
+        <div class="page-header">
+            <h2>Healthchecks</h2>
+        </div>
+        {{#ifGT data.healthcheckResults.length 0}}
+            <div class="table-container">
+                <table class="table table-striped">
+                    <thead>
                         <tr>
-                            <td>
-                               {{timestampFormatted timestamp}}
-                            </td>
-                            <td>
-                               {{durationMillis}}{{#if durationMillis}}ms{{/if}}
-                            </td>
-                            <td>
-                               {{statusCode}}
-                            </td>
-                            <td> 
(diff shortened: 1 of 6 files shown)


tpetr:
Looks good overall, just a couple nitpicks:
- might want to rename `Load Balancer State` to `Load Balancer Updates`, since it's table of the updates to the load balancer(s)
- can we consolidate on calling them `Healthchecks` ?
- only show the LB table if `task.taskRequest.loadBalanced` is `true`
- only show the healthchecks table if `task.deploy.healthcheckUri` is a non-empty string
- if `task.deploy.skipHealthchecksOnDeploy` is `true`, display a message like `Healthchecks are disabled for this deploy` with a treatment similar to the `No Health Checks` one.

benrlodge:
Cool - couple quick questions/clarifications

*if task.deploy.skipHealthchecksOnDeploy is true...*
 - I only see this value in the metadata - `deploy.metadata.deploy` as a JSON string - should I parse this, or will there be an actual `task.deploy.healthcheckUri` property if it's true?
screenshot 2015-03-30 11 04 29


*only show the healthchecks table if task.deploy.healthcheckUri is a non-empty string*
 - can I assume this value will always be a string (i.e. can just check the value's .length in hbs)

tpetr:
- yeah, since `healthcheckUri` is an `Optional<String>`, it won't show up in the JSON if it's absent
- can't assume it'd always be a string, but you can coerce it into a boolean using `!!`

benrlodge:
sorry - for that first point, I meant `skipHealthchecksOnDeploy` (not healthcheckUri) doesn't have it's own property but is within a JSON string in `task.taskRequest.deploy.metadata.deploy`

tpetr:
ah sorry -- i misread what you said... yeah, both fields are optional, so the field may not show up in the JSON if they're absent. if `skipHealthchecksOnDeploy` is absent, assume it's `false`

benrlodge:
@tpetr, made these updates.

SingularityUI/app/templates/taskDetail/taskHealthChecks.hbs
```diff
@@ -0,0 +1,55 @@
...
+                                    </a>
+                                </td>
+                            </tr>
+                        {{/each}}
+                    </tbody>
+                </table>
+            </div>
+        {{else}}
+            <div class="empty-table-message">
+                No Health Checks
```
tpetr: could we make this `No Healthchecks` ?

SingularityUI/app/templates/taskDetail/taskHealthChecks.hbs
```diff
@@ -0,0 +1,55 @@
+{{#if synced}}
+    {{#ifCoercedBooleanIsTrue data.task.taskRequest.deploy.healthcheckUri}}
```
tpetr: i think we can get away with just `{{#if data.task.taskRequest.deploy.healthcheckUri}}` here -- is there a situation where it doesnt?

benrlodge: Yea we can get away with that.

PR #496, opened 2015-03-30
don't continually bounce the view to the bottom of the log while tailing if you've scrolled up

benrlodge:
@tpetr - the data was being refetched during the global refresh, causing the view to start tailing again, moved it out to only during initialization.


tpetr:
aha, nice catch

PR #497, opened 2015-03-31
match log type/name for s3 logs

ssalinas:
@tpetr should compliment the new logging and only download logs whose original name matches the `-l` glob
note: will continue to download all logs in the old format


tpetr:
this implementation is a little too HubSpot-specific -- could you have it accept the same formatting strings that are defined in the runnable configs?

ssalinas:
Updated this to take in the file pattern same as executor.s3.uploader.pattern. We can use it more extensively if we need. For now, if filename is in the pattern, then try to match if `-l` is set. If filename isn't in the pattern string, match and download everything unless `--no-name-fetch-off` is set (then skip s3 logs)
@tpetr

tpetr:
LGTM

PR #500, opened 2015-04-02
less lb cleanup, more lb updates, noop should be load_balance_failed

ssalinas:
@tpetr a few of the lb improvements separated out from baragon client PR

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
@@ -462,7 +462,6 @@ public class SingularityCleaner {
 
     switch (lbRemoveUpdate.getLoadBalancerState()) {
       case SUCCESS:
-      case INVALID_REQUEST_NOOP:
         return CheckLBState.DONE;
       case FAILED:
       case CANCELED:
@@ -474,6 +473,11 @@ public class SingularityCleaner {
       case CANCELING:
       case WAITING:
         LOG.trace("Waiting on LB cleanup request {} in state {}", loadBalancerRequestId, lbRemoveUpdate.getLoadBalancerState());
+        break;
+      case INVALID_REQUEST_NOOP:
+        exceptionNotifier.notify(String.format("LB removal failed for %s", lbAddUpdate.get().getLoadBalancerRequestId().toString()),
+          ImmutableMap.of("state", lbRemoveUpdate.getLoadBalancerState().name(), "loadBalancerRequestId", loadBalancerRequestId.toString(), "addUpdate", lbAddUpdate.get().toString()));
+        return CheckLBState.LOAD_BALANCE_FAILED;
     }
 
     return CheckLBState.WAITING;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
```diff
@@ -474,6 +473,11 @@ private CheckLBState checkLbState(SingularityTaskId taskId) {
       case CANCELING:
       case WAITING:
         LOG.trace("Waiting on LB cleanup request {} in state {}", loadBalancerRequestId, lbRemoveUpdate.getLoadBalancerState());
+        break;
+      case INVALID_REQUEST_NOOP:
+        exceptionNotifier.notify(String.format("LB removal failed for %s", lbAddUpdate.get().getLoadBalancerRequestId().toString()),
+          ImmutableMap.of("state", lbRemoveUpdate.getLoadBalancerState().name(), "loadBalancerRequestId", loadBalancerRequestId.toString(), "addUpdate", lbAddUpdate.get().toString()));
+        return CheckLBState.LOAD_BALANCE_FAILED;
     }
```
wsorenson: does this retry forever?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -471,7 +471,9 @@ public void handleCompletedTask(Optional<SingularityTask> task, SingularityTaskI
       stateCache.getActiveTaskIds().remove(taskId);
     }
 
-    taskManager.createLBCleanupTask(taskId);
+    if (task.isPresent() && task.get().getTaskRequest().getRequest().isLoadBalanced()) {
+      taskManager.createLBCleanupTask(taskId);
+    }
```
wsorenson: Is this an optimization? Is it 100% safe? What if the task isn't present but it's being load balanced due to a data corruption issue in Singularity?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
ssalinas: only `WAITING` and `RETRY` will continue to try again. `LOAD_BALANCE_FAILED` goes to the obsolete/ignored list since the removal failed. Same end result as having it go to `DONE` except that this way we'll get a sentry telling us why it wasn't removed properly

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
ssalinas: This was to try and cut down on non-load-balanced tasks in the lb cleanup queue, especially since we are now exposing more of that data in the ui. Might be a good idea to still add it in the case where the task isn't present, I'll fix that.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
wsorenson: :+1:

ssalinas:
@wsorenson fixed

PR #502, opened 2015-04-06
Better tailing error when a file has not been generated yet

tpetr:
(reopening #480 against `master` instead of `hs_staging`)
- - -
This will replace the more "harsh" error message:

screenshot 2015-03-13 16 02 50

With a "calmer" message: (open to wording tweaks)

screenshot 2015-03-13 15 47 40

Aside from just being less harsh, the additional purpose of this PR is to support a separate PR in the works that will redirect a user to a tailing file after they run a task, when that file may or may not have be generated by the time they arrive...so...this will provide them with a better waiting area if they are waiting on their file to load.

Right now this only handles a bad request (400) and more specifically, when a file/directory is not found. However, I've set things up so we could pass various types of error messages in the future.

How things are setup -
- If a 400 error occurs, we search the returned message to see if it contains "does not have a directory yet", and from there bypass the standard error message. (A more specific code returned from the request may be cleaner, but I'm working with what I have with this current setup, which isn't terrible.)
- In the collection, an event is triggered if we hit this error, with the relevant info passed along with it to the view which is listening, and that info is then passed in to the template to show/hide the message.
- Last an interval is set polling for the initial data to be loaded, at which point the interval and message are removed.

@tpetr

### DIFF
--- a/SingularityUI/app/application.coffee
+++ b/SingularityUI/app/application.coffee
@@ -101,6 +101,20 @@ class Application
             return if @blurred and jqxhr.statusText is 'timeout'
 
             url = settings.url.replace(config.appRoot, '')
+            
+            try
+                serverMessage = JSON.parse(jqxhr.responseText).message or jqxhr.responseText
+            catch
+                serverMessage = jqxhr.responseText
+
+            serverMessage = _.escape serverMessage
+            
+            # Check if the file/directory doesn't exist yet so we
+            # can handle the error with somee style on tailing pages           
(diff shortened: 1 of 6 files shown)


tpetr:
@benrlodge ^ commit above changes errors to be in JSON form, let me know if you need anything tweaked for it

tpetr:
actually, i should have looked closer at my own code! yeah, we should just refactor `parse()` in to something like `getRenderData()`, and have the view's `render()` method call that instead of `toJSON()`

PR #504, opened 2015-04-06
automatically go to tailing service.log (or other log file) after clicking 'run now'

tpetr:
(reopening #475 against `master` instead of `hs_staging`)


Eseb:
@benrlodge I'm not a fan of how you're passing 2 extra collections `views/request` for the sake of a bunch of logic I don't think should even be in there. The 80-ish lines you're adding to the view don't have anything to do with that part of the UI.

What I'd do here is put this stuff in its own class, maybe call it `lib/TaskPoller` or something like that. Then you can just use that inside of `views/requests` while keeping all of the logic in its own neat little package that can be re-used if necessary. Speaking of re-using, does all of this polling stuff work if you use the requests table run-now button?

I haven't been following SUI development over the past few months so sorry if I'm sticking my nose somewhere it's not welcome or if I'm misreading what you're doing! I got an email notification related to this and thought I'd have a look at it for the craic.

cc @tpetr @timmfin

SingularityUI/app/styles/colors.styl
```diff
@@ -2,7 +2,7 @@ $linkBlue = #08c
 
 $beige = #f3efe0
 $grey = #444444
-
+$gray-light = #dadada
```
Eseb: For the sake of consistency maybe use the [correct](http://en.wikipedia.org/wiki/Grey) (joke (not really (well, maybe a little))) spelling here.

SingularityUI/app/views/request.coffee
```diff
@@ -97,4 +106,84 @@ class RequestView extends View
...
+        vex.dialog.alert
+            overlayClosesOnClick: false
+            message: """
+                <h3>Launching</h3>
+                <ol class="auto-tail-checklist">
+                    <li class="auto-tail-task-start">Waiting for task to launch</li>
+                    <li class="auto-tail-file-exists">Waiting for <code>#{ @autoTailFilename }</code> to exist</li>
+                </ol>
+                <div class='page-loader centered cushy'></div>
+            """
```
Eseb: I'm not sure if this is something you were intending on later or not but these should be in templates.

benrlodge:
@tpetr I moved the auto tailing logic out of the `request` view and into a reusable `lib/AutoTailer` class. I also included this class in the `views/requests` table view as well

PR #505, opened 2015-04-06
docker support for SingularityExecutor

ssalinas:
@tpetr an initial draft of docker support within the Singularity Executor. If container info is present and type is docker, it will render a different template to runner.sh that translates the normal routine into docker env vars, ports, and volumes. The new template script will also handle signals and run docker stop rather than killing pids and hoping the container goes with it


SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -212,8 +212,9 @@ private void checkDocker(SingularityDeploy deploy) {
       final int numPorts = deploy.getResources().get().getNumPorts();
 
       if (!dockerInfo.getPortMappings().isEmpty()) {
-        checkBadRequest(dockerInfo.getNetwork().or(Protos.ContainerInfo.DockerInfo.Network.HOST) == Protos.ContainerInfo.DockerInfo.Network.BRIDGE,
-            "Docker networking type must be BRIDGE if port mappings are set");
+        checkBadRequest(dockerInfo.getNetwork().or(Protos.ContainerInfo.DockerInfo.Network.HOST) == Protos.ContainerInfo.DockerInfo.Network.BRIDGE ||
+            dockerInfo.getNetwork().or(Protos.ContainerInfo.DockerInfo.Network.HOST) == Protos.ContainerInfo.DockerInfo.Network.HOST,
```
ssalinas: for clarity here, you can do port mapping when network is host/bridge, we were failing before when anyone asked for port mappings and network was host

SingularityExecutor/src/main/resources/docker.sh.hbs
```diff
@@ -0,0 +1,126 @@
...
+while true; do
+  if [ "$running" = "1" ]; then
+    sleep 1
+  else
+    break
+  fi
+done
+
+exit_code=`sudo -E -u {{{ runContext.user }}} docker wait "$cid"`
+exit "$exit_code"
```
tpetr: just curious -- how closely does this resemble the mesos executor? could you link to it?

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
tpetr: port mappings in host mode is a no-op, though, right?

ssalinas: Ah, you are correct on that,  on closer read I misinterpreted the documentation, I'll fix this

SingularityExecutor/src/main/resources/docker.sh.hbs
ssalinas: Mesos code is here
https://github.com/apache/mesos/blob/6b675dd1b2d8a392b1c6600266aacacf1378ff4e/src/slave/containerizer/docker.hpp
The code above was based more on the old docker-exec project so as not to put too much extra junk in the executor. The only difference I think I still need to work on is making sure cpu and memory usage get picked up appropriately. The mesos code separately finds the docker processes and marks them as the ones to watch for resource usage. I'm hoping with a few edits I can make it so the docker process is a child process of the wrapper above so the resources are monitored correctly.

ssalinas:
Ok, think this is ready for a review. Some things I've implemented:
- start script will move docker root pid and children to the cgroup for the mesos executor. So, it will currently only get correct metrics with cgroup isolation in use.
- The script will stop and remove the container, but if for some reason the script gets into a bad state, the SingularityExecutorTaskCleanup will also ensure that the container is removed
- the sandbox/(taskid) folder is linked to /mnt/mesos/sandbox in the container
- all environment variables are correctly passed to the container
- the thread checker will monitor the root pid of the docker container process, not the executor process to get a more accurate thread count metric
@tpetr

ssalinas:
A few more things added:
- TaskCleanup will also attempt to remove the stopped container in case the bash script died too abruptly
- the working directory inside the container will be set to the taskAppDirectory on the mounted volume
- added another step to the vagrant provision so SingularityExecutor is ready to use by default

ssalinas:
@tpetr 
I found that tasks with large images to pull were registering as running before they were truly running since the image pull was part of the wrapper script. I introduced the spotify docker-client to help with images rather than managing shell processes, and moved the image pull to just before the artifact fetch and file staging so that it occurs in the task starting period

ssalinas:
/fyi @tpetr I updated the [container docs](https://github.com/HubSpot/Singularity/blob/docker_execute/Docs/containers.md) to reflect the addition of the SingularityExecutor docker support as well as some more detail on using the default mesos-docker-containerizer

ssalinas:
:tada: :balloon: :dancer: :smiley:

tpetr:
:leaves: 

:whale:

PR #506, opened 2015-04-06
Tailing permalinks

benrlodge:
Creates deep linking for each line in tailing files; when a line is clicked, it highlights the line, and updates the url with that line's offset (similar to the action of clicking the line number in github files).

@tpetr

### DIFF
--- a/SingularityUI/app/views/tail.coffee
+++ b/SingularityUI/app/views/tail.coffee
@@ -12,6 +12,7 @@ class TailView extends View
         _.extend super,
             'click .tail-top-button': 'goToTop'
             'click .tail-bottom-button': 'goToBottom'
+            'click .copy-link' : 'copyLink'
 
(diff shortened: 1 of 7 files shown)


SingularityUI/app/views/tail.coffee
```diff
@@ -57,10 +62,14 @@ class TailView extends View
         # Well, or just render them if we're starting fresh
         $firstLine = @$linesWrapper.find '.line:first-child'
         $lastLine  = @$linesWrapper.find '.line:last-child'
+        url = window.location.host + window.location.pathname
```
tpetr: i dont think this is necessary for `@linesTemplate` -- `{{url}}` isnt in the template anywhere

benrlodge: right thanks, that was when i was going down the path of copying the link.

PR #507, opened 2015-04-06
return SingularityPendingRequestParent when immediately scheduling a request

tpetr:
Surfacing the `SingularityPendingRequest` data when scheduling a request to run immediately will make some UI things much easier.


tpetr:
@wsorenson turns out the timestamp in `SingularityPendingTaskId` and `SingularityPendingRequest` for on-demand and run now'd scheduled requests don't match, which makes my first commit useless... the 2nd commit adjusts scheduling to use the pending request's timestamp for these cases (havent merged this commit to any other branch yet)

wsorenson:
Okay. Makes sense.

tpetr:
(related to #504)

tpetr:
turns out using the pending request's timestamp isn't good enough... the task ID's timestamp is of when it was launched, not when it should have run. i'm not wild about making a change as big as that, so i'm just going to revert this 2nd commit

PR #508, opened 2015-04-06
Add log rotation for /var/log/singularity/singularity.log

eherot:


cookbook/recipes/logrotate.rb
```diff
@@ -0,0 +1,11 @@
+logrotate_app 'singularity' do
+  path          node[:singularity][:log_file]
+  frequency     'daily'
```
tpetr: huge nitpick, but could we make this an attribute?

eherot:
OK, frequency has been "attribute-ized"

tpetr:
danke

PR #509, opened 2015-04-07
set container info in ExecutorInfo, if using custom executor

tpetr:
@ssalinas i think the reason why docker wasn't working for you originally was because we were only setting `ContainerInfo` on the `TaskInfo` proto, but I believe we need to set it on `ExecutorInfo` when using a custom executor. can you give this a try and tell me how it works?


ssalinas:
Interesting, this actually ended up running `/usr/local/bin/singularity-executor` from the `customExecutorCmd` field inside the container after pulling it. This seems to trigger one level up the chain from all of our custom executor code

ssalinas:
Did some more fooling with this and I can't get it to hit the singularity executor at all. It always bypasses that and executes the `customExecutorCmd` within the container

ssalinas:
closing this in favor of the other docker PR

PR #515, opened 2015-04-08
Adding shell commands

wsorenson:
Adds ability to run certain commands from UI and passed over the framework to the executor.

Executor impl. WIP;

Probably will need a more sophisticated data model.

Needs some UI work.

### DIFF
--- a/SingularityUI/app/styles/table.styl
+++ b/SingularityUI/app/styles/table.styl
@@ -117,6 +117,3 @@ th #schedule
 
 #empty-table
     margin-bottom 20px
-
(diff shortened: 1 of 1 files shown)


ssalinas:
@wsorenson I've been seeing a number of `TASK_LOST` items with `Abnormal executor termination` or `Executor terminated` messages. Specifically when mixing this with some of the docker executor code. Haven't been able to trace it back to a specific problem area just yet, but wanted you to know

eabbott:
(golem)

PR #518, opened 2015-04-10
add timestamp to columns in slave view

benrlodge:
@tpetr


tpetr:
what do you think about moving the timestamp column to the right of `State` and renaming it to `Since`? LGTM besides that

benrlodge:
yea sounds good

PR #517, opened 2015-04-10
Add task detail

benrlodge:
- Moves healthchecks table to bottom of task detail page, collapsed by default
- Reordered healthchecks table to newest --> oldest, instead of oldest --> newest
- When the deploy associated with the task is PENDING (meaning a deploy is in progress), additionally display the current healthcheck at the top of the task detail page (underneath the well) with a link to the full healthcheck history

@tpetr - this was opened from hs_staging (otherwise I'd open against master)

PR #521, opened 2015-04-14
Add task detail

tpetr:
(reopening #517 against `master` instead of `hs_staging`)

Moves healthchecks table to bottom of task detail page, collapsed by default
Reordered healthchecks table to newest --> oldest, instead of oldest --> newest
When the deploy associated with the task is PENDING (meaning a deploy is in progress), additionally display the current healthcheck at the top of the task detail page (underneath the well) with a link to the full healthcheck history


tpetr:
this branch is all screwy, see #526 instead

PR #522, opened 2015-04-14
Fix rendering dashboard when no user is set

Eseb:
When you open up SUI with no username set it derps and throws an exception because it tries to call `undefined.split`. This fixes that. It's been broken since October ([0eb428](https://github.com/HubSpot/Singularity/commit/0eb428f2cb2001b8102ab36bb0caba3ffce1a737)).

~~Also just because I had to change this, @ssalinas and @wsorenson please use `is` for CoffeeScript comparisons, not `==`. `is` translates to the more strict `===` operator in JS.~~

@tpetr @benrlodge @gchomatas

### DIFF
--- a/SingularityUI/app/views/dashboard.coffee
+++ b/SingularityUI/app/views/dashboard.coffee
@@ -18,38 +18,38 @@ class DashboardView extends View
     render: =>
         deployUser = app.user.get 'deployUser'
 
-        partials = 
+        partials =
             partials:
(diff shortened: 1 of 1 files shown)


benrlodge:
CoffeeScript also compiles `==` into `===`. That said, `is` is used more frequently in the code base so I agree for consistency's sake we stick with `is`

Eseb:
@benrlodge My bad, you're right :rabbit:

PR #523, opened 2015-04-15
Add client-side pagination option for tables.

benrlodge:
We currently have server-side pagination for some endpoints, but not all (e.g. S3 Logs). This adds the ability to use client-side pagination.

- Created a new collection for client-side specific pagination. Extends the server-side `PaginableCollection`
- Added logic in the `expandableTableSubview`, same view used by server-side `PaginableCollection`
- Added this new option to the S3 log table in task view

@tpetr


benrlodge:
Took another approach to this, moved clientside/serverside specific logic to separate view classes that extend `ExpandableTableSubview`

ssalinas:
closing this since client side pagination was implemented (differently) in the deploy detail page. We can redo the s3 logs in that style for consistency

PR #529, opened 2015-04-22
(WIP) Introducing react to singularity

benrlodge:
Taking a first stab at introducing react to singularity. 

Changes so far include adding react dependencies, and replacing the dashboard with react components. This still continues to use backbone's models/collections.

If you pull this branch, you'll need to `bower install` and `npm install` to get new react dependencies.


tpetr:
^ ignore that reference, wrong number

Eseb:
Exciting stuff! Can I ask why you're putting the react stuff into its own dir? I don't really see why you cannot have them all together (the controllers, too). I'm not necessarily saying it's wrong, I'm just curious to know what the thinking behind this is. On the email app we also have a mix of `Backbone.View` and `React.Component`. We have a structure that's the equivalent of:

```
coffee/
    models/
    collections/
    views/
    controllers/
    components/ <- same place
```

A few nitpicky comments regarding consistency:
- Mix of 2 and 4 width tabs
- `@props` in some paces, `this.props` in others, sometimes on the same line
- `render: -> return (<div></div>)` in some places, `render: -> <div></div>` in others

benrlodge:
@Eseb this was experimental, so I was keeping everything in its own folder while I was thinking things out. I agree though, I was thinking this as well, I think they the react views/controllers can live together with backbone, and the components can move in with the rest of those directories.

tpetr:
let's revisit this when we have more bandwidth to focus on it -- i'll keep the branch hanging around

PR #430, opened 2015-01-30
Status page (WIP)

ssalinas:
New PR for new status page, going to hs_staging now instead of master

PR #530, opened 2015-04-22
Status page

tpetr:
opening #430 against `master` instead of `hs_staging`


tpetr:
a few extra things:
- remove the links from the Racks and Slaves headers, since all the links below go to the same page
- add anchors to the different rack and slave types (i.e. #active #decomissioning, etc) and update the status page links to point to the right thing
- have the overdue number link to the scheduled task page

benrlodge:
cool, that's all set.

kwm4385:
Fix for numbers in the column on update: https://github.com/HubSpot/Singularity/pull/720

PR #532, opened 2015-04-23
check for no data in beforeClose

benrlodge:
I noticed in QA that you will get an error when trying to cancel out of the updated run-now prompt; because we added the beforeClose method, we need to test if the data is empty in that method, since it’s called after the callback. @tpetr


benrlodge:
this will be rolled into #504

PR #534, opened 2015-04-23
don't 404 deleted requests, show a message instead

benrlodge:
If a request no longer exists, instead of showing a 404, the request detail page will continue to show, and a message is displayed that it has been deleted. 

I copied the look of the existing state/request-type section for requests still running, but used a yellow alert instead. 

Thoughts?
@tpetr @wsorenson 

screenshot 2015-04-23 17 00 55

*Note - The header subview would have been a good section for this, but that expects a model that no longer exists for the request (hence the 404), and uses the simpleSubview so only expects a model or collection, so it seemed simpler to just add a new subview.*

### DIFF
--- a/SingularityUI/app/controllers/RequestDetail.coffee
+++ b/SingularityUI/app/controllers/RequestDetail.coffee
@@ -17,6 +17,7 @@ class RequestDetailController extends Controller
 
     templates:
         header:         require '../templates/requestDetail/requestHeader'
+        taskHistoryMsg: require '../templates/requestDetail/taskHistoryMsg'
         stats:          require '../templates/requestDetail/requestStats'
 
         activeTasks:    require '../templates/requestDetail/requestActiveTasks'
@@ -55,6 +56,13 @@ class RequestDetailController extends Controller
         @subviews.header = new SimpleSubview
             model:      @models.request
             template:   @templates.header
+        
+        # would have used header subview for this info,
+        # but header expects a request model that
+        # no longer exists if a request is deleted
+        @subviews.taskHistoryMsg = new SimpleSubview
+            collection: @collections.requestHistory
+            template:   @templates.taskHistoryMsg
 
         @subviews.stats = new SimpleSubview
             model:      @models.activeDeployStats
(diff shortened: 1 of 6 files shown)


SingularityUI/app/templates/requestDetail/taskHistoryMsg.hbs
```diff
@@ -0,0 +1,23 @@
...
+                    {{humanizeText data.[0].request.requestType }} 
+                </span>                
+            </h4>
+            <h2>
+                {{ data.[0].request.id }} 
+            </h2>
+        </div>
+    </div>
+    <div class='alert alert-warning'>
+        <strong>Note: </strong> This request was deleted {{timestampFromNow data.[0].createdAt}}. All active and scheduled tasks will not run again unless it is reposted to Singularity.
```
tpetr: Could we include the user too? (assuming `data.[0].user` is set)

SingularityUI/app/controllers/RequestDetail.coffee
```diff
@@ -17,6 +17,7 @@ class RequestDetailController extends Controller
 
     templates:
         header:         require '../templates/requestDetail/requestHeader'
+        taskHistoryMsg: require '../templates/requestDetail/taskHistoryMsg'
```
tpetr: how come we're calling it `taskHistoryMsg` -- isn't this information about the request?

benrlodge: yep, that was an oversight, thanks

SingularityUI/app/templates/requestDetail/taskHistoryMsg.hbs
benrlodge: all set

tpetr:
i might not have made this clear the first time, sorry -- the `user` is optional, so we should probably wrap the `, by XXX` in an if block

LGTM once that's squared away

benrlodge:
my bad, i'll add that in. thanks.

SingularityUI/app/templates/requestDetail/requestHistoryMsg.hbs
```diff
@@ -0,0 +1,23 @@
...
+                    {{humanizeText data.[0].request.requestType }} 
+                </span>                
+            </h4>
+            <h2>
+                {{ data.[0].request.id }} 
+            </h2>
+        </div>
+    </div>
+    <div class='alert alert-warning'>
+        <strong>Note: </strong> This request was deleted {{timestampFromNow data.[0].createdAt}}{{#if data.[0].user}}, by {{data.[0].user}} {{/if}}. All active and scheduled tasks will not run again unless it is reposted to Singularity.
```
tpetr: can you fix the extra space between the username and the period?

benrlodge: all set

PR #535, opened 2015-04-27
Add maven install option

eherot:


cookbook/attributes/default.rb
```diff
@@ -10,7 +10,7 @@
   :user => 'singularity',
   :group => 'singularity',
   :git_ref => 'e2405eb5ca1a1ba006a89a27bdb3299433ae96d5',
-  :version => '0.4.2',
+  :version => '0.4.1',
```
tpetr: does `git_ref` need to be updated in response to this change? could we do one better and use a tag name instead of a legit SHA?

cookbook/files/default/migrations.sql
```diff
@@ -0,0 +1,64 @@
...
+  DROP COLUMN `pendingType`,
+  DROP COLUMN `createdAt`,
+  DROP COLUMN `directory`,
+  MODIFY `lastTaskStatus` VARCHAR(25) DEFAULT NULL,
+  MODIFY `updatedAt` TIMESTAMP NOT NULL DEFAULT '1971-01-01 00:00:01',
+  CHANGE `task` `bytes` BLOB NOT NULL,
+  DROP KEY `requestId`,
+  DROP KEY `requestId_3`;
+
+DROP TABLE `taskUpdates`;
```
tpetr: not wild about having multiple copies of this floating around... would a symlink work instead?

eherot: A symlink to what? The problem is that the only canonical source for this (that I know about) is the Singularity source code.

eherot: Oh you mean a symlink *within* the Git repo. I suppose we could make the other copy a symlink to this one. Let me try out some options here...

cookbook/attributes/default.rb
```diff
@@ -9,8 +9,8 @@
 default[:singularity] = {
   :user => 'singularity',
   :group => 'singularity',
-  :git_ref => 'e2405eb5ca1a1ba006a89a27bdb3299433ae96d5',
-  :version => '0.4.2',
+  :git_ref => 'Singularity-0.4.1',
```
eherot: @tpetr how's that?

cookbook/files/default/migrations.sql
```diff
@@ -0,0 +1 @@
+../mysql/migrations.sql
```
eherot: Testing now to see what Chef does with this...

cookbook/attributes/default.rb
tpetr: looks good, thanks

PR #543, opened 2015-05-11
multiplart upload for items over 5GB

ssalinas:
@tpetr Upload items greater that 5GB using multipart in 100MB chunks


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3Uploader.java
```diff
@@ -32,6 +34,9 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(SingularityS3Uploader.class);
 
+  private final long maxSingleUploadSizeBytes = 5368709120L;
+  private final long uploadPartSize = 1048576L;
```
tpetr: can you move these values unto the config?

ssalinas: :+1:

PR #545, opened 2015-05-12
add s3 logs to request page

benrlodge:
Adds a new table on the request page for all its tasks s3 logs.

we should merge in https://github.com/HubSpot/Singularity/pull/523 before rolling this out though as we'll run into that same issue here.

@tpetr


benrlodge:
this is all set, merge issue with #523 mentioned ^ is not longer an issue.

tpetr:
closing this in favor of moving towards a more consolidated task log view

PR #546, opened 2015-05-12
Include cleaning tasks when checking for SEPARATE placement

ssalinas:
Allows us to do a bounce without having the cleaning and starting tasks on the same host


wsorenson:
Hmm, this change also makes them affect rack placements, which doesn't seem right.

wsorenson:
It also changes the counting mechanism. 

Also, any change like this should include specific tests showing the functionality we've added now works as designed, and that other functionality we specifically decided not to change works as designed.

ssalinas:
@wsorenson updated so that it should not impact rack sensitive or `OPTIMISTIC` placement

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -153,7 +161,7 @@ public SlaveMatchState doesOfferMatch(Protos.Offer offer, SingularityTaskRequest
 
     switch (slavePlacement) {
       case SEPARATE:
-        if (numOnSlave > 0) {
+        if (numOnSlave > 0 || cleaningOnSlave > 0) {
           LOG.trace("Rejecting SEPARATE task {} from slave {} ({}) due to numOnSlave {}", taskRequest.getRequest().getId(), slaveId, host, numOnSlave);
```
tpetr: might want to surface `cleaningOnSlave` in the log here too

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -129,13 +130,20 @@ public SlaveMatchState doesOfferMatch(Protos.Offer offer, SingularityTaskRequest
     final int numDesiredInstances = taskRequest.getRequest().getInstancesSafe();
     double numOnRack = 0;
     double numOnSlave = 0;
+    double cleaningOnSlave = 0;
```
tpetr: huge nitpick, i know, but might be best to call this `numCleaningOnSlave`

ssalinas:
@tpetr 
Want you opinion on this. This SEPARATE placement currently wasn't working for a new deploy because we are only checking against tasks from the same deploy (which is why it worked fine for a bounce). I've altered the code at the moment to check against all tasks for the request and additionally keep track of other deploys still on the same machine (only taking it into account for separate placement).

Do you think that warrants another slave placement category or do you think separate should work this way and imply that essentially no two tasks will ever come in contact? I bring that up because you would then need a count of slaves double your instances to deploy (granted it is that way now with bounces)

stevenschlansker:
:+1: we will totally use this feature

Another option, if we do not want different `SEPARATE_BY_REQUEST` vs `SEPARATE_BY_DEPLOY`, would be a different configuration option `placementConsidersTasksFromAllDeploys` or the like.  Then you could have both flavors for `SEPARATE` as well as both flavors for `OPTIMISTIC`.

ssalinas:
We would want to be able to use `SEPARATE_BY_REQUEST` and `SEPARATE_BY_DEPLOY` side by side for our puposes. I'm happy to add the other flavor for `OPTIMISTIC`, but I think I'd lean towards having everything available per request rather than a more global config setting.

stevenschlansker:
FWIW, adding a new Enum value to the task placement breaks old SingularityClient versions:

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Can not construct instance of com.hubspot.singularity.SlavePlacement from String value 'SEPARATE_BY_REQUEST': value not one of declared Enum instance names: [SEPARATE, OPTIMISTIC, GREEDY]
at [Source: org.jboss.netty.buffer.ChannelBufferInputStream@17f2b076; line: 1, column: 77424] (through reference chain: java.util.ArrayList[21])
```

This may be unavoidable -- but we may wish to improve the client by having the placement type fail deserialization gracefully.  Maybe making it an `Optional<SlavePlacement>` in case we don't know the placement strategy?

stevenschlansker:
Any idea of whether this PR will move forward?  We have been trying it out with success locally, but I'm worried to commit to it too much until it hits master.

ssalinas:
@tpetr this has been running fine for us in hs_stable, any problem merging to master?

ssalinas:
@stevenschlansker Would you have a use for adding BY_REQUEST/BY_DEPLOY for `OPTIMISTIC` placement as well? It would be easy to implement, I just couldn't think of a use case for it at the time since it's just a slightly more picky optimistic placement.

As for the enum piece, the field should already be an `Optional<SlavePlacement>`, falling back to the default placement strategy from the `SingularityConfiguration` if it isn't set.

/cc @tpetr

stevenschlansker:
Not really -- the only time we use this is if the instances actually cannot co-exist (need a 1-per-machine resource, for example) and as such `OPTIMISTIC` isn't helpful anyway.

Re: enum, I think it is inappropriate to default the client; I think it is more appropriate to have an "I don't know" response, rather than silently substituting an incorrect default.

PR #548, opened 2015-05-14
Remove cookbook & refer to new home

jeffbyrnes:
Now that it’s a new home, let’s send folks over there.

Attn: @tpetr @wsorenson


jeffbyrnes:
Lemme try this again…

PR #549, opened 2015-05-14
Remove cookbook, referring folks to new repository.

jeffbyrnes:
Now that it’s a new home, let’s send folks over there.

Attn: @tpetr @wsorenson


tpetr:
looks good, but I think this will break the Vagrantfile -- mind updating it to point to your new location?

jeffbyrnes:
I'm off on vacation for 2 weeks, so I'll pass off to @eherot.

ssalinas:
The vagrant issues will be taken care of in #553 which uses the public version of the cookbook. I'm going to go ahead and merge this as well as closing the other cookbook PRs

/fyi @tpetr @eherot @jeffbyrnes

Issue #550, opened 2015-05-17
Maven plugin error during "vagrant up test"


kdemarest:
Following the install instructions from

`https://github.com/HubSpot/Singularity/blob/master/Docs/vagrant.md`

While attempting to "vagrant up test" for local testing, on a fairly fresh machine, received:

```
==> test: Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.23:install-node-and-npm (install node and npm) on project SingularityService: The plugin com.github.eirslett:frontend-maven-plugin:0.0.23 requires Maven version 3.1.0 -> [Help 1]
```

Perhaps I am missing a certain vagrant plugin?

Here is the relevant part of the log, with a bit of preceding context.
```
==> test: [info] SingularityMesosClient ] project SingularityMesosClient
==> test: [info] SingularitySwagger ] project SingularitySwagger
==> test: [info] SingularityService ] project SingularityService
==> test: [info] SingularityRunnerBase ] project SingularityRunnerBase
==> test: [info] SingularityS3Base ] project SingularityS3Base
==> test: [info] SingularityClient ] project SingularityClient
==> test: [info] SingularityExecutor ] project SingularityExecutor
==> test: [info] SingularityExecutorCleanup ] project SingularityExecutorCleanup
==> test: [info] SingularityLogWatcher ] project SingularityLogWatcher
==> test: 
==> test: [info] SingularityS3Uploader ] project SingularityS3Uploader
==> test: [info] SingularityS3Downloader ] project SingularityS3Downloader
==> test: [info] SingularityOOMKiller ] project SingularityOOMKiller
==> test: [info] EmbedSingularityExample ] project EmbedSingularityExample
==> test: [INFO] 
==> test: [INFO] --- jacoco-maven-plugin:0.7.2.201409121644:prepare-agent (default) @ SingularityService ---
==> test: [INFO] argLine set to -javaagent:/home/vagrant/.m2/repository/org/jacoco/org.jacoco.agent/0.7.2.201409121644/org.jacoco.agent-0.7.2.201409121644-runtime.jar=destfile=/singularity/SingularityService/target/jacoco.exec
==> test: [INFO] 
==> test: ------------------------------------------------------------------------
==> test: [INFO] 
==> test: Reactor Summary:
==> test: [INFO] 
==> test: [INFO] 
==> test: Singularity ....................................... SUCCESS [47.106s]
==> test: 
[...]

PR #551, opened 2015-05-18
manually install maven 3.3.3 in provision-singularity.sh

tpetr:
Successfully ran `vagrant up test` -- just a workaround until we rebuild `singularity-develop-0.21.0` with the newer Maven installed. @kdemarest, can you confirm this works for you too?

re: #550

### DIFF
--- a/vagrant/provision-singularity.sh
+++ b/vagrant/provision-singularity.sh
@@ -72,7 +72,7 @@ EOF
 function build_singularity {
   # lame hack to install a recent mvn, thanks ubuntu...
   if [ ! -f /usr/share/apache-maven-3.3.3/bin/mvn ]; then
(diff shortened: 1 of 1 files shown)


vagrant/provision-singularity.sh
```diff
@@ -70,8 +70,15 @@ EOF
 }
 
 function build_singularity {
+  # lame hack to install a recent mvn, thanks ubuntu...
+  if [ ! -f /usr/share/apache-maven-3.3.3/bin/mvn ]; then
+    wget http://apache.spinellicreations.com/maven/maven-3/3.3.3/binaries/apache-maven-3.3.3-bin.zip -O /tmp/apache-maven-3.3.3-bin.zip
```
kdemarest: Might want to drop "-q" at the end. Slightly less ugly that way.

kdemarest:
Confirmed, this worked for me. The bower problem is still there. Will make a separate issue for it.

Issue #552, opened 2015-05-18
brunch missing during "vagrant up test"


kdemarest:
During a "vagrant up test", while building SingularityService, the module SingularityUI is built. In SingularityUI's package.json file, the postinstall script calls brunch but returns the following error:
```
==> test: [INFO] > bower install --allow-root && brunch build --production
==> test: [ERROR] sh: 1: brunch: not found
```
However, checking the package.json everything appears fine (see below) and when I `vagrant ssh` into the machine, the node_modules/brunch/bin/ directory contains brunch.

Here is the excerpt from the package.json for reference:

```
"bin": {
    "bower": "./node_modules/bower/bin/bower",
    "brunch": "./node_modules/brunch/bin/brunch"
  },
  "scripts": {
    "start": "brunch watch --server",
    "postinstall": "bower install --allow-root && brunch build --production"
  },
  "dependencies": {
    "bower": "~1.4.1",
    "brunch": "^1.7.14",
```

Here is the full output for reference:

```
==> test: Downloaded: https://repo.maven.apache.org/maven2/com/google/guava/guava/10.0.1/guava-10.0.1.jar (1467 KB at 290.1 KB/sec)
==> test: [INFO] Node v0.10.28 is already installed.
==> test: [INFO] Found NPM version 1.4.9
==> test: [INFO] 
==> test: [INFO] --- frontend-maven-plugin:0.0.23:npm (npm install) @ SingularityService ---
==> test: [INFO] Running 'npm install --color=false' in /singularity/SingularityUI
==> test: [INFO] 
==> test: [INFO] > SingularityUI@0.3.0 postinstall /singularity/SingularityUI
==> test: [INFO] > bower install --allow-root && brunch build --production
==> test: [INFO] 
==> test: [ERROR] sh: 1: brunch: not found
==> test: [ERROR] 
==> test: [ERROR] npm ERR! SingularityUI@0.3.0 postinstall: `bower install --allow-root && brunch build --production`
==> test: [ERROR] npm ERR! Exit status 127
==> test: [ERROR] npm ERR! 
==> test: [ERROR] npm ERR! Failed at the SingularityUI@0.3.0 postinstall script.
==> test: [ERROR] npm ERR! This is most likely a problem with the SingularityUI package,
==> test: [ERROR] npm ERR! not with npm itself.
==> test: [ERROR] npm ERR! Tell the author that this fails on your system:
==> test: [ERROR] npm ERR!     bower install --allow-root && brunch build --production
[...]

PR #553, opened 2015-05-18
upgrade to new base_image

ssalinas:
base_image:
- now relies on public version of singularity cookbook (/cc @eherot )
- upgrades maven after chef run (singularity cookbook does not support specifying a maven version)
- upgrade docker after chef run (singularity cookbook does not support specifying a docker version)
- run mesos-slave with cgroups isolation turned on as well

test:
- installs config such that the SingularityExecutor is usable right away
- don't need to upgrade maven in provision any more

This should fix the issue seen in #550 from the maven version. As of the last few runs I did not see the brunch error mentioned in #552 either

@tpetr

### DIFF
--- a/vagrant/Berksfile
+++ b/vagrant/Berksfile
@@ -1,3 +1,7 @@
 source 'https://supermarket.getchef.com'
 
-cookbook 'singularity', path: '../cookbook'
+cookbook 'singularity'
+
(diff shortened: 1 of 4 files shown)


ssalinas:
merging this to going along with the moved cookbook from #549

PR #556, opened 2015-05-21
S3 uploader checks

ssalinas:
let's do away with all these s3 uploader warnings... Don't try and upload a file if some other process still has it open

@tpetr


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3Uploader.java
```diff
@@ -182,4 +188,21 @@ private void uploadSingle(int sequence, Path file) throws Exception {
     LOG.info("{} Uploaded {} in {}", logIdentifier, key, JavaUtils.duration(start));
   }
 
+  public static boolean fileOpen(Path path) {
+    try {
+      SimpleProcessManager lsof = new SimpleProcessManager(LOG);
+      List<String> cmd = ImmutableList.of("lsof", "|", "grep", path.toAbsolutePath().toString());
```
tpetr: i think we can get away with just `("lsof", path.toAbsolutePath().toString())` (http://www.thegeekstuff.com/2012/08/lsof-command-examples/, 2nd item)

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3Uploader.java
```diff
@@ -148,19 +150,23 @@ private void uploadBatch(List<Path> toUpload) {
...
-        Files.delete(file);
-      } catch (S3ServiceException se) {
-        metrics.error();
-        LOG.warn("{} Couldn't upload {} due to {} ({}) - {}", logIdentifier, file, se.getErrorCode(), se.getResponseCode(), se.getErrorMessage(), se);
-      } catch (Exception e) {
-        metrics.error();
-        LOG.warn("{} Couldn't upload or delete {}", logIdentifier, file, e);
-      } finally {
-        context.stop();
+      if (!fileOpen(file)) {
```
tpetr: chatted about this in person, but let's make this configurable in the S3 uploader config file

ssalinas:
@tpetr made it configurable and updated the command. I merge the multipart stuff in here because I had already made the configuration available in the right spots and was going to have merge conflicts there anyways

PR #559, opened 2015-05-27
Hubspot basepom

jhaber:
Would like to use [HubSpot/basepom](https://github.com/HubSpot/basepom) as the parent on all of our opensource projects to more sanely manage dependencies across all of them

@tpetr @ssalinas @wsorenson


jhaber:
@tpetr @ssalinas not sure whats up with the build failure on travis (master has the same thing), but it works for me locally

ssalinas:
I think the travis failures are npm related, they seem to fail intermittently on there for some reason. Builds fine for me too locally. We can put it through hs_staging/qa first just to be sure

PR #560, opened 2015-05-28
Add links to logs from request service page

tejasmanohar:
I've added quick links to logs from the request service page so developers who want to take a peek at the logs can do so without going through the file explorer.

### DIFF
--- a/SingularityUI/config.coffee
+++ b/SingularityUI/config.coffee
@@ -46,7 +46,7 @@ exports.config =
             defaultCpus: process.env.SINGUALRITY_DEFAULT_CPUS ? 1
             defaultMemory: process.env.SINGULARITY_DEFAULT_MEMORY ? 128
             defaultHealthcheckIntervalSeconds: process.env.SINGULARITY_DEFAULT_HEALTHCHECK_INTERVAL_SECONDS ? 5
-            defaultHealthcheckTimeoutSeconds: process.env.SINGULARITY_HEALTHCHECK_TIMEOUIT_SECONDS ? 5
+            defaultHealthcheckTimeoutSeconds: process.env.SINGULARITY_HEALTHCHECK_TIMEOUT_SECONDS ? 5
             defaultDeployHealthTimeoutSeconds: process.env.SINGULARITY_DEPLOY_HEALTH_TIMEOUT_SECONDS ? 120
             hideNewDeployButton: process.env.SINGULARITY_HIDE_NEW_DEPLOY_BUTTON ? "false"
             hideNewRequestButton: process.env.SINGULARITY_HIDE_NEW_REQUEST_BUTTON ? "false"


SingularityUI/app/assets/_index.mustache
```diff
@@ -27,9 +27,9 @@
             hideNewRequestButton: {{{hideNewRequestButton}}},
             defaultCpus: {{{defaultCpus}}},
             defaultMemory: {{{defaultMemory}}},
-            defaultHealthcheckIntervalSeconds: {{{defaultHealthcheckIntervalSeconds}}},
-            defaultHealthcheckTimeoutSeconds: {{{defaultHealthcheckTimeoutSeconds}}},
-            defaultDeployHealthTimeoutSeconds: {{{defaultDeployHealthTimeoutSeconds}}},
+            defaultHealthcheckIntervalSeconds: 5,
+            defaultHealthcheckTimeoutSeconds: 10,
+            defaultDeployHealthTimeoutSeconds: 120,
```
tpetr: can you un-commit these defaults? i'll roll out the legit fix for this to master

tejasmanohar: sorry, yeah. didn't mean to `add` those files but did a `commit -a` by accident

tejasmanohar: i didn't want to amend the commits and force push since that makes things messy but i restored the values in another commit 71b524d058f594da6a878dfef5408435f0b93cd8

tpetr:
This is a good first step, but we need to make these customizable (i.e. `{{taskId}}/service.log` is valid when we're using our custom Singularity executor, but when using the default Mesos executor, `stdout` is a better option). I'd suggest these steps:
- Move these paths into `window.config.runningTaskLogPath` and `window.config.finishedTaskLogPath` (set in `_index.mustache`)
- Add good defaults in `config.coffee` (let's have them both be `stdout`, since the default out-of-the-box usage of Singularity would be to use the default Mesos executor)
- Update the Request view to fill in `{{ taskId }}` and then pass the values through to the template
- Update `IndexView.java` to surface the `{{ runningTaskLogPath }}` and `{{ runningTaskLogPath }}` values

tejasmanohar:
@tpetr Thanks for the feedback! I added those three settings and tested.

SingularityService/src/main/java/com/hubspot/singularity/views/IndexView.java
```diff
@@ -30,6 +30,9 @@
   private final long defaultHealthcheckTimeoutSeconds;
   private final long defaultDeployHealthTimeoutSeconds;
 
+  private String runningTaskLogPath;
+  private String finishedTaskLogPath;
```
tpetr: can you make these fields match the rest and use `private final` ?

SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -103,3 +103,6 @@ Handlebars.registerHelper 'humanizeFileSize', (bytes) ->
 Handlebars.registerHelper 'usernameFromEmail', (email) ->
     return '' if not email
     email.split('@')[0]
+
+Handlebars.logger.level = 0;
+
```
tpetr: what does setting this do? was this something needed for testing, or is it not a big deal to leave here?

tejasmanohar: my bad. fixed- f2559221499181be3332a14eb36e21ffd477370c

SingularityUI/app/views/expandableTableSubview.coffee
```diff
@@ -51,6 +51,7 @@ class ExpandableTableSubview extends View
         @$el.html @template
             synced:  @collection.synced
             data:    _.pluck @collection.models, 'attributes'
+            config: config
```
tpetr: one thing that I probably didn't make clear enough (2nd to last bullet point in https://github.com/HubSpot/Singularity/pull/560#issuecomment-106590733) is we need to interpolate `TASK_ID` in the log path filename(s) to cover the situation where the Singularity executor is used. I'd suggest specifically using the string `$TASK_ID` for interpolation. This makes things a little hairier because we probably don't want to be doing this for every usage of `ExpandableTableSubview`, so i'd consider adding a `getRenderData()` method that subclasses can override with custom data

SingularityUI/app/handlebarsHelpers.coffee
tejasmanohar: @tpetr the Handlebars logger level makes `{{log ...}}` values actually show in console (equivalent of console.log)- we don't do this in our code but not sure if we want to leave this setting on or off. In Rails, there's also a logger level and it's on in production as well.

tejasmanohar:
Updated the fork.

SingularityUI/app/templates/requestDetail/requestActiveTasks.hbs
```diff
@@ -37,6 +37,12 @@
                         {{timestampFromNow updatedAt}}
                     </td>
                     <td class="actions-column hidden-xs">
+                        <a href="{{ appRoot }}/task/{{ taskId.id }}/tail/{{ subtituteTaskId ../../taskLogPath taskId.id }}" title="Log">
```
tpetr: typo

tejasmanohar:
@tpetr resolved

SingularityService/src/main/java/com/hubspot/singularity/views/IndexView.java
```diff
@@ -58,6 +61,9 @@ public IndexView(String singularityUriBase, String appRoot, SingularityConfigura
     this.defaultHealthcheckIntervalSeconds = configuration.getHealthcheckIntervalSeconds();
     this.defaultHealthcheckTimeoutSeconds = configuration.getHealthcheckTimeoutSeconds();
     this.defaultDeployHealthTimeoutSeconds = configuration.getDeployHealthyBySeconds();
+
+    this.runningTaskLogPath = configuration.getUiConfiguration().getRunningTaskLogPath();
+    this.finishedTaskLogPath = configuration.getUiConfiguration().getRunningTaskLogPath();
```
tpetr: typo here -- need `getFinishedTaskLogPath()`

PR #565, opened 2015-06-02
save run now command-line-input to localstorage per request

benrlodge:
When running a request, this will save the command line input to local storage. This is similar to the saving of the filename if redirecting to tailing (added with this branch), but this will save the input for each request individually. 

@tpetr

### DIFF
--- a/SingularityUI/app/models/Request.coffee
+++ b/SingularityUI/app/models/Request.coffee
@@ -11,6 +11,7 @@ class Request extends Model
 
     # When we show the JSON dialog, we will ignore these attributes
     ignoreAttributes: ['id', 'paused', 'deleted', 'hasActiveDeploy', 'canBeRunNow', 'canBeBounced', 'starred']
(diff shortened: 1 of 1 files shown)


SingularityUI/app/models/Request.coffee
```diff
@@ -146,8 +149,9 @@ class Request extends Model
                     @run( @data.commandLineInput ).done callback( @data )
                     return true
 
-            afterOpen: -> 
+            afterOpen: => 
                 $('#filename').val localStorage.getItem('taskRunRedirectFilename')
+                $('#commandLineInput').val localStorage.getItem(@localStorageCommandLineInputKeyPrefix + @id)
```
tpetr: can't seem to find where `@id` gets set -- are you sure this works properly?

benrlodge: `@id` comes from the request model (in addition to the attribute id - `@get("id")`). I'm having no trouble in my tests.

PR #568, opened 2015-06-04
start tailing only after model and collection are fetched

benrlodge:
Right now, on initial load of the tailing log file view, `startTailing` can run before the model is fetched, which leads to tailing not being started (as `isStillRunning` returns false). This change will require both the collection and model to be fetched before tailing starts.

@tpetr

### DIFF
--- a/SingularityUI/app/controllers/Tail.coffee
+++ b/SingularityUI/app/controllers/Tail.coffee
@@ -18,13 +18,9 @@ class TailController extends Controller
         app.showView @view
 
         if @offset?
-            $.when( @models.taskHistory.fetch(), @collections.logLines.fetchOffset(@offset) ).then =>
-                @view.afterInitialData()
+            $.when( @collections.logLines.fetchOffset(@offset), @refresh() ).then => @view.afterInitialOffsetData()
         else
-            $.when( @models.taskHistory.fetch(), @collections.logLines.fetchInitialData() ).then =>
-                @view.afterInitialData()
-        
-        @refresh()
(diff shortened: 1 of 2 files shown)


tpetr:
i'd consider moving this into the Tail controller instead. that way, you can use `$.when()` to track when both initial fetches have completed, and then trigger a new event (`allFetched`? something like that) which you can utilize in the view

benrlodge:
sure, np. updated.

SingularityUI/app/controllers/Tail.coffee
```diff
@@ -16,11 +16,14 @@ class TailController extends Controller
...
-            @collections.logLines.fetchOffset @offset
+            $.when( @models.taskHistory.fetch(), @collections.logLines.fetchOffset(@offset) ).then =>
+                @view.afterInitialData()
         else
-            @collections.logLines.fetchInitialData()
-
+            $.when( @models.taskHistory.fetch(), @collections.logLines.fetchInitialData() ).then =>
+                @view.afterInitialData()
+        
         @refresh()
```
tpetr: nitpick -- consider storing the fetchOffset / fetchInitialData result as a variable, and then call `$.when(fetchPromise, @refresh())`, so that we're not calling `taskHistory.fetch()` twice

benrlodge: ah right, didn't mean to still call @refresh again there. i like the idea of storing the fetchOffset / fetchInitialData result, but just realized I need to call a different method if the link is an offset (so it doesn't tail), so I removed the additional refresh call.

Issue #557, opened 2015-05-22
overdue to finish email missing data


tpetr:
`It was expected to run again at .`

tpetr:
fixed in https://github.com/HubSpot/Singularity/pull/633

PR #571, opened 2015-06-10
Mailer missing values

tejasmanohar:
Solves #557 

1) Wrong variable name was being used
2) Text doesn't properly express what the value is showing as you mentioned


ssalinas:
closing in favor of #633

PR #572, opened 2015-06-11
Service.log never complete

tejasmanohar:
resolves issue w/ browser tailing ending too soon by checking terminal states.

>If I have a job, and I wanna monitor it's progress-- I click into the service.log and I see it chugging along, but as it finishes, it's very likely that I won't see the last few seconds worth of logs which can be the difference between success or fail and why it fails, so then to figure that out, I need to click back out and then into tail-of-finished-service.log.
It's not a huge issue, but a minor annoyance that adds up over time. Would be awesome if I could monitor the progress all the way to the bitter end.


ssalinas:
closing in favor of #634

PR #573, opened 2015-06-11
Task state labels

tejasmanohar:
added task state labels as @tpetr suggested
> color-coded badges for task states throughout the singularity UI. Makes it easier to understand what's going on with a quick glance

they are controlled by a switch statement, simpler to read than object literals in this situation IMO since there are many options for 1 color (see http://toddmotto.com/deprecating-the-switch-statement-for-object-literals/)

### DIFF
--- a/SingularityUI/app/handlebarsHelpers.coffee
+++ b/SingularityUI/app/handlebarsHelpers.coffee
@@ -107,7 +107,15 @@ Handlebars.registerHelper 'usernameFromEmail', (email) ->
 Handlebars.registerHelper 'substituteTaskId', (value, taskId) ->
     value.replace('$TASK_ID', taskId)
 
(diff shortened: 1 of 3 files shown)


SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -106,3 +106,16 @@ Handlebars.registerHelper 'usernameFromEmail', (email) ->
 
 Handlebars.registerHelper 'substituteTaskId', (value, taskId) ->
     value.replace('$TASK_ID', taskId)
+
+Handlebars.registerHelper 'getLabelClass', (state) ->
```
simplyianm: I think an object literal would be easier to maintain.

ssalinas:
will be rolled in with #629 , closing

PR #578, opened 2015-06-16
make mode optional in SingularityVolume

ssalinas:
@tpetr 
fix that NPE


tpetr:
@stevenschlansker just FYI in case you guys have hit this before (not including `mode` field in `SingularityVolume` JSON results in scheduler NPE)

stevenschlansker:
Thanks for the heads up, we actually haven't used volumes at all, so we never noticed this :)

PR #587, opened 2015-06-30
check if command is present first (fixes #586)

### LINKED ISSUE #586: Deploying with a custom executor but no command triggers UNRECOVERABLE_ERROR in scheduler
https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java#L285

/cc @edhurtig

tpetr:


EdHurtig:
Fantastic

Issue #599, opened 2015-07-08
IllegalStateException in SingularityMesosScheduler.statusUpdate


chriskite:
```
Jul 08 11:43:16 ip-10-42-10-87.ec2.internal bash[24028]: ! java.lang.IllegalStateException: Optional.get() cannot be called on an absent value
Jul 08 11:43:16 ip-10-42-10-87.ec2.internal bash[24028]: ! at com.google.common.base.Absent.get(Absent.java:47) ~[singularity.jar:0.4.2-SNAPSHOT]
Jul 08 11:43:16 ip-10-42-10-87.ec2.internal bash[24028]: ! at com.hubspot.singularity.mesos.SingularityMesosScheduler.statusUpdate(SingularityMesosScheduler.java:297) ~[singularity.jar:0.4.2-SNAPSHOT]
Jul 08 11:43:16 ip-10-42-10-87.ec2.internal bash[24028]: ! at com.hubspot.singularity.mesos.SingularityMesosSchedulerDelegator.statusUpdate(SingularityMesosSchedulerDelegator.java:218) ~[singularity.jar:0.4.2-SNAPSHOT]
Jul 08 11:43:16 ip-10-42-10-87.ec2.internal bash[24028]: ERROR [2015-07-08 16:43:16,621] com.hubspot.singularity.SingularityAbort: Singularity on ip-10-42-10-87.ec2.internal is aborting due to UNRECOVERABLE_ERROR
```

chriskite:
The patch I wrote fixed the issue, and my Singularity isn't stuck in an endless start-fail-abort loop now.

tpetr:
Thanks for the PR, @chriskite! Would you mind giving a little more detail about the steps that lead up to Singularity getting into its death spiral? (i.e. what were you intending to run on Singularity, were you manually adding/removing data from ZK, were you running multiple frameworks in the Mesos cluster, etc...) Just trying to make sure there isn't more task data cleanup we should do in your PR.

chriskite:
Sure, my environment is basically like this:

Mesos with Marathon, Singularity, and my own framework Mescal.
3 node ZK cluster
Not running MySQL for Singularity
Singularity running via Docker on CoreOS with systemd
About 15 scheduled tasks in Singularity
No data manually touched in ZK

Singularity was running flawlessly for about a week. Then this morning it died with the error I posted. Once it aborts, systemd restarts it, but it dies again in the same way.

[...]

PR #600, opened 2015-07-08
check presence of task object in SingularityMesosScheduler.statusUpdate

chriskite:
Potential patch for issue #599


wsorenson:
Hey Chris; thanks for the patch.

How did Singularity get into this state? My goal thus far has been to handle unknown and poor state without going overboard, and I think this is probably worth handling, but in years and millions of tasks I haven't seen it so I assume it's related to manual actions in ZooKeeper or some corruption in Mesos?

One reason I'd like to know is because I would like to make sure we handle it correctly - should we simply return immediately from the update function rather than continue? In any case we will probably want a test to accompany any scheduler change test verifying it's correct behavior.

wsorenson:
Closing in favor of #631

PR #609, opened 2015-07-10
Validate and add TASK_ERROR support

tpetr:
https://github.com/HubSpot/Singularity/pull/598 + `TASK_ERROR` support

From http://mesos.apache.org/blog/mesos-0-21-0-released/:

> We have also introduced the notion of a TASK_ERROR state, distinct from TASK_LOST. The semantic difference is that tasks that are updated as lost can be rescheduled and should succeed, while tasks with status error will continue to fail if they are rescheduled. In Mesos 0.21.0 we do not send TASK_ERROR but it has been defined so frameworks can prepare to receive it. We will start sending it in Mesos 0.22.0.


stevenschlansker:
:+1:

wsorenson:
Are we going to add another PR for actually handling this state successfully? How should Singularity interpret TASK_ERROR?

tpetr:
Right now we treat TASK_ERROR as a terminal failure state, same as TASK_LOST. I imagine if the task is related to a deploy, we should immediately fail it. Not 100% sure what to do for the other situations (bounce, scale, etc.)

tpetr:
Better TASK_ERROR handling will be addressed in another PR -- filed #645 to track it for now.

PR #610, opened 2015-07-10
Switch to HubSpot's basepom

jhaber:
Builds for me, let me know if you have the same issue @wsorenson 

@ssalinas @tpetr


jhaber:
Nevermind it has the same dependency problems, I was running `mvn package` locally but I forgot it doesn't run all the basepom validation plugins at that phase (:-1:)

PR #612, opened 2015-07-10
update docker.sh.hbs template

ssalinas:
with certain variables put in, this template breaks with things like `DOCKER_ENV="$DOCKER_ENV -e JVM_DEFAULT_OPTS="-Xms${JVM_MIN_HEAP:-128m}""`

sets of quotes end up giving us issues. This interprets the variables better


ssalinas:
This has been behaving well, new template handles many more cases for env variables than before, going to merge

PR #616, opened 2015-07-13
More careful & less sanitization of rack + hostnames

wsorenson:
When we do sanitize, make it explicit.

In other parts of the code, pass along the original host and rack.


SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskId.java
```diff
@@ -77,26 +78,58 @@ public int compare(SingularityTaskId o1, SingularityTaskId o2) {
...
-    this.host = host;
+    this.sanitizedHost = sanitizedHost;
+    this.sanitizedRackId = sanitizedRackId;
+  }
+
+  @JsonCreator
+  public SingularityTaskId(@JsonProperty("requestId") String requestId, @JsonProperty("deployId") String deployId, @JsonProperty("nextRunAt") Long nextRunAt, @JsonProperty("startedAt") Long startedAt,
+      @JsonProperty("instanceNo") int instanceNo, @JsonProperty("host") String host, @JsonProperty("sanitizedHost") String sanitizedHost,
+      @JsonProperty("sanitizedRackId") String sanitizedRackId, @JsonProperty("rackId") String rackId) {
+    this(requestId, deployId, Objects.firstNonNull(startedAt, nextRunAt), instanceNo, Objects.firstNonNull(sanitizedHost, host), Objects.firstNonNull(sanitizedRackId, rackId));
```
jhaber: Just a heads up, `Objects.firstNonNull` blows up if both args are null so if any of these are optional it will cause problems

stevenschlansker:
:+1: We'd love to have non-sanitized slave hostnames in the API

PR #618, opened 2015-07-14
Docker dev rework

ssalinas:
- Move around some files to keep structure consistent and get rid of old unused files
- Create a base image for the SingularityExecutor slave that we can build on (ie java/logrotate/mesos installed). This is built separately right now, and won't need to be updated often. Images for mesos 0.21.1 and 0.22.1 with Java 8u45 are built and pushed to docker hub.
- Build a mesos-slave image with the SingularityExecutor configured and uses this for our slaves
- Run each slave on a different port during integration tests in case we move to host network mode
- More configurable properties in integration tests (ie hostnames/ports)
- create a dev wrapper script for starting/stopping/rebuilding all images and containers (eventually this will use all maven plugin, but until we can run in host mode via the plugin it uses docker-compose)

I'm happy to axe the vagrant stuff with this too, but I figured we would want to use the docker dev workflow for a bit first to make sure it suits our needs

@tpetr
/cc @kwm4385


ssalinas:
no actual functionality in singularity impacted here, just dev workflow, going to merge this

tpetr:
:dancer:

PR #619, opened 2015-07-14
Update abstractions for docker protos items

ssalinas:
We have abstractions that mirror the mesos Protos fields for some things but not others, this will complete those for docker items

/cc @tpetr


ssalinas:
FYI, going to still add extra constructors in so previous client versions will still work.

tpetr:
sounds good, let's deprecate those the proto-based constructors and remove in a future version

ssalinas:
Ran into a small snag make the old proto-based constructors worked. Since two were Optional<> 's I ran into 'method with same erasure' issues trying to implement the second constructor. Let me know if can think of a better workaround for the same erasure issue

tpetr:
@ssalinas this good to merge?

ssalinas:
Yep

PR #623, opened 2015-07-17
New table sorting

kwm4385:
Sortable lib is in /vendor right now due to brunch not concatenating it to app.js/.css. 
@tpetr Should we get that worked out or pull as is?


tpetr:
@benrlodge when you have a minute, could you help diagnose the bower issue kevin is facing? would prefer not to keep sortable in the vendor folder

benrlodge:
@kwm4385 check out the overrides property in the package.json. Likely brunch can't find the js/css files it's looking for (based on the sortable package.json). Check out how it was done with vex as a reference. If that's not the issue feel free to ping me and I'll take a closer look. 

overrides": {
    "vex": {
      "main": "js/vex.combined.min.js"
    },

kwm4385:
Thanks @benrlodge

SingularityUI/app/templates/dashboardTable/dashboardStarred.hbs
```diff
@@ -2,12 +2,12 @@
     <table class="table table-striped">
```
ssalinas: should this table also have the `sortable-theme-bootstrap" data-sortable` attributes?

kwm4385: That it should

ssalinas:
@kwm4385 
Three things on this after taking a closer look at this. First, there is something funky with the alignment on the requests table headers:

screen shot 2015-07-24 at 9 20 53 am

Second thing, when the auto refresh happens every ten seconds or so, the current sort is erased. Not sure how tough it would be, but it'd be nice to keep the table sorted the same after the refresh.

Third thing, on the tasks table, when sorting my Memory usage, it is sort as a string, not as the number since the label is on there, meaning 111111 MB will be above 60 MB will be above 8 MB, etc. A user would expect that sort to go high <--> low by the actual memory value, so we should probably implement it that way. A similar thing happens for the 'Started' field

Last thing, sortable tables everywhere make me very happy :smile:

kwm4385:
When we refresh the table, the entire subview that its in is destroyed and re-rendered so from sortable's point of view, its a whole new table it's initializing. We could save the state and reapply it to the new table, but sortable doesn't seem to have a way to programmically interact with the table that way.

tpetr:
what's with the tejas commits at the top of this PR? except for rare cases, we shouldn't be pulling in commits from other PRs.

kwm4385:
I didn't realize what branch I was on when I went to merge that PR into hs_staging. It has already been merged there however so it shouldn't have any effect.

ssalinas:
this got folded into #629 , closing this one

PR #625, opened 2015-07-20
purge dead slaves after a configurable amount of time

### LINKED ISSUE #617: Eventually prune dead slaves
We had a misconfiguration which caused slaves to endlessly restart with new Slave IDs.  All of these slaves are now forevermore in our history as dead -- there's thousands of them!  Obviously they shouldn't be there in the first place, but if a slave has been dead for months, eventually we should probably prune it.

wsorenson:
Resolves #617


stevenschlansker:
Yeah! :smile:

PR #626, opened 2015-07-20
downgrade jackson-datatype-protobuf back to 0.4.0

tpetr:
`jackson-datatype-protobuf` was upgraded from 0.4.0 to 0.8.0 in #611, which caused Singularity to have perf issues (hitting `zkAsyncTimeout` in TaskManager.getActiveTasks()). We believe the slowdown is from this commit: https://github.com/HubSpot/jackson-datatype-protobuf/commit/daab7a5e1e44dce6e16d6e14f239fd664d53c9da

Since we haven't hit any issues that the protobuf updates have addressed, it's probably fine to keep it on the old version for now.

FYI @stevenschlansker in case you pulled this stuff into your custom build

/cc @jhaber

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,7 +116,7 @@
       <dependency>
         <groupId>com.hubspot.jackson</groupId>
         <artifactId>jackson-datatype-protobuf</artifactId>
-        <version>0.8.0</version>
+        <version>0.4.0</version>
       </dependency>
 
       <dependency>


jhaber:
I pushed version 0.9.0 with a [lame fix](https://github.com/HubSpot/jackson-datatype-protobuf/commit/00895aeb768904c832d0f7f3584016e7a9f818b6). Jackson has its own Serializer/Deserializer caches but using our own seems to be an order of magnitude faster for some reason :panda_face: The main downside is that the new cache is global so it doesn't properly respect different `ObjectMapper` configurations

stevenschlansker:
tag @cowtowncoder, maybe he happens to take a look and see something obvious

cowtowncoder:
@stevenschlansker @jhaber

I think the difference could be due to contextualization being done in case of lookup, so while instances are cached, `ContextualDeserializer.createContextual()` is called. Although if that's a no-op it should be cheap call. Profiler should show what is going on.

Deserializers are expected to usually resolve dependant deserializers on `createContextual` if possible.
But unlike with serializers which can also use `PropertySerializerMap` for efficient local caching, there is nothing for deserializers.

So I guess I would actually be interested in seeing what might be causing performance problems with lookups. Especially against 2.6.0, which was just released.

stevenschlansker:
@jhaber @tpetr Is there a good test dataset that we could run a profiler against?

jhaber:
I uploaded some sample data [here](https://s3.amazonaws.com/uploads.hipchat.com/10819/200218/0io3SjVHYUFEFL3/data.zip). It has a JSON array of `SingularityTask` objects and deserializing it shows a massive performance difference from jackson-datatype-protobuf 0.8.0 (no global cache) to 0.9.0 (global cache)

cowtowncoder:
Looking at code again with bit more thought, I think the approach Jackson would take here would be to statically determine set of deserializers needed, during contextualization (and/or during `resolve()`, similar to how `BeanDeserializer` works).
Since field types are known in advance, it seems possible to fully resolve set of deserializers that are needed, assign that to fields, and delegate so that message deserializes mostly by delegation.

This is why databind does not optimize caching of deserializers that much: as types are static (with exception of polymorphic type, for which simple caching is used using `ConcurrentHashMap` that `TypeDeserializerBase` holds onto) it is a one-off cost.

If the more dynamic approach is preferred, use of `ConcurrentHashMap` makes sense, but I think it really should be scoped by `ProtobufDeserializer` -- there will only be one instance of these (or, at most one per property, if contextualization is used), and as deserializers are reused it will work fine performance-wise, as well as be scoped by `ObjectMapper`.
With that change I think this would be a reasonable solution.

jhaber:
I tried that but it didn't help performance, maybe because the deserializer factory returns a new deserializer each time. I'll try changing that and retesting

jhaber:
I couldn't think of a great way to fix the problem of the deserializer factory returning new instances each time, so I added a non-static cache there as well. Since it makes a new deserializer factory each time the module is installed, I think this should be well-behaved. Performance looks good too, commit is [here](https://github.com/HubSpot/jackson-datatype-protobuf/commit/cd46b31f791467c6a98b30f130e7c6d08f4b6bb3)

cowtowncoder:
FWIW, `DeserializerFactory` is not caching, and should not be called directly by application code. `DeserializationContext` handles caching aspects.
But as long as performance is good, that should be fine.

cowtowncoder:
Commit makes sense, looks good.
Just one small suggestion: instead of doing `if (cache.containsKey(...))`, just use `get`, see if null (or is it `Optional`?) is returned. This avoids doing lookup (hashing, traversal) twice.

For serializer, you could also consider using `PropertySerializerMap` (its API is bit specialized, better check out its usage by `jackson-databind` if so); but I doubt performance-wise it makes huge difference.
It just uses bit less memory, and if there are usually just one or two alternatives, it is probably slightly faster. But as I said unlikely to matter a whole lot.

jhaber:
Thanks for the pointers, just released version 0.9.1

PR #629, opened 2015-07-23
Deploy detail page

kwm4385:


SingularityUI/app/views/expandableTableSubview.coffee
```diff
@@ -25,6 +25,7 @@ class ExpandableTableSubview extends View
 
     initialize: ({@collection, @template}) ->
         @listenTo @collection, 'sync', @render
+        @listenTo @collection, 'reset', @render
```
benrlodge: I'm not sure if we can drop sync completely here. I imagine reset should only fire when we call `collection.reset`, so other views using expandableTableSubview may not render.

benrlodge: sorry, misread that, thought you remove sync

kwm4385: It doesn't look like you can do multiple events in the same listenTo call. simpleSubview does it in a loop: 
```
  for eventName in ['sync', 'add', 'remove', 'change', 'reset']
    @listenTo @data, eventName, @render
```
I'll switch it to that. Though a bit overkill for now, it may be less confusing for anyone adding another.

SingularityUI/app/templates/deployDetail/deployHeader.hbs
```diff
@@ -0,0 +1,40 @@
...
+    </div>
+
+    <div class="row">
+        <div class="col-md-8">
+            <h3>
+                <span>
+                    {{ data.deployId }}
+                </span>
+                <span class="deploy-state" data-state="{{ data.deployResult.deployState }}">
+                    {{ data.deployResult.deployState }}
```
ssalinas: This can occasionally be blank, let's see if we can get a PENDING or something similar here for a deploy that is still in progress

SingularityUI/app/templates/deployDetail/activeTasks.hbs
```diff
@@ -0,0 +1,76 @@
...
+                        Deploy
+                    </th>
+                    <th class="hidden-sm hidden-xs">
+                        Started
+                    </th>
+                    <th class="hidden-xs">
+                        Updated
+                    </th>
+                    <th class="hidden-xs">
+                        {{! Actions }}
```
ssalinas: Let's make this header (the one above the `...` link) say `Logs` , the tooltip syas it, but a header would be helpful to know at first glance what it is

ssalinas:
@kwm4385 One other thing that would be very useful on this page is a consolidated-ish view of the healthchecks. If a deploy fails, one of the first things to look at would be a failed healthcheck (followed closely by the logs which there is already a link for)

SingularityUI/app/templates/deployDetail/activeTasks.hbs
kwm4385: Put on hold

kwm4385:
Need to add a link back to the associated deploy on the task detail page

SingularityUI/app/templates/deployDetail/deployInfo.hbs
```diff
@@ -0,0 +1,60 @@
...
+        </li>
+        <li class="col-md-4">
+            <div>
+                <h4>Completed<a data-clipboard-text="{{ timestampFromNow data.deployResult.timestamp }}">Copy</a></h4>
+                <p>{{ timestampFromNow data.deployResult.timestamp }}</p>
+            </div>
+        </li>
+        <li class="col-md-4">
+            <div>
+                <h4>Command<a data-clipboard-text="{{ timestampFromNow data.deploy.command }}">Copy</a></h4>
```
ssalinas: I think this needs to be data.deploy.executorData.cmd on this line and the one below. Currently coming up blank in UI

ssalinas:
One last place it would make sense to make a link. The requestHeader.hbs (under requestDetail), can show something like this.
screen shot 2015-07-30 at 8 53 02 am

would be good to have that link as well

that and the previous comment on the cmd and I think we can move this to hs_qa

kwm4385:
Staged

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -107,9 +110,42 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
...
+      @ApiParam("Request ID for deploy") @PathParam("requestId") String requestId,
+      @ApiParam("Deploy ID") @PathParam("deployId") String deployId) {
+
+    List<SingularityTaskId> taskIds = new ArrayList<SingularityTaskId>();
+    for (SingularityTaskId id : taskManager.getActiveTaskIdsForRequest(requestId)) {
+      if (id.getDeployId().equals(deployId)) {
+        taskIds.add(id);
+      }
+    }
+    return taskHistoryHelper.getHistoriesFor(taskIds);
```
tpetr: i'd move these lines into a new method in `TaskHistoryHelper`, perhaps `getTaskHistoryForDeploy()`

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -107,9 +110,42 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
...
+
+  @GET
+  @Path("/request/{requestId}/deploy/{deployId}/tasks/inactive")
+  @ApiOperation("Retrieve the task history for a specific deploy.")
+  public List<SingularityTaskIdHistory> getInactiveDeployTasks(
+      @ApiParam("Request ID for deploy") @PathParam("requestId") String requestId,
+      @ApiParam("Deploy ID") @PathParam("deployId") String deployId) {
+
+    List<SingularityTaskId> taskIds = new ArrayList<SingularityTaskId>();
+    for (SingularityTaskId id : taskManager.getInactiveTaskIdsForRequest(requestId)) {
```
tpetr: This isn't good enough, unfortunately. Singularity users can optionally persist non-operational data (i.e. history) into mysql, but this method only hits ZK for data. Consider adding a new `getBlendedHistory()` method in TaskHistoryHelper that accepts both a request and deploy ID.

kwm4385:
cc @tpetr

kwm4385:
Also ignore me forgetting to switch accounts.

SingularityService/src/main/java/com/hubspot/singularity/data/history/DeployTaskHistoryHelper.java
```diff
@@ -0,0 +1,92 @@
...
+    while(it.hasNext()) {
+      if (!it.next().getTaskId().getDeployId().equals(deployKey.getDeployId())) {
+        it.remove();
+      }
+    }
+    return histories;
+  }
+
+  public List<SingularityTaskIdHistory> getActiveDeployTasks(SingularityDeployKey key, Integer limitCount, Integer limitStart) {
+    List<SingularityTaskIdHistory> histories = this.getBlendedHistory(key, limitStart, limitCount);
```
tpetr: remove `limitCount` and `limitStart`, and just use `getFromZk()` instead

ssalinas:
@kwm4385 noticed something with the blended history helper. It seems that the count/limit is being applied before we filter things (and then filter them again in the case of inactive. For some older deploys this is causing nothing to show up since we fetch the most recent 5 tasks by request id, filter for deploy id, then filter again for inactive. Can we make it so that we get the correct number of tasks, after all the filtering is done?

SingularityService/src/main/java/com/hubspot/singularity/data/history/DeployTaskHistoryHelper.java
```diff
@@ -0,0 +1,112 @@
...
+  private final SingularityConfiguration singularityConfiguration;
+
+  @Inject
+  public DeployTaskHistoryHelper(TaskManager taskManager, HistoryManager historyManager, SingularityConfiguration singularityConfiguration) {
+    this.taskManager = taskManager;
+    this.historyManager = historyManager;
+    this.singularityConfiguration = singularityConfiguration;
+  }
+
+  public List<SingularityTaskIdHistory> getHistoriesFor(Collection<SingularityTaskId> taskIds) {
```
wsorenson: This method should not be public since it is not used publicly. Also looks like a direct copy/paste from TaskHistoryHelper. If so, it should either use a library function or inherit the function.

SingularityService/src/main/java/com/hubspot/singularity/data/history/DeployTaskHistoryHelper.java
```diff
@@ -0,0 +1,112 @@
...
+      public boolean apply(SingularityTaskIdHistory input) {
+        return input.getTaskId().getDeployId().equals(deployKey.getDeployId());
+      }
+    });
+
+    return ImmutableList.copyOf(deployHistories);
+  }
+
+  public List<SingularityTaskIdHistory> getActiveDeployTasks(SingularityDeployKey key) {
+    List<SingularityTaskIdHistory> histories = this.getFromZk(key);
```
wsorenson: Wouldn't there be active deploy history inside of MySQL?

SingularityService/src/main/java/com/hubspot/singularity/data/history/DeployTaskHistoryHelper.java
```diff
@@ -0,0 +1,112 @@
...
+    List<SingularityTaskIdHistory> histories = this.getBlendedHistory(key, 0, limit);
+
+    final Iterable<SingularityTaskIdHistory> inactiveHistories = Iterables.filter(histories, new Predicate<SingularityTaskIdHistory>() {
+      @Override
+      public boolean apply(SingularityTaskIdHistory input) {
+        return input.getLastTaskState().get().isDone();
+      }
+    });
+
+    ImmutableList result = ImmutableList.copyOf(inactiveHistories);
```
wsorenson: This is not parametrized and should be generate an IDE warning. It should be fixed.

SingularityService/src/main/java/com/hubspot/singularity/data/history/DeployTaskHistoryHelper.java
```diff
@@ -0,0 +1,112 @@
...
+      }
+    });
+
+    return ImmutableList.copyOf(activeHistories);
+  }
+
+  public List<SingularityTaskIdHistory> getInactiveDeployTasks(SingularityDeployKey key, Integer limitCount, Integer limitStart) {
+    // We don't know our limits yet before filtering task state
+    Integer limit = (int) (singularityConfiguration.getHistoryPurgingConfiguration().getDeleteTaskHistoryAfterTasksPerRequest().or(10000) * 1.2);
+    List<SingularityTaskIdHistory> histories = this.getBlendedHistory(key, 0, limit);
```
wsorenson: This feels a bit dodgy. Should we build a data structure for this? If not, are we sure we aren't potentially making a disastrous call to MySQL/ZK?

ssalinas: going to add a migration (in new pr) that will give us a deployId column, that will make things much cleaner since right now we can't find tasks for a deploy without looking at a much larger number of items then filtering

SingularityService/src/main/java/com/hubspot/singularity/data/history/DeployTaskHistoryHelper.java
ssalinas: There won't be active tasks in mysql, we don't persist them if they are still active, still in lb cleanup, or part of a pending deploy. The task history by requestId does this the same way

wsorenson: Okay, these methods are poorly named. I read it as tasks for the active deploy. It should read like activeTasksForDeploy and inactiveTasksForDeploy, though I wonder - is there a time where the APP will only ask for one and not the other?

PR #634, opened 2015-07-23
Log never complete

kwm4385:
Tested and working.

### DIFF
--- a/SingularityUI/app/views/tail.coffee
+++ b/SingularityUI/app/views/tail.coffee
@@ -1,5 +1,7 @@
 View = require './view'
 
+TERMINAL_TASK_STATES = ['TASK_KILLED', 'TASK_LOST', 'TASK_FAILED', 'TASK_FINISHED']
+
 class TailView extends View
 
     pollingTimeout: 3000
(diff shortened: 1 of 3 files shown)


SingularityUI/app/views/tail.coffee
```diff
@@ -1,5 +1,7 @@
 View = require './view'
 
+TERMINAL_TASK_STATES = ['TASK_KILLED', 'TASK_LOST', 'TASK_FAILED', 'TASK_FINISHED']
```
tpetr: we have this floating around in a few places -- we should consolidate them in a constants file

SingularityUI/app/views/tail.coffee
```diff
@@ -176,7 +178,8 @@ class TailView extends View
         @$el.addClass 'tailing'
 
     stopTailing: ->
-        return if @isTailing isnt true
+        task = @model.get('taskUpdates')[..].pop()
```
tpetr: seems wasteful to copy the array just to pop -- `_.last()` would get the job done too

SingularityUI/app/views/tail.coffee
```diff
@@ -176,7 +178,8 @@ class TailView extends View
...
+        return if @isTailing isnt true and task.taskState in TERMINAL_TASK_STATES
```
tpetr: this logic is confusing to me -- it'd be clearer to modify https://github.com/HubSpot/Singularity/blob/log_never_complete/SingularityUI/app/views/tail.coffee#L36 to call a new method that does one final tail fetch, and then stops the tailing interval

SingularityUI/app/views/tail.coffee
tejasmanohar: sorry, don't know what i was thinking there. could've just been `arr[arr.length-1]` but cool that there's a _.last() helper

ssalinas:
this got folded into #629 , closing this one

PR #635, opened 2015-07-27
Truncate beginning of S3 log filenames

kwm4385:
Truncate the beginning of s3 log files to enable reading of useful information.

### DIFF
--- a/SingularityUI/app/handlebarsHelpers.coffee
+++ b/SingularityUI/app/handlebarsHelpers.coffee
@@ -120,3 +120,6 @@ Handlebars.registerHelper 'getLabelClass', (state) ->
             'danger'
         else
             'default'
+
+Handlebars.registerHelper 'trimS3File', (filename) ->
+    return '...' + filename.substring(filename.indexOf("service.log"));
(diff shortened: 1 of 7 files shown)


SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -106,3 +106,6 @@ Handlebars.registerHelper 'usernameFromEmail', (email) ->
 
 Handlebars.registerHelper 'substituteTaskId', (value, taskId) ->
     value.replace('$TASK_ID', taskId)
+
+Handlebars.registerHelper 'trimS3File', (filename) ->
+    return '...' + filename.substring(filename.indexOf("service.log"));
```
ssalinas: unfortunately the log name will not always be `service.log`. That is the default for the SingularityExecutor, but it could be anything for s3 logs. I might go more on the route of taking the taskId out of the name (since we are already on the task page) to make a name that is more readable and shorter without having to truncate

kwm4385: I was afraid that was the case. This model doesn't include the task id but I should be able to add it in.

tpetr:
This looks good but to be fully complete, we should surface the `s3KeyFormat` to the API (https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/config/S3Configuration.java#L28-L34), and strip things off from there.

PR #638, opened 2015-07-29
Add no-op main class to SingularityServiceIntegrationTests

kwm4385:
Sonatype needs a main jar file and associated javadocs for each module when releasing. Adding a no-op main class to solve this.
/cc @tpetr


tpetr:
LGTM, thanks

PR #640, opened 2015-08-07
adds ability to purge old task history from DB

wsorenson:
- introduces DB testing without mocks


SingularityService/src/test/java/com/hubspot/singularity/SingularityHistoryPurgerTest.java
```diff
@@ -0,0 +1,154 @@
...
+    super(true);
+  }
+
+  @Before
+  public void createTestData() throws Exception {
+    Handle handle = dbiProvider.get().open();
+
+    Database database = DatabaseFactory.getInstance().findCorrectDatabaseImplementation(new JdbcConnection(handle.getConnection()));
+
+    Liquibase liquibase = new Liquibase("singularity_test.sql", new FileSystemResourceAccessor(), database);
```
tpetr: you should use `../mysql/migrations.sql` instead so that we don't have to update the schema in two different places

PR #641, opened 2015-08-11
add shutdown hook to singularity executor

ssalinas:
right now SingularityExecutor isn't gracefully shutting down when it receives a signal. Adding a shutdown hook to trigger the same shutdown process (`monitor.shutdown()`) as would happen when mesos tells it to shut down

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
@@ -7,9 +7,12 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import com.google.inject.name.Named;
+import com.hubspot.singularity.executor.config.SingularityExecutorModule;
 import org.apache.mesos.ExecutorDriver;
 import org.apache.mesos.Protos;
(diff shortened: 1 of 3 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorRunner.java
```diff
@@ -54,6 +58,16 @@ public SingularityExecutorRunner(@Named(SingularityRunnerBaseModule.PROCESS_NAME
 
     final MesosExecutorDriver driver = new MesosExecutorDriver(singularityExecutor);
 
+    Runtime.getRuntime().addShutdownHook(new Thread("SingularityExecutorRunnerGracefulShutdown") {
+
+      @Override
+      public void run() {
+        LOG.info("Executor is shutting down...");
```
wsorenson: Why is it shutting down? Always be specific :)

ssalinas:
@tpetr @wsorenson the lock on shutdown works the way we thought it would. No processes left hanging waiting for locks

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
```diff
@@ -87,7 +91,16 @@ public SingularityExecutorMonitor(SingularityExecutorLogging logging, ExecutorUt
     SUBMITTED, REJECTED, TASK_ALREADY_EXISTED;
   }
 
-  public void shutdown(Optional<ExecutorDriver> driver) {
+  public void shutdownWithLock(Optional<ExecutorDriver> driver) {
+    try {
+      shutdownLock.lock();
```
tpetr: consider using a timeout

ssalinas:
@tpetr tried one last thing, this version of things will shut down cleanly every time that I have tried now, need to test more still though

PR #656, opened 2015-08-17
update volumes for integration test slaves

ssalinas:
@tpetr


ssalinas:
Ran a `mvn clean verify` on each branch, integration tests are back to normal with these changes in place.

PR #663, opened 2015-08-24
Resource Scheduler Interface (Experimental)

ssalinas:
The beginnings of some refactoring of the resource scheduler so that it might be possible to plug in something like https://github.com/Netflix/Fenzo into Singularity. Still a WIP


ssalinas:
closing since I won't work on this for a while, would need a rebase whenever I do anyways

PR #664, opened 2015-08-24
Runc support (Experimental)

ssalinas:
Support running containers directly via `runc` in SingularityExecutor. Still a WIP (built off mesos_23 branch)


ssalinas:
closing since I won't work on this for a while, would need a rebase whenever I do anyways

PR #667, opened 2015-08-25
add logsearch and use-cache option for live logs

ssalinas:
Adds two features:

- `-U / --use-cache` option to prefer the cached version of a live log file instead of re-downloading it
- `logsearch` command as an offline way of searching through files in the cache directory (with a shortcut of adding `--search` to the `logfetch` command for less retyping)


ssalinas:
Also added some better usage of cached files for downloading logs still on a machine:
- If it's a zipped file, it probably hasn't changed, use the cache
- If the task is not in a running state, and the cached file was downloaded after the task finished, use the cached file

Issue #666, opened 2015-08-24
Track run request through processing


sannessa:
When I submit a run request via the API through /api/requests/request/{requestId}/run I don't have a way to verify the completion of that request.

I can hit /api/history/request/{requestId}/tasks/active to see active tasks and /api/history/request/{requestId}/tasks to see tasks that have finished but I cannot link a run request to a finished task.

A UID that tracks the run request throughout the entire process would solve my problem.

wsorenson:
This is available in master now.

PR #669, opened 2015-08-25
add an optional runId to attach to running tasks over the API

wsorenson:
If no runId is specified, one is generated.

CC @jhaber @also


wsorenson:
Fixes #666

PR #672, opened 2015-08-31
add container fields, other updates to validator

ssalinas:
@wsorenson @stevenschlansker 
fixes to help avoid the reason for #668

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
@@ -190,6 +190,7 @@ public class SingularityValidator {
 
     if (request.isLoadBalanced()) {
       checkBadRequest(deploy.getServiceBasePath().isPresent(), "Deploy for loadBalanced request must include serviceBasePath");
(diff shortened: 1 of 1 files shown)


stevenschlansker:
Cool.  I'm still of the opinion that since we use `Optional<T>` pervasively in the domain objects, any non-Optional field should be assumed to be `@NonNull` and verified as such.  End users will expect that behavior.  I am neither specifically pushing for my PR nor against this PR, just trying to make sure my "philosophy" here is consistent with yours.

ssalinas:
Yes, definitely agree with the philosophy there. We are already doing our other checks for other invalid data or combinations of data in the validators (e.g. deploy id or not having a base path for a load balanced service), thought it made sense to have all of our defense against bad data in the same place

stevenschlansker:
:+1:

jeffbyrnes:
Adding some bits here after a convo in EverTrue’s #mesos Slack channel.

If a Request is `loadBalanced`, but is sent a Deploy without a `loadBalancerGroup`, the deploy will fail, with the load balancer response from Baragon being buried in the Deploy JSON.

PR #673, opened 2015-08-31
ON_DEMAND (oneOff) tasks will persist scheduling

wsorenson:
- This changes pre-existing behavior that cleared pending tasks when a
new schedule request is issued.

Given the nature of ON_DEMAND tasks, this is probably the user expected
behavior


wsorenson:
cc @jhaber

jhaber:
:+1:

PR #675, opened 2015-09-02
Simpler way to fetch tasks by deploy id

ssalinas:
@wsorenson addresses the comments from #629 
/cc @tpetr 

Also, would welcome any suggestions on the query to populate deployId. The requestId can also have dashes, so I'm currently substring-ing twice to get the depolyId as the first element, then trim off the rest

### DIFF
--- a/mysql/migrations.sql
+++ b/mysql/migrations.sql
@@ -77,5 +77,7 @@ ALTER TABLE `taskHistory` MODIFY `bytes` MEDIUMBLOB NOT NULL;
 ALTER TABLE `taskHistory` ADD COLUMN runId VARCHAR(100) NULL;
 
 --changeset ssalinas:7 dbms:mysql
-ALTER TABLE `taskHistory` ADD COLUMN deployId VARCHAR(100) NULL;
+ALTER TABLE `taskHistory`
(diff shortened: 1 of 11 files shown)


tpetr:
also FYI @kwm4385

kwm4385:
:thumbsup:

mysql/migrations.sql
```diff
@@ -75,3 +75,7 @@ ALTER TABLE `taskHistory` MODIFY `bytes` MEDIUMBLOB NOT NULL;
 
 --changeset wsorenson:6 dbms:mysql
 ALTER TABLE `taskHistory` ADD COLUMN runId VARCHAR(100) NULL;
+
+--changeset ssalinas:7 dbms:mysql
+ALTER TABLE `taskHistory` ADD COLUMN deployId VARCHAR(100) NULL;
+UPDATE `taskHistory` SET `deployId`` = SUBSTRING_INDEX(SUBSTRING_INDEX(`taskId``, '-', -5), '-', 1) WHERE `deployId`` IS NULL;
```
tpetr: how should we add / edit indexes in response to adding the deployId column?

PR #676, opened 2015-09-02
Deploy detail page

kwm4385:
Fixed some small issues brought up by @wsorenson in #629. Also #675


ssalinas:
@kwm4385 see #675, also that method does need to be public, it is called in the HistoryResource

PR #677, opened 2015-09-02
Don't enable task/request table search until collection loads (and then focus the field)

benrlodge:
The purpose of this PR is simply to focus the input field, but it made it easier to do this after refactoring the code a bit - removing the `preventSearchOverwrite` method, and just disabling the input field until the table loads, so no need to prevent anything.

@tpetr


eabbott:
(yay)

PR #679, opened 2015-09-04
let the container know what cgroup it is in

ssalinas:
@also

### DIFF
--- a/SingularityExecutor/src/main/resources/docker.sh.hbs
+++ b/SingularityExecutor/src/main/resources/docker.sh.hbs
@@ -105,9 +105,12 @@ DOCKER_NETWORK="--net=${raw_network,,}"
 DOCKER_NETWORK="--net=host"
 {{/if}}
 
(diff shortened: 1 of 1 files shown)


SingularityExecutor/src/main/resources/docker.sh.hbs
```diff
@@ -105,9 +105,12 @@ DOCKER_NETWORK="--net=${raw_network,,}"
 DOCKER_NETWORK="--net=host"
 {{/if}}
 
+CGROUP_PARENT="mesos/$CGROUPS_GUID"
+echo "CGROUP_PARENT=mesos/$CGROUPS_GUID" >> docker.env
```
tpetr: perhaps `echo "CGROUP_PARENT=${CGROUP_PARENT}"` instead to reduce copypasta?

ssalinas:
wont' actually need this after all, going to close for now

PR #681, opened 2015-09-08
when bouncing a request, record BOUNCED history event

tpetr:
No easy way to tell in the UI that a request was bounced otherwise.

/cc @ssalinas @wsorenson


wsorenson:
:+1:

PR #683, opened 2015-09-09
Add shell commands ui

kwm4385:


SingularityService/src/main/java/com/hubspot/singularity/views/IndexView.java
```diff
@@ -68,6 +74,15 @@ public IndexView(String singularityUriBase, String appRoot, SingularityConfigura
     this.finishedTaskLogPath = configuration.getUiConfiguration().getFinishedTaskLogPath();
 
     this.commonHostnameSuffixToOmit = configuration.getCommonHostnameSuffixToOmit().or("");
+
+    ObjectWriter ow = new ObjectMapper().writer().withDefaultPrettyPrinter();
```
tpetr: instead of instantiating a new `ObjectMapper` each time here, inject one in `UiResource` and pass it through the `IndexView` constructor

SingularityService/src/main/java/com/hubspot/singularity/views/IndexView.java
```diff
@@ -68,6 +74,15 @@ public IndexView(String singularityUriBase, String appRoot, SingularityConfigura
...
+    String json = "";
+    try {
+      json = ow.writeValueAsString(configuration.getUiConfiguration().getShellCommands());
```
tpetr: why not just `this.shellCommands = ow.writeValueAsString()`?

kwm4385:
Is the IndexView even being used right now since the UI is deployed separately and these shell commands coming from the deploy config for that?

SingularityUI/app/styles/forms.styl
```diff
@@ -3,4 +3,10 @@
...
 	&.label-light
-		font-weight normal
\ No newline at end of file
+		font-weight normal
+
+.cmd-description
+	margin-top 10px
+
+#cmd-confirm
+	color #5cb85c
```
benrlodge: lets try to use variables for colors, and try our best to keep the color pallete to a reasonable number (though if it makes sense to add new ones feel free) - https://github.com/HubSpot/Singularity/blob/master/SingularityUI/app/styles/colors.styl

tpetr:
@kwm4385 we might not be using it for our special static build, but in general (i.e. other Singularity users), `IndexView` is in use

kwm4385:
:+1:

kwm4385:
@wsorenson Since Eric is going to be doing work on shell commands, it would make sense for us to merge these UI components into the main branch. Sound good?

SingularityUI/app/views/task.coffee
```diff
@@ -9,6 +10,9 @@ class TaskView extends View
             'click [data-action="viewObjectJSON"]': 'viewJson'
             'click [data-action="viewJsonProperty"]': 'viewJsonProperty'
             'click [data-action="remove"]': 'killTask'
+            'submit form#runShell': 'executeCommand'
```
benrlodge: for consistency's sake let's stick with using the data-action selectors for all the task action events.

SingularityUI/app/views/task.coffee
```diff
@@ -61,4 +66,30 @@ class TaskView extends View
             setTimeout (=> @trigger 'refreshrequest'), 1000
 
 
+    executeCommand: (event) ->
+        event.preventDefault()
+        cmd = $("#cmd option:selected").text()
+        options = $('#cmd-option').val()
+        return if @$('button[type="submit"]').attr 'disabled'
```
benrlodge: let's be more specific and reference this button using an ID, always better to be explicit

SingularityUI/app/views/task.coffee
```diff
@@ -61,4 +66,30 @@ class TaskView extends View
...
+        return if !cmd
+        taskModel = new Task id: @taskId
+        taskModel.runShellCommand(cmd, options)
+        $('#cmd-confirm').text('Command Sent')
+
+
+    cmdSelected: (event) ->
+        cmd = config.shellCommands.filter((cmd) ->
+          return cmd.name == $("#cmd option:selected").text()
```
benrlodge: personally I prefer using `===, &&, ||`, but for consistency of how things have been coded in Singularity we should stick with using `is` for `===`, `or` for `||`, `and` for `&&`, etc

SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -119,3 +119,10 @@ Handlebars.registerHelper 'getLabelClass', (state) ->
             'danger'
         else
             'default'
+
+Handlebars.registerHelper 'isRunningState', (list, options) ->
+    switch _.last(list).taskState
```
benrlodge: use `@` for `this`. my personal (some others too) opinion, but if we're just testing one condition, I find if statements easier to read and just as fast

kwm4385:
Can now redirect to the output file when a command launches.

SingularityUI/app/handlebarsHelpers.coffee
kwm4385: ^ Fixed. Commit didn't close because it's on the line below.

tpetr:
@kwm4385 Can we just completely hide the options stuff if you've selected a command that doesn't define any?

screen shot 2015-10-06 at 11 36 37 am

tpetr:
We should also make sure to only show this if a task is running under the Singularity executor.

kwm4385:
I assumed the field we are looking at to determine if the task is running on the singularity executor is `data.task.taskRequest.deploy.customExecutorCmd` being something like `/usr/local/bin/singularity-executor`. Let me know if this isn't the case.

tpetr:
(merged into #515)

PR #685, opened 2015-09-10
show decom button for frozen slaves

tpetr:
Already supported in the API -- the button was just missing in the original PR.


tpetr:
this isn't working exactly right -- need to take another stab

tpetr:
^ ignore that, i had forgotten to push my merge to hs_stable

PR #688, opened 2015-09-15
tweaks to upload logs with proper file extensions

tpetr:
- updates dateformat to include file extension (i.e. what was once `service.log-20150915.gz` is now `service.log-201509-15.log.gz`)


ssalinas:
This will have some logfetch impacts. Once of the things folks use is a search for 'service.log' or 'access.log' which looks for that string in the file name. We'd need to make sure to have logfetch take the file extension off, put a matcher for the date in and search for that string instead since the original file name is not longer present in full

mjball:
:heart_eyes_cat:

tpetr:
thoughts on leaving the filename unchanged and changing `dateformat` from `-%Y%m%d` to `-%Y%m%d.EXT` (where EXT is the extension)? or is that too hacky?

PR #689, opened 2015-09-15
fix more unix cron schedules

wsorenson:


tpetr:
@wsorenson is this g2g?

wsorenson:
:ship:

PR #692, opened 2015-09-16
grab thread count via cgroups

tpetr:
Locates the executor's cgroup via `/proc/EXECUTOR_PID/cgroup`, and then calculates the total number of threads in use by all procs inside cgroup via `/cgroup/cpu/CONTAINER/tasks`.


eabbott:
Does this weed out Processes that aren't under the mesos cgroup path? /cgroup/<cpu>/mesos/*

tpetr:
@eabbott updated PR body to reflect what it does. This counts the # of threads in use by all processes under the executor's cgroup.

eabbott:
Can you use /proc/self/cgroup?

eabbott:
Ok, read the PR body, that makes sense. :+1:

PR #693, opened 2015-09-17
Docker --privileged flag in SingularityExecutor

ssalinas:
/cc @also


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
```diff
@@ -113,8 +113,8 @@ private ProcessBuilder buildProcessBuilder(TaskInfo taskInfo, ExecutorData execu
     EnvironmentContext environmentContext = new EnvironmentContext(taskInfo);
     if (taskInfo.hasContainer() && taskInfo.getContainer().hasDocker()) {
       task.getLog().info("Writing a runner script to execute {} in docker container", cmd);
-
-      templateManager.writeDockerScript(getPath("runner.sh"), new DockerContext(environmentContext, runnerContext, configuration.getDockerPrefix(), configuration.getDockerStopTimeout()));
+      boolean privileged = (taskInfo.getContainer().getDocker().hasPrivileged() && taskInfo.getContainer().getDocker().getPrivileged());
```
jhaber: I don't think the `hasPrivileged()` check is necessary

ssalinas: Yeah, wasn't sure on that one, I'll remove it and give it a try

PR #694, opened 2015-09-17
Remove paused requests from LB

ssalinas:
Now that Baragon supports the `DELETE` request action for a few releases, this introduces lb cleanup for paused requests. Since our current LB cleanup queue is very task based, it would be difficult to accurately determine from a task's point of view that a request could be deleted form the LB. So, I implemented a request-based lb cleanup queue to go alongside our task-based one.

@tpetr


SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClient.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.List;
 
+import com.google.common.base.Optional;
```
tpetr: unnecessary import

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -269,4 +278,27 @@ public void deleteRequest(SingularityRequest request, Optional<String> user) {
...
+
+  public SingularityCreateResult createLBCleanupRequest(String requestId) {
+    return create(getLBCleanupPath(requestId));
+  }
+
+  public SingularityDeleteResult deleteLBCleanupRequest(String requestId) {
+    return delete(getLBCleanupPath(requestId));
+  }
+
+  public Optional<SingularityLoadBalancerUpdate> getLoadBalancerState(String requestId) {
```
tpetr: what happens if `getLoadBalancer()` state is called after `createLBCleanupRequest()`, but no `saveLoadBalancerState()`?

ssalinas: From what I interpreted it would return an Optional.absent, which would tell the cleanup method that no request has been sent yet and the cleanup still needs to be started

tpetr:
LGTM but I'd like to get @wsorenson's 2 cents on it too

wsorenson:
Are there tests?

ssalinas:
Not yet, I can write some up today though

SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
```diff
@@ -200,4 +206,15 @@ public SingularityLoadBalancerUpdate cancel(LoadBalancerRequestId loadBalancerRe
 
     return sendRequestWrapper(loadBalancerRequestId, LoadBalancerMethod.CANCEL, requestBuilder.build(), BaragonRequestState.UNKNOWN);
   }
+
+  @Override
+  public SingularityLoadBalancerUpdate delete(LoadBalancerRequestId loadBalancerRequestId, SingularityRequest request, SingularityDeploy deploy) {
+    final List<String> serviceOwners = request.getOwners().or(Collections.<String> emptyList());
+    final Set<String> loadBalancerGroups = deploy.getLoadBalancerGroups().or(Collections.<String> emptySet());
+    final BaragonService lbService = new BaragonService(request.getId(), serviceOwners, deploy.getServiceBasePath().get(), loadBalancerGroups, deploy.getLoadBalancerOptions().orNull());
```
wsorenson: Does the service not take optionals ? Seems awkward to have to transition between Optionals and empty collections here.

wsorenson:
What is the difference between deleting request vs deleting tasks in Baragon?

On unpause, will task additions automatically recreate what's been deleted? (I see no corresponding call)

ssalinas:
A `DELETE` for baragon removes the config completely from the load balancer. An `UPDATE` just adds or removes upstreams (which is what we do for tasks). If a new task is started after the service was already deleted, it will recreate everything

wsorenson:
So do we issue a DELETE on request removal?

ssalinas:
Yes, `DELETE` for requests removal, `UPDATE` for anything to do with tasks

ssalinas:
Added a test for the request lb cleanup queue and one small tweak for processing the cleanup

PR #696, opened 2015-09-18
Upgrade to basepom 12.6

jhaber:
@tpetr @ssalinas @wsorenson


wsorenson:
Can you usher this slowly @ssalinas

ssalinas:
sure thing, will do the baragon one as well

jhaber:
Baragon one will have a lot more version changes because its moving 2 major versions of basepom ahead, this should be a lot safer

PR #698, opened 2015-09-21
Logfetch date fixes

ssalinas:
more inclusive date range when hh:mm:ss not specified, fix date parsing bug in logsearch


ssalinas:
Also bumped the gevent version. 1.0.2 will no longer install when a newer version of clang is present due to issues with libev. ie. if you use a mac and you've updated it recently, you'll get errors...

PR #704, opened 2015-09-23
Optionally verify S3 artifacts via detached signatures

jonathanwgoodwin:


SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/ArtifactManager.java
```diff
@@ -87,6 +97,37 @@ private void downloadAndCheck(RemoteArtifact artifact, Path downloadTo) {
 
     checkFilesize(artifact, downloadTo);
     checkMd5(artifact, downloadTo);
+    checkGpgSignature(artifact, downloadTo);
+  }
+
+  private void checkGpgSignature(RemoteArtifact baseArtifact, Path downloadTo) {
+    if (baseArtifact.getGpgSignatureArtifact().isPresent() && baseArtifact.getGpgSignatureArtifact().get() instanceof RemoteArtifact) {
```
wsorenson: Always better to reduce indentation. If you have an if block around an entire method, you should either split the inner contents into its own method or negate the IF and return (my preference.)

SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/ArtifactManager.java
```diff
@@ -87,6 +97,37 @@ private void downloadAndCheck(RemoteArtifact artifact, Path downloadTo) {
...
+
+      Path signatureArtifactPath =  fetch((RemoteArtifact) baseArtifact.getGpgSignatureArtifact().get());
```
wsorenson: Casting is pretty sucky. Why not just add this field to the top-level artifact class if we're going to check it here?

SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/ArtifactManager.java
```diff
@@ -87,6 +97,37 @@ private void downloadAndCheck(RemoteArtifact artifact, Path downloadTo) {
 
     checkFilesize(artifact, downloadTo);
     checkMd5(artifact, downloadTo);
+    checkGpgSignature(artifact, downloadTo);
+  }
+
+  private void checkGpgSignature(RemoteArtifact baseArtifact, Path downloadTo) {
```
wsorenson: This method probably needs logging.

SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/config/SingularityS3Configuration.java
```diff
@@ -67,6 +67,23 @@
...
+  @NotEmpty
+  @JsonProperty
+  private String gpgBinaryPath = "/usr/bin/gpg";
+
+  @NotEmpty
+  @Obfuscate
+  @JsonProperty
+  private String gpgKeyPassword;
+
+  @NotEmpty
```
wsorenson: As these are NotEmpty does it imply gpg is now a "required" part of the Singularity S3 configuration?

ssalinas:
:ship:

PR #706, opened 2015-09-28
Add timers to some critical functions

wsorenson:


SingularityService/src/main/java/com/hubspot/singularity/guice/BootstrapAwareModule.java
```diff
@@ -0,0 +1,29 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import com.google.inject.Binder;
+import com.google.inject.Module;
+
+import io.dropwizard.setup.Bootstrap;
+
+public abstract class BootstrapAwareModule implements Module {
+  private Bootstrap<?> bootstrap = null;
```
tpetr: `ConfigurationAwareModule` (which this is based off of) had this variable marked as `volatile` -- not sure if necessary or not

wsorenson: @jhaber ?

jhaber: afaik startup happens in a single thread so that shouldn't be necessary but it couldn't hurt and performance is not an issue here

tpetr: added in https://github.com/HubSpot/Singularity/commit/842ff3e20493446468c5165ccc67b8747693a21a just to be safe

PR #707, opened 2015-09-28
Previous name field for LB

ssalinas:
wire up a `previousName` field that will go to Baragon's `replaceServiceId`. This means that we can have once service replace another in the load balancer with no downtime. Before we would need to stop one, delete it from Baragon, then deploy the new one. Now it's all one step

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
@@ -56,6 +56,7 @@ public class SingularityDeploy {
   private final Optional<String> serviceBasePath;
   private final Optional<List<String>> loadBalancerGroups;
   private final Optional<Map<String, Object>> loadBalancerOptions;
+  private final Optional<String> previousName;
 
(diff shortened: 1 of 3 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
```diff
@@ -56,6 +56,7 @@
   private final Optional<String> serviceBasePath;
   private final Optional<List<String>> loadBalancerGroups;
   private final Optional<Map<String, Object>> loadBalancerOptions;
+  private final Optional<String> previousName;
```
tpetr: `previousName` is too ambiguous for my taste -- i'd follow the Baragon field naming and call it something like `loadBalancerReplaceServiceId`

ssalinas:
@tpetr agreed on the naming, but I also realized something that unfortunately makes this much more complicated. If the service being replaced in the lb is still running when it is replaced (which is kind of the point), the load balancer removal for that service is guaranteed to fail when you eventually shut it down due to base path conflicts. Going to have to think about how is best to handle this

ssalinas:
going to be much more complicated than I though, closing this for now

PR #708, opened 2015-09-28
Show extra cmd line arguments and allow rerunning of tasks

kwm4385:
New UI features:
- Show the command line arguments that were run on the task and deploy detail pages, under "Info" section.
- Add a Rerun button (↺) in the task history table for on-demand tasks that will open the run now dialog with the cmd line arguments box prepopulated with what that task originally ran with.
- In the dialog shown after clicking "Run Now" you can now use the up and down arrow keys to scroll though argument history for that request.

### DIFF
--- a/SingularityUI/app/models/Request.coffee
+++ b/SingularityUI/app/models/Request.coffee
@@ -183,25 +183,25 @@ class Request extends Model
                 localStorage.setItem(@localStorageCommandLineInputKeyPrefix + "historyIndex", 0);
                 localStorage.setItem(@localStorageCommandLineInputKeyPrefix + "historyLength", commands.length);
 
-                $('#commandLineInput').keydown (e) =>
-                    inc = 0
-                    switch e.which
-                      when 38
-                          inc = 1 # up
-                      when 40
-                          inc = -1 # down
-                      else
-                          return
-                    index = parseInt(localStorage.getItem(@localStorageCommandLineInputKeyPrefix + 'historyIndex'))
-                    length = parseInt(localStorage.getItem(@localStorageCommandLineInputKeyPrefix + 'historyLength'))
-                    index += inc
-                    if index < 0
-                        index = 0
-                    else if index > length - 1
-                        index = length - 1
-                    localStorage.setItem @localStorageCommandLineInputKeyPrefix + 'historyIndex', index
-                    $('#commandLineInput').val commands[index]
-                    e.preventDefault()
+                # $('#commandLineInput').keydown (e) =>
+                #     inc = 0
+                #     switch e.which
+                #       when 38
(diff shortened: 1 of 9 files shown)


SingularityUI/app/templates/vex/requestRun.hbs
```diff
@@ -12,3 +12,34 @@
 <div class="vex-dialog-input">
 	<input id='filename' name="filename" type="text" class="vex-dialog-prompt-input" placeholder="e.g. service.log" value="">
 </div>
+
+<script>
```
benrlodge: i'm sure we can find a way to get this out of the markup.

kwm4385: Yeah I'll do another pass through on this.

SingularityUI/app/models/Request.coffee
```diff
@@ -152,7 +157,16 @@ class Request extends Model
                     return false
 
                 else
-                    localStorage.setItem(@localStorageCommandLineInputKeyPrefix + @id, commandLineInput) if commandLineInput?
+                    history = localStorage.getItem(@localStorageCommandLineInputKeyPrefix + @id)
```
benrlodge: it would be nice if we could move all this local storage related stuff out to some utility file. it always bugged me that we already have all this vex stuff in a model file, maybe we can tackle that in another issue, but it would be nice if we could lean this out a bit.

SingularityUI/app/templates/vex/requestRun.hbs
benrlodge: give me a shout if you get stuck - there's a few different ways to use vex, some more shorthand than others (you can see different examples throughout singularity) and the vex docs obviously as well could be useful to read through http://github.hubspot.com/vex/

SingularityUI/app/models/Request.coffee
kwm4385: I noticed that too. It would make sense move localstorage and vex interactions into utilities. We should probably make another issue though since it would add a lot of clutter.

tpetr:
i'd like to roll the rerun button out, but I think we need more work on the argument history stuff. could we move those changes into a separate PR?

kwm4385:
Would disabling the history stuff (but still keeping the code in) be good enough to merge this and then start another PR?

tpetr:
@kwm4385 yeah that should be OK

kwm4385:
@tpetr Good to go.

tpetr:
looks good, but task rerun dialog doesnt seem to work for me -- never gets past "waiting for task to launch"

tpetr:
^ we'll address the auto-tail issue in a different PR

PR #709, opened 2015-09-29
Highlight decomissioning history states

kwm4385:


tpetr:
can you include an example screenshot?

kwm4385:
image

wsorenson:
Should we have some kind of message bar that floats in at the top?

kwm4385:
That could also work. Would it be in place of the current highlighting or in combination?

kwm4385:
Rolled into task alerts PR, closing this one in favor of that.

https://github.com/HubSpot/Singularity/pull/721

PR #713, opened 2015-10-01
Give meaningful titles to all pages

kwm4385:


SingularityUI/app/controllers/DeployDetail.coffee
```diff
@@ -20,6 +20,8 @@ class DeployDetailController extends Controller
     healthChecks:       require '../templates/deployDetail/deployHealthChecks'
 
   initialize: ({@requestId, @deployId}) ->
+    @title @deployId
```
tpetr: I don't think the deploy ID is descriptive enough here -- something like `"#{ @requestId } deploy #{ @deployId }"` would make more sense

kwm4385:
:+1:

PR #717, opened 2015-10-05
thread checker, case where docker container has already stopped

ssalinas:
@tpetr found our docker threads issue. In the case where we inspect the container, but it has already stopped, the pid returns as 0. This is due to the docker daemon's response to inspect, not the client's processing. The wrapper checks every few seconds for the container to be running, so a badly timed thread check will throw up the error we saw with /proc/0/cgroup

So, I've added a check that if the pid is 0, we check that it is running. If it isn't running, we return 0. No need for throwing sentry errors when we are about to shut down anyways and nothing is really wrong

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorThreadChecker.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorThreadChecker.java
@@ -115,7 +115,7 @@ public class SingularityExecutorThreadChecker {
         String containerName = String.format("%s%s", configuration.getDockerPrefix(), taskProcess.getTask().getTaskId());
         int possiblePid = dockerClient.inspectContainer(containerName).state().pid();
         if (possiblePid == 0) {
-          LOG.warn(String.format("Container %s has pid 0. Running: (%s). Will not try to get threads", containerName, dockerClient.inspectContainer(containerName).state().running()));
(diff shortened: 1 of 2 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorThreadChecker.java
```diff
@@ -113,7 +113,13 @@ private int getNumUsedThreads(SingularityExecutorTaskProcessCallable taskProcess
     if (taskProcess.getTask().getTaskInfo().hasContainer() && taskProcess.getTask().getTaskInfo().getContainer().hasDocker()) {
       try {
         String containerName = String.format("%s%s", configuration.getDockerPrefix(), taskProcess.getTask().getTaskId());
-        dockerPid = Optional.of(dockerClient.inspectContainer(containerName).state().pid());
+        int possiblePid = dockerClient.inspectContainer(containerName).state().pid();
+        if (possiblePid == 0) {
+          LOG.warn(String.format("Container %s has pid 0. Running: (%s). Will not try to get threads", containerName, dockerClient.inspectContainer(containerName).state().running()));
```
tpetr: i'd suggest `"Container %s has pid %s (running: %s). Defaulting to 0 threads running."`

ssalinas:
@tpetr updated the message

PR #721, opened 2015-10-06
Task alerts

kwm4385:
Show an alert on the task page if it is a scheduled task that has been running for longer than twice the average for tasks in the request. /cc @tpetr 

screen shot 2015-10-06 at 3 59 44 pm

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
@@ -56,8 +56,6 @@ public class UIConfiguration {
   private boolean hideNewDeployButton = false;
   private boolean hideNewRequestButton = false;
 
-  private Optional<Integer> warnIfScheduledJobIsRunningPastNextRunPct;
-
   /**
    * If true, the root of the server (http://.../singularity/) will open the UI. Otherwise,
    * the UI URI (http://.../singularity/ui/) must be used.
@@ -152,11 +150,4 @@ public class UIConfiguration {
     this.taskS3LogOmitPrefix = taskS3LogOmitPrefix;
   }
 
-  public Optional<Integer> getWarnIfScheduledJobIsRunningPastNextRunPct() {
-    return warnIfScheduledJobIsRunningPastNextRunPct;
-  }
-
-  public void setWarnIfScheduledJobIsRunningPastNextRunPct(Optional<Integer> warnIfScheduledJobIsRunningPastNextRunPct) {
-    this.warnIfScheduledJobIsRunningPastNextRunPct = warnIfScheduledJobIsRunningPastNextRunPct;
-  }
 }
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
@@ -5,6 +5,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import java.util.Locale;
 
 import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
 
 import org.hibernate.validator.constraints.NotEmpty;
@@ -62,6 +63,8 @@ public class UIConfiguration {
 
   private boolean hideNewDeployButton = false;
   private boolean hideNewRequestButton = false;
+
+  private Optional<Integer> warnIfScheduledJobIsRunningPastNextRunPct;
 
   /**
    * If true, the root of the server (http://.../singularity/) will open the UI. Otherwise,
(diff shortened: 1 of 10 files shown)


kwm4385:
Alert if a task is killed due to slave decommissioning.

screen shot 2015-10-08 at 11 53 59 am

SingularityUI/app/controllers/TaskDetail.coffee
```diff
@@ -123,6 +132,38 @@ class TaskDetailController extends Controller
                 app.caughtError()
                 delete @models.resourceUsage
 
+    getAlerts: (requestTaskHistory) =>
+        task = @models.task
+        alerts = []
+        # Is this a scheduled task that has been running much longer than previous ones?
```
tpetr: instead of calculating this yourself, you should use `averageRuntimeMillis` in the [SingularityDeployStatistics](https://github.com/HubSpot/Singularity/blob/master/SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeployStatistics.java#L27) object, which can be grabbed from the deploy history endpoint (`.../api/history/request/REQUEST_ID/deploy/DEPLOY_ID`). it would also be wise to surface `warnIfScheduledJobIsRunningPastNextRunPct` from https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java#L172 into the window.config object so that we don't have to hardcode the threshold value

SingularityUI/app/controllers/TaskDetail.coffee
```diff
@@ -123,6 +132,38 @@ class TaskDetailController extends Controller
...
+
+            avg = Math.round avg / times.length
+            current =  new Date().getTime() - task.get('task').taskId.startedAt
+            # Alert if current uptime is longer than twice the average
+            if current > (avg * 2)
+                alerts.push
+                  title: 'Warning:',
+                  message: 'This scheduled task has been running longer than twice the average for the request and may be stuck.',
+                  level: 'warning'
+        # Was this task killed by a discomissioning slave?
```
tpetr: typo

kwm4385:
Updated

SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
```diff
@@ -63,6 +64,8 @@ public String getFinishedTaskLogPath() {
   private boolean hideNewDeployButton = false;
   private boolean hideNewRequestButton = false;
 
+  private Optional<Integer> warnIfScheduledJobIsRunningPastNextRunPct;
```
tpetr: i don't think this is necessary -- using the value inside SingularityConfiguration is probably good enough

SingularityUI/app/controllers/TaskDetail.coffee
```diff
@@ -123,6 +132,34 @@ class TaskDetailController extends Controller
...
+        # Is this a scheduled task that has been running much longer than previous ones?
+        if task.attributes.task.taskRequest.request.requestType == 'SCHEDULED' and task.get('isStillRunning')
+            avg = deployInfo.get('deployStatistics')?.averageRuntimeMillis
+            current =  new Date().getTime() - task.get('task').taskId.startedAt
+            threshold = window.config.warnIfScheduledJobIsRunningPastNextRunPct / 100
+            # Alert if current uptime is longer than twice the average
+            if current > (avg * threshold)
+                alerts.push
+                  title: 'Warning:',
+                  message: 'This scheduled task has been running longer than twice the average for the request and may be stuck.',
```
tpetr: can you fix the message to mention the percentage instead of "twice" ?

SingularityUI/config.coffee
```diff
@@ -53,6 +53,7 @@ exports.config =
             runningTaskLogPath:  process.env.SINGULARITY_RUNNING_TASK_LOG_PATH ? "stdout"
             finishedTaskLogPath: process.env.SINGULARITY_FINISHED_TASK_LOG_PATH ? "stdout"
             commonHostnameSuffixToOmit: process.env.SINGULARITY_COMMON_HOSTNAME_SUFFIX_TO_OMIT ? ""
+            warnIfScheduledJobIsRunningPastNextRunPct: process.env.WARN_IF_SCHEDULED_JOB_IS_RUNNING_PAST_NEXT_RUN_PCT ? 200
```
tpetr: prefix this with `SINGULARITY_` to match all the others

SingularityUI/app/controllers/TaskDetail.coffee
```diff
@@ -135,6 +172,17 @@ class TaskDetailController extends Controller
         @models.task.fetch()
             .done =>
                 @fetchResourceUsage() if @models.task.get('isStillRunning')
+            .success =>
+                requestId = @models.task.attributes.task.taskRequest.request.id
+                deployId = @models.task.attributes.task.taskRequest.deploy.id
+                deployInfo = new DeployDetails
+                  deployId: deployId
+                  requestId: requestId
+                deployInfo.fetch().success =>
```
tpetr: technically we could only make this call only if it's needed (i.e. if the task is scheduled + currently running)

PR #724, opened 2015-10-07
allow JAR name format to be overridden

tpetr:
For example, set `-Dsingularity.jar.name.format=\${project.artifactId}` as a maven arg to generate JARs without the version in the filename.

/cc @ssalinas @jhaber


ssalinas:
:+1:

PR #727, opened 2015-10-08
Fix tooltip positioning on edit request page

kwm4385:
/cc @tpetr


tpetr:
overall LGTM, but i noticed when the tootip appears, the button gets ~1px wider

kwm4385:
^ Was due to a bootstrap bug with button groups. Made a workaround.

tpetr:
:dancer: thanks

PR #729, opened 2015-10-09
optionally take start and end time query params for s3 log search

ssalinas:
this is just an initial thought, not sure if it is the best way to do this but wanted to open for feedback. This is related to our logfetch issue and the fact that `/logs/request/{requestId}` and even `/logs/request/{requestId}/deploy/{deployId}` will consistently time out for a service with too many logs. (Think scheduled job that runs every ~10 mins). Also combating this on logfetch side, but this could definitely help

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
@@ -19,6 +19,7 @@ import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 
 import org.jets3t.service.S3Service;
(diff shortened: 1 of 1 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
```diff
@@ -235,7 +257,10 @@ private void checkS3() {
   @GET
   @Path("/task/{taskId}")
   @ApiOperation("Retrieve the list of logs stored in S3 for a specific task.")
-  public List<SingularityS3Log> getS3LogsForTask(@ApiParam("The task ID to search for") @PathParam("taskId") String taskId) throws Exception {
+  public List<SingularityS3Log> getS3LogsForTask(
+      @ApiParam("The task ID to search for") @PathParam("taskId") String taskId,
+      @ApiParam("Start timestamp (millis, 13 digit)") @QueryParam("start") long start,
```
wsorenson: Can they not be optional in here?

ssalinas: Not sure you can use `Optional` in `@QueryParam`, I think if the query param is not specified it will always come back as a null no matter what the class. I'll double check though

tpetr: the query params will be zero if not specified (since they're primitive types), so `fromNullable()` wont work

PR #731, opened 2015-10-12
close files, less dns hits in logfetch

ssalinas:
Two fixes that will help us for downloading large numbers of logs with logfetch:
- requests does not cache dns, so we will only resolve each hostname once and save it
- with large numbers of logs to download, we can hit the open files limit (specifically on OSX). Noticed we weren't explicitly closing as we went along with the downloads, so now we are

Also for ease of use:
- better error in logtail when the requested logfile isn't found
- if we can't find the log file, also attempt to return a list of the available files for use in `-l / --logfile` option

### DIFF
--- a/scripts/logfetch/callbacks.py
+++ b/scripts/logfetch/callbacks.py
@@ -14,6 +14,7 @@ def generate_callback(request, destination, filename, chunk_size, verbose):
         sys.stderr.write(colored('Downloaded ', 'green') + colored(path, 'white') + '\n')
       else:
         sys.stderr.write(colored('.', 'green'))
+      f.close()
 
   return callback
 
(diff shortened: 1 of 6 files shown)


scripts/logfetch/callbacks.py
```diff
@@ -14,6 +14,7 @@ def callback(response, **kwargs):
         sys.stderr.write(colored('Downloaded ', 'green') + colored(path, 'white') + '\n')
       else:
         sys.stderr.write(colored('.', 'green'))
+      f.close()
```
tpetr: pretty sure the `with` block should auto-close -- are you sure this is necessary?

ssalinas:
things I thought would help the issues, did not. Will open a new pr once I do some more digging

PR #733, opened 2015-10-15
Logfetch User Friendliness

ssalinas:
- Better errors when tailing
- Use the new start/end parameters when fetching s3 logs from singularity api
- Add a progress bar and better messaging on progress of downloads/unpacking/finding files


tpetr:
LGTM thanks

PR #734, opened 2015-10-15
Change allocated cpu units to floats in the json object

kwm4385:
/cc @tpetr


tpetr:
could you use `double` to match the original protobuf? https://github.com/apache/mesos/blob/0.23.0/include/mesos/mesos.proto#L555-L650

kwm4385:
Can do

tpetr:
@kwm4385 after looking closer, those two fields are supposed to be `long`, not `double`, so I dont think they're causing the resources UI bug. made appropriate changes in commit above.

kwm4385:
Does this solve the original issue?

PR #736, opened 2015-10-21
Add blazar configs

jhaber:
Alternatively, we could change the default in the POM to this format. I don't think that would pose much of a backwards-compatibility issue because people would only pick up the change when they upgrade to Singularity 0.4.6 which will require changing the JAR name either way (since the current format includes the version)

@tpetr


tpetr:
going to revisit this down the road

PR #738, opened 2015-10-21
add + wire up mysql support in docker integration test environment

tpetr:


SingularityServiceIntegrationTests/pom.xml
```diff
@@ -185,7 +190,7 @@
                       <MESOS_MASTER>zk://singularity-test-zk:2181/mesos</MESOS_MASTER>
                       <MESOS_CONTAINERIZERS>docker,mesos</MESOS_CONTAINERIZERS>
                       <MESOS_HOSTNAME>${mesos.slave2.hostname}</MESOS_HOSTNAME>
-                      <MESOS_PORT>5052</MESOS_PORT>
+                      <MESOS_PORT>5051</MESOS_PORT>
```
ssalinas: what's your reasoning on changing the ports around here? We made it that way originally because we ran with net=HOST and ports were overlapping. Right now you are mapping mesos.slave2.port to port 5052 in the container:

```
<ports>
    <port>mesos.slave2.port:5052</port>
</ports>
```
 so the mesos process will be unreachable on 5051

SingularityServiceIntegrationTests/pom.xml
```diff
@@ -10,6 +10,11 @@
 
   <artifactId>SingularityServiceIntegrationTests</artifactId>
 
+  <properties>
+    <singularity.mysql.password>hellscape</singularity.mysql.password>
```
ssalinas: :+1:

SingularityServiceIntegrationTests/pom.xml
tpetr: my mistake -- need to update that setting too. the reason why i'm changing ports is that I noticed that we have slave ports hardcoded in `MesosClient`: https://github.com/HubSpot/Singularity/blob/master/SingularityMesosClient/src/main/java/com/hubspot/mesos/client/MesosClient.java#L28

we can fix the hardcoded port in a separate PR

PR #739, opened 2015-10-23
Add request and task links to breadcrumbs on tail view

kwm4385:
screen shot 2015-10-23 at 11 47 33 am

I'm open to discussion on where to place these links. Putting them on their own row above the breadcrumbs took up too much precious vertical space, and making the links their full names instead of Request and Task would end up pushing out any longer file names.

### DIFF
--- a/SingularityUI/app/styles/tailer.styl
+++ b/SingularityUI/app/styles/tailer.styl
@@ -42,7 +42,7 @@
 .tail-contents
     overflow scroll
     position absolute
-    top 130px
+    top 178px
     left 0
     bottom 0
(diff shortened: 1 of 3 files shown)


tpetr:
what do you think about this?
screen shot 2015-10-28 at 5 34 21 pm

In this way, we're consistent with the breadcrumbs in the task page. If we're smart about padding / margins i think it should be OK

kwm4385:
@tpetr Moved them to their own row to match the other pages.

screen shot 2015-10-29 at 11 15 55 am

PR #740, opened 2015-10-23
Fuzzy search

kwm4385:
The search on tables and in the global search is much better now. Not only does it not have to make requests every search, but as long as you vaguely slap your hands on the keyboard in a similar pattern to what you want, it'll find it. 
```
> testservice Oreo? 
> oh you mean ORION_TestServiceDocker-web of course!
```

### DIFF
--- a/SingularityUI/app/views/globalSearch.coffee
+++ b/SingularityUI/app/views/globalSearch.coffee
@@ -1,5 +1,7 @@
 View = require './view'
 
+Requests = require '../collections/Requests'
+
 class GlobalSearchView extends View
 
     lastSearchQuery: ''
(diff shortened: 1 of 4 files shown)


tpetr:
what exactly are those `threshold` values, and where'd you get 0.6 and 0.4 from?

kwm4385:
`threshold` is at what point does the match algorithm give up. A threshold of 0.0 requires a perfect match (of both letters and location), a threshold of 1.0 would match anything. 
Setting the values was mostly guess and check to strike a balance with sensitivity and number of relevant results. The global search has a higher value (0.6) because it's only searching one field (name) instead of name and usernames/emails in the table so it can afford to be more loose.

PR #743, opened 2015-10-30
thread pool of 1 if prefixes is 0, avoid IllegalArgument

ssalinas:
@tpetr

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
@@ -238,7 +238,7 @@ public class S3LogResource extends AbstractHistoryResource {
   }
 
   private List<SingularityS3Log> getS3Logs(Optional<String> group, Collection<String> prefixes) throws InterruptedException, ExecutionException, TimeoutException {
-    if (prefixes.size() == 0) {
+    if (prefixes.isEmpty()) {
       return Collections.emptyList();
     }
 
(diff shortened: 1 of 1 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
```diff
@@ -238,7 +238,7 @@ public SingularityS3Log call() throws Exception {
   }
 
   private List<SingularityS3Log> getS3Logs(Optional<String> group, Collection<String> prefixes) throws InterruptedException, ExecutionException, TimeoutException {
-    ListeningExecutorService executorService = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(Math.min(prefixes.size(), configuration.get().getMaxS3Threads()),
+    ListeningExecutorService executorService = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(Math.min(prefixes.size() == 0 ? 1 : prefixes.size(), configuration.get().getMaxS3Threads()),
```
tpetr: idk why i suggested that route -- can't we just short-circuit to `Collections.emptyList()` if there are no prefixes?

ssalinas: you and your logical ideas... :stuck_out_tongue_winking_eye: I'll update

SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
```diff
@@ -238,6 +238,10 @@ public SingularityS3Log call() throws Exception {
   }
 
   private List<SingularityS3Log> getS3Logs(Optional<String> group, Collection<String> prefixes) throws InterruptedException, ExecutionException, TimeoutException {
+    if (prefixes.size() == 0) {
```
wsorenson: isEmpty() is preferable I believe.

Issue #636, opened 2015-07-27
Capture task placement decisions in task JSON


stevenschlansker:
Currently, when a task is being placed, the only debugging aid is a bunch of `LOG.trace` statements e.g.
```
LOG.trace("Rejecting OPTIMISTIC task {} from slave {} ({}) due to numOnSlave {}", taskRequest.getRequest().getId(), slaveId, host, numOnSlave);
```

This means if a task cannot be placed (possibly failing a deploy as `OVERDUE`) you have little recourse until you are willing to roll a new logging config that enables trace, retry the deploy, and then disable it when you are done.

It would be nice if decisions such as this were collected in the task JSON, similar to how the healthcheck results are, so end-users can get some self-service diagnostics (even if the UI is not very pretty at the start).

wsorenson:
True, there are a few other cases where you must see the logs instead of interpreting API response (particularly when load balancer + deploys fail.)

ssalinas:
After adding #893 we have more of this information. Any failed deploys will now have a failure reason enum, one of the values of which is `TASK_COULD_NOT_BE_SCHEDULED` and has a message related to the fact that there might not be enough resources. Other failure reasons are served up as well and available on the deploy page and in the deployResult json

As for the tracking of the lacking resources as they happen, that is what #898 gets at it seems, so I am going to close this issue and we can track the progress of surfacing overdue task information better in that issue

PR #747, opened 2015-11-06
Attribute placement

ssalinas:
Going to complete this a bit more and add some better messaging so we can know when/why tasks might be/not be getting placed. Maybe along the lines of something that will help solve #636 as well

Overall adds a few things:
- Singularity will keep track of all attributes for a slave, not just rack
- you can reserve slaves with certain attributes (i.e. only run tasks that request them on those slaves)
- you can have tasks only run on salves that match the attributes they request
- added unmatched offer information to the pending task json so we can see why a task might not be getting launched


ssalinas:
@tpetr slimmed down the info in SingularityOfferState. Now, it just has the PendingTaskId, and either the offer id and host is was placed on or reasons that offers were declined for that task

ssalinas:
Ok, changed up a number of things here:
- Got rid of the /offerState node and corresponding ui
- I am now storing a list of unmatched offers in the PendingTask object (updated fresh each time we process offers). This is a map of `SlaveMatchState` to a list of hosts that returned that slave match state. This should be helpful enough to give us some insight in the json view when a pending task seems to not be launching

screen shot 2015-11-18 at 12 59 57 pm

/cc @tpetr

SingularityBase/src/main/java/com/hubspot/singularity/SingularityPendingTask.java
```diff
@@ -40,13 +45,18 @@ public boolean apply(@Nonnull SingularityPendingTask input) {
...
+
   @JsonCreator
   public SingularityPendingTask(@JsonProperty("pendingTaskId") SingularityPendingTaskId pendingTaskId, @JsonProperty("cmdLineArgsList") List<String> cmdLineArgsList,
-      @JsonProperty("user") Optional<String> user, @JsonProperty("runId") Optional<String> runId) {
+      @JsonProperty("user") Optional<String> user, @JsonProperty("runId") Optional<String> runId, @JsonProperty("unmatchedOffers") Map<SlaveMatchState, List<String>> unmatchedOffers) {
     this.pendingTaskId = pendingTaskId;
     this.user = user;
     this.cmdLineArgsList = JavaUtils.nonNullImmutable(cmdLineArgsList);
     this.runId = runId;
+    this.unmatchedOffers = unmatchedOffers == null ? new HashMap<SlaveMatchState, List<String>>() : unmatchedOffers;
```
tpetr: `EnumMap<>` would be better to use here

ssalinas:
This has been i use and stable for a while, think we are good to merge this?

ssalinas:
Commenting so that I remember:
The reserveSlavesWithAttributes should be a Map<String, List<String>> so that we can reserve slaves with multiple different values for a certain attribute

ssalinas:
@tpetr updated to take a list of values for reserved slave attributes

ssalinas:
@tpetr added some tests in for the scheduler now as well, let me know if you have any other changes you'd like before we merge this in

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -112,11 +91,36 @@ public SlaveMatchState doesOfferMatch(Protos.Offer offer, SingularityTaskRequest
...
+      } else {
+        String message = String.format("Slaves with attributes %s are reserved for matching tasks. No attributes specified for task %s", reservedSlaveAttributes, taskRequest.getPendingTask().getPendingTaskId().getId());
+        LOG.trace(message);
+        return SlaveMatchState.SLAVE_ATTRIBUTES_DO_NOT_MATCH;
+      }
+    }
+
+    if (taskRequest.getRequest().getRequiredSlaveAttributes().isPresent()
+      && !slaveAndRackHelper.hasRequiredAttributes(slaveAndRackHelper.getTextAttributes(offer), taskRequest.getRequest().getRequiredSlaveAttributes().get())) {
+      String message = String.format("Task requires slave with attributes %s, (slave attributes are %s)", taskRequest.getRequest().getRequiredSlaveAttributes().get(), slaveAndRackHelper.getTextAttributes(offer));
```
wsorenson: Don't use String.format()

Use log.( {} )

SingularityService/src/test/java/com/hubspot/singularity/SingularitySchedulerTestBase.java
```diff
@@ -158,11 +161,28 @@ protected Offer createOffer(double cpus, double memory, String slave, String hos
...
+    List<Attribute> attributesList = new ArrayList<>();
+    for (Map.Entry<String, String> entry : attributes.entrySet()) {
+      try {
+        attributesList.add(Attribute.newBuilder()
+            .setType(Type.TEXT)
+            .setName(entry.getKey())
+            .setText(Text.newBuilder().setValue(entry.getValue()).build())
+            .build());
+      } catch (Exception e) {
+        Throwables.propagate(e);
```
wsorenson: I think you want to throw this.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosModule.java
```diff
@@ -8,6 +9,7 @@
 import com.google.inject.Scopes;
 import com.google.inject.Singleton;
 import com.google.inject.name.Named;
+import com.hubspot.singularity.SingularityTaskRequest;
```
wsorenson: Are these unused imports?

SingularityService/src/test/java/com/hubspot/singularity/SingularitySchedulerTestBase.java
jhaber: the method throws it

PR #752, opened 2015-11-09
Error handling

ssalinas:
- Don't throw s3 uploader err to sentry unless all retries fail
- Catch NoSuchFileException in ExecutorCleanup
- Properly catch ExecutionException so we throw back SlaveNotFoundException when a slave is no longer running

@tpetr

### DIFF
--- a/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/config/MissingConfigException.java
+++ b/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/config/MissingConfigException.java
@@ -1,7 +1,7 @@
 package com.hubspot.singularity.runner.base.config;
 
 public class MissingConfigException extends IllegalStateException {
-  public MissingConfigException(Exception e) {
-    super(e);
+  public MissingConfigException(String message) {
+    super(message);
   }
 }
--- a/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/config/MissingConfigException.java
+++ b/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/config/MissingConfigException.java
@@ -1,0 +1,7 @@
+package com.hubspot.singularity.runner.base.config;
+
+public class MissingConfigException extends IllegalStateException {
+  public MissingConfigException(Exception e) {
+    super(e);
+  }
+}
--- a/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/config/MissingConfigException.java
+++ b/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/config/MissingConfigException.java
@@ -1,6 +1,6 @@
 package com.hubspot.singularity.runner.base.config;
 
-public class MissingConfigException extends IllegalStateException {
+public class MissingConfigException extends RuntimeException {
   public MissingConfigException(String message) {
     super(message);
   }
--- a/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/JsonObjectFileHelper.java
+++ b/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/JsonObjectFileHelper.java
@@ -28,15 +28,17 @@ public class JsonObjectFileHelper {
 
     log.info("Reading {}", file);
 
-    byte[] bytes = Files.readAllBytes(file);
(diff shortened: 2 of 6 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
```diff
@@ -55,7 +56,7 @@ public SlaveNotFoundException(Exception e) {
       }
 
       return objectMapper.readValue(response.getResponseBodyAsStream(), MESOS_FILE_OBJECTS);
-    } catch (ConnectException ce) {
+    } catch (ExecutionException | ConnectException ce) {
```
tpetr: the issue it that `ConnectException` is the inner exception of `ExecutionException` -- i'd suggest checking the `getCause()` value for `ConnectException`, and throwing SNFE if so and `Throwables.propagate()` if not

SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
```diff
@@ -91,7 +92,7 @@ public SlaveNotFoundException(Exception e) {
       }
 
       return Optional.of(objectMapper.readValue(response.getResponseBodyAsStream(), MesosFileChunkObject.class));
-    } catch (ConnectException ce) {
+    } catch (ExecutionException | ConnectException ce) {
```
tpetr: ditto for here

ssalinas: Ah gotcha, was reading it wrong as the type of exception was different, I can update that

SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/JsonObjectFileHelper.java
```diff
@@ -28,15 +28,17 @@ public JsonObjectFileHelper(ObjectMapper objectMapper) {
...
+      bytes = Files.readAllBytes(file);
 
-    if (bytes.length == 0) {
-      return Optional.absent();
-    }
+      log.trace("Read {} bytes from {} in {}", bytes.length, file, JavaUtils.duration(start));
+
+      if (bytes.length == 0) {
+       return Optional.absent();
+      }
```
tpetr: instead of these checks, i'd just explicitly handle `NoSuchFileException` in the try-catch

SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/JsonObjectFileHelper.java
```diff
@@ -28,17 +29,21 @@ public JsonObjectFileHelper(ObjectMapper objectMapper) {
 
     log.info("Reading {}", file);
 
-    byte[] bytes = Files.readAllBytes(file);
+    byte[] bytes = new byte[0];
```
tpetr: nitpick -- i dont think we need to bother giving this a default, we could just move the variable into the try-catch

ssalinas: we get a compile time error if we don't set it at all since we use it in the catch block.

tpetr: nevermind, forgot `bytes` was being used in the catch

tpetr:
LGTM thanks

SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/config/MissingConfigException.java
```diff
@@ -0,0 +1,7 @@
+package com.hubspot.singularity.runner.base.config;
+
+public class MissingConfigException extends IllegalStateException {
```
tpetr: not sure `IllegalStateException` is the best exception:

> Signals that a method has been invoked at an illegal or inappropriate time.

Might be best if we just subclass from `RuntimeException`

ssalinas: Yeah that works, I had done IllegalState to start with because thats what Preconditions was throwing before. I'll update to Runtime

PR #755, opened 2015-11-12
remove fuzzy matching in requests and tasks pages

tpetr:
the Fuse library can't handle long search queries (ex. searching for tasks on specific slaves) -- removing from the requests and tasks pages for now

@kwm4385


kwm4385:
@tpetr I increased the max pattern length to 128 for all the searches and put that limit on the inputs so there won't be any errors. I think this should cover all of our use cases.

PR #760, opened 2015-11-13
Docker shell commands

ssalinas:
@tpetr make shell commands work in docker:
- There are often multiple processes in docker, read the pid we want to operate on from a file if it exists (i.e. so we can write the pid of a java process under a wrapper/supervisor to the file, and operate on the correct pid)
- if not otherwise specified, pid in the docker container will be 1
- add a `switchUser` configuration to the descriptor. Most times we will want the command to run as the same user as the task, so defaults to true. When true we will prepend the user switch already set in our config
- If the task is docker, the executor will insert a `docker exec (container id)` after the user switch (if in use). So now the only part of the command you specify in the config is the command itself, the executor takes care of the context


ssalinas:
closing this and re-opening against master now that shell commands is merged

PR #761, opened 2015-11-13
fix npe when check for exception cause

ssalinas:
@tpetr


SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
```diff
@@ -99,7 +99,7 @@ public SlaveNotFoundException(Exception e) {
     } catch (ConnectException ce) {
       throw new SlaveNotFoundException(ce);
     } catch (Exception e) {
-      if (e.getCause().getClass() == ConnectException.class) {
+      if (e.getCause()!= null && e.getCause().getClass() == ConnectException.class) {
```
tpetr: insane nitpick -- can you add a space before `!==`? might also be good to wrap the two equality checks in parens for readability

tpetr:
:+1:

PR #762, opened 2015-11-16
Docker shell

ssalinas:
New PR against master for shell commands with docker

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
@@ -247,6 +247,9 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   @JsonProperty
   private String shellCommandUserPlaceholder = "{USER}";
 
+  @JsonProperty
+  private String shellCommandPidFile = ".shell_command_pid";
+
   public SingularityExecutorConfiguration() {
     super(Optional.of("singularity-executor.log"));
   }
@@ -607,6 +610,14 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
     this.shellCommandOutFile = shellCommandOutFile;
   }
 
+  public String getShellCommandPidFile() {
+    return shellCommandPidFile;
+  }
+
+  public void setShellCommandPidFile(String shellCommandPidFile) {
+    this.shellCommandPidFile = shellCommandPidFile;
+  }
+
   @Override
   public String toString() {
     return "SingularityExecutorConfiguration[" +
@@ -654,6 +665,7 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
             ", shellCommandOutFile='" + shellCommandOutFile + '\'' +
             ", shellCommandPidPlaceholder='" + shellCommandPidPlaceholder + '\'' +
             ", shellCommandUserPlaceholder='" + shellCommandUserPlaceholder + '\'' +
+            ", shellCommandPidFile='" + shellCommandPidFile + '\'' +
             ']';
   }
 
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
@@ -248,7 +248,7 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   private String shellCommandUserPlaceholder = "{USER}";
 
   @JsonProperty
-  private String shellCommandPidFile = ".shell_command_pid";
+  private String shellCommandPidFile = ".task-pid";
 
   public SingularityExecutorConfiguration() {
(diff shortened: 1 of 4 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/shells/SingularityExecutorShellCommandDescriptor.java
```diff
@@ -22,6 +24,10 @@
   @NotNull
   private List<SingularityExecutorShellCommandOptionDescriptor> options = Collections.emptyList();
 
+  @JsonProperty
+  @NotNull
+  private boolean switchUser = true;
```
tpetr: can we make this more descriptive? (i.e. `runAsTaskUser`)

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/shells/SingularityExecutorShellCommandRunner.java
```diff
@@ -119,14 +124,39 @@ public boolean apply(SingularityExecutorShellCommandDescriptor input) {
 
     final SingularityExecutorShellCommandDescriptor shellCommandDescriptor = matchingShellCommandDescriptor.get();
 
-    List<String> command = new ArrayList<>(shellCommandDescriptor.getCommand());
+    List<String> command = new ArrayList<>();
+
+    if (shellCommandDescriptor.isSwitchUser()) {
+      String switchUserCommand = String.format(executorConfiguration.getSwitchUserCommandFormat(), taskProcess.getTask().getExecutorData().getUser().or(executorConfiguration.getDefaultRunAsUser()));
+      command.addAll(Arrays.asList(switchUserCommand.split("\\s+")));
```
tpetr: seems a little janky, might be better to explicitly define what the switch user command for shell commands is

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/shells/SingularityExecutorShellCommandRunner.java
```diff
@@ -119,14 +124,39 @@ public boolean apply(SingularityExecutorShellCommandDescriptor input) {
...
 
     for (int i = 0; i < command.size(); i++) {
       if (command.get(i).equals(executorConfiguration.getShellCommandPidPlaceholder())) {
-        if (!taskProcess.getCurrentPid().isPresent()) {
-          throw new InvalidShellCommandException("No PID found");
+        int pid;
+        Path pidFilePath = MesosUtils.getTaskDirectoryPath(getTask().getTaskId()).resolve(executorConfiguration.getShellCommandPidFile());
+        if (Files.exists(pidFilePath)) {
+          try {
+            pid = Integer.parseInt(new Scanner(pidFilePath).useDelimiter("\\Z").next());
```
tpetr: what is `\\Z` ?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -247,6 +247,9 @@
   @JsonProperty
   private String shellCommandUserPlaceholder = "{USER}";
 
+  @JsonProperty
+  private String shellCommandPidFile = ".shell_command_pid";
```
tpetr: this is the PID file for the command that the shell command should operate on, right? `.shell_command_pid` seems a little ambiguous to me, can we make it clearer? maybe `.task-pid`?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/shells/SingularityExecutorShellCommandRunner.java
ssalinas: Does it make sense to have another switch user command when we already configure the one that the executor uses for everything else? That's why I went with this originally.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/shells/SingularityExecutorShellCommandRunner.java
ssalinas: `\Z` is the end of string anchor, used that so we are essentially just read the whole file as one string

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
ssalinas: :+1: updated

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/shells/SingularityExecutorShellCommandRunner.java
```diff
@@ -119,14 +123,36 @@ public boolean apply(SingularityExecutorShellCommandDescriptor input) {
...
+          } catch (Exception e) {
+            throw new InvalidShellCommandException("No PID found");
```
tpetr: may want to clarify why the PID wasn't found by including the exception message

tpetr:
:ship:

PR #764, opened 2015-11-17
Prevent 'remove' from being double bound

kwm4385:
The Kill Task button is already being bound to the action in the overview subview. Remove the redundant event in the parent view.
@tpetr


SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -152,7 +152,7 @@ Handlebars.registerHelper 'isRunningState', (list, options) ->
             options.inverse(@)
 
 Handlebars.registerHelper 'isSingularityExecutor', (value, options) ->
-    if value.indexOf 'singularity-executor' != -1
+    if value and value.indexOf 'singularity-executor' != -1
```
tpetr: is this still necessary now that #765 is merged?

tpetr: nevermind, was thinking of something else

PR #766, opened 2015-11-18
Bounce deny

ssalinas:
For `OPTIMISTIC` and `GREEDY` it shouldn't matter. But, if placement is `SEPARATE*`, don't allow us to start a bounce we know we can't complete

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
@@ -209,7 +209,7 @@ public class RequestResource extends AbstractRequestResource {
     if (placement != SlavePlacement.GREEDY && placement != SlavePlacement.OPTIMISTIC) {
       int currentActiveSlaveCount = slaveManager.getNumObjectsAtState(MachineState.ACTIVE);
       int requiredSlaveCount = requestWithState.getRequest().getInstancesSafe() * 2;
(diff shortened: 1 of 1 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -197,6 +205,13 @@ public SingularityRequestParent bounce(@ApiParam("The request ID to bounce") @Pa
 
     checkConflict(requestWithState.getState() != RequestState.PAUSED, "Request %s is paused. Unable to bounce (it must be manually unpaused first)", requestWithState.getRequest().getId());
 
+    SlavePlacement placement = requestWithState.getRequest().getSlavePlacement().or(configuration.getDefaultSlavePlacement());
+    if (placement != SlavePlacement.GREEDY && placement != SlavePlacement.OPTIMISTIC) {
+      int currentActiveSlaveCount = slaveManager.getNumObjectsAtState(MachineState.ACTIVE);
+      int requiredSlaveCount = requestWithState.getRequest().getInstancesSafe() * 2;
+      checkBadRequest(currentActiveSlaveCount >= requiredSlaveCount, "Not enough slaves to successfully complete a bounce of reqeust %s (minimum required: %s, current: %s)", requestId, requiredSlaveCount, currentActiveSlaveCount);
```
tpetr: can you clarify this to `Not enough active slaves...` and maybe include guidance to consider deploying instead of bouncing or changing the slave placement?

PR #767, opened 2015-11-19
Introduce an incremental bounce

ssalinas:
Currently, in order to bounce a request which has `SEPARATE*` placement, you need to be running instance_count * 2 slaves. This can be quite a pain for requests with many instances. This PR creates a new type of bounce where we shut down old tasks as new ones are spun up (vs waiting for all new tasks to be ready first). Using this, you only need a minimum of instance_count + 1 slaves to bounce the same request

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
@@ -133,7 +133,7 @@ public class SingularityCleaner {
     // For an incremental bounce, shut down old tasks as new ones are started
     if (taskCleanup.getCleanupType() == TaskCleanupType.INCREMENTAL_BOUNCE) {
       String key = String.format("%s-%s", taskCleanup.getTaskId().getRequestId(), taskCleanup.getTaskId().getDeployId());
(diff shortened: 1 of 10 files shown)


ssalinas:
Got this working well in staging, also added an option for it in the UI
screen shot 2015-11-24 at 4 48 01 pm

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
```diff
@@ -362,9 +386,19 @@ private void drainTaskCleanupQueue() {
       return;
     }
 
+    Map<String, Integer> incrementalBounceRemainingInstanceMap = new HashMap<>();
```
tpetr: can you clarify what this map does? it tracks how many tasks still need to bounce per request?

ssalinas: When doing our cleanup, each task doesn't know what has happened to the task before it. This way we can keep track of how many cleaning tasks were already killed in this cleanup run. (i.e. if we have room to kill one cleaning task in our incremental bounce, now we'll know when we get to the second that we've already done that and we shouldn't kill the second)

tpetr:
LGTM

PR #775, opened 2015-11-23
If inside a HTTP request, include URL in sentry error

tpetr:
Will update PR with full implementation soon -- wanted to get @jhaber's eyes on the request stash stuff.


SingularityService/src/main/java/com/hubspot/singularity/jersey/RequestStash.java
```diff
@@ -0,0 +1,25 @@
...
+import com.google.common.base.Optional;
+
+public enum RequestStash {
+  INSTANCE;
+
+  private final ThreadLocal<String> url;
+
+  RequestStash() {
+    this.url = new ThreadLocal<>();
+  }
```
jhaber: nit: ditch the constructor and initialize inline

jhaber:
lgtm, afaik Guice doesn't have anything to make this easier. The other option would be to have a provider that catches the `OutOfScopeException`, ie, something like this:

```
  @Provides
  @Named("request.path")
  public Optional<String> providesRequestPath(Provider<ContainerRequest> request) {
    try {
      return Optional.of(request.get().getPath());
    } catch (OutOfScopeException e) {
      return Optional.absent();
    }
  }
```

This is simpler in some ways, but the drawback is you need to be very careful to always inject a `Provider<Optional<String>>` (if you accidentally inject `Optional<String>` into a singleton without wrapping in a provider, it will cache an empty value forever and be hard to track down)

tpetr:
@jhaber gotcha, thanks. I originally hit issues with your OOSE-catching approach in a standalone provider, but I think I just made a mistake on my end because it's working now. I'm not too worried about the only using `Provider<>` caveat because we're only using the provider in one location.

jhaber:
cool lgtm

PR #777, opened 2015-11-24
Add data attribute

kwm4385:
Sort requests properly on timestamp.


kwm4385:
Taking a look at how new rows are appended when loading more.

kwm4385:
@tpetr All the tables should be sorting properly now. It should stay sorted when you load more as well.

kwm4385:
Much better now.

PR #781, opened 2015-11-24
Add new favicon

kwm4385:
Just shrunk down the one we use for auxy (https://s3.amazonaws.com/uploads.hipchat.com/10819/63596/SWxrLvGqYJyjwuo/singularity.png)

In the future (when we have someone better at photoshop than me) we should make a clean, vector one like Blazar's.


tpetr:
paging @wolfd

kwm4385:
:bagel:

wolfd:
@tpetr I think I sent you a few bad vectors I had created for Singularity right before I handed in my laptop. Do you still have those? If not, I may be able to whip something up.

tpetr:
@wolfd i remember seeing them, but i can't seem to find them :(

PR #783, opened 2015-11-25
tweaks to make grepping a file easier

tpetr:
- Don't grep if the grep query param is empty
- Add optional `nextOffset` field to `MesosFileChunkObject`, and if grepping, relay what the next file offset should be (needed for the UI). If `nextOffset` is not set, it's safe to assume the next offset you should grab is `offset + data.length`.

/cc @kwm4385


kwm4385:
:dancer:

PR #786, opened 2015-12-03
add support for default healthcheckMaxRetries and healthcheckMaxTotalTimeoutSeconds values in SingularityConfiguration

tpetr:
/cc @wsorenson #604


SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -102,6 +102,12 @@
 
   private long healthcheckTimeoutSeconds = 5;
 
+  @NotNull
+  private Optional<Integer> healthcheckMaxRetries = Optional.absent();
```
wsorenson: I like that you used Optionals here. Previously, I had set Integers or Longs to -1 to disable them. This is better.

PR #788, opened 2015-12-07
fix router base path when hosted at /

tpetr:
I believe this fixes the issue you had with hosting Singularity at `/`, @scalp42. When you have a chance, could you let me know if this works for you?

### DIFF
--- a/SingularityUI/app/application.coffee
+++ b/SingularityUI/app/application.coffee
@@ -42,7 +42,7 @@ class Application
 
         # so sneaky
         el = document.createElement('a')
-        el.href = config.appRoot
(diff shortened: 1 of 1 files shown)


scalp42:
I *should* be able to test this tomorrow (need to deploy it again)

scalp42:
@tpetr sorry for the late delay here but I can report that it's still "broken" on Singularity 0.4.9 unfortunately:

tpetr:
@scalp42 no worries -- 0.4.9 was cut before this PR was merged, so it'll be in the 0.4.10 release.

PR #791, opened 2015-12-08
Fix search input

kwm4385:
After rendering move the cursor to the end of the input and keep it focused if there is text.

@wsorenson

### DIFF
--- a/SingularityUI/app/views/view.coffee
+++ b/SingularityUI/app/views/view.coffee
@@ -49,10 +49,4 @@ class View extends Backbone.View
           next: '<span class="glyphicon glyphicon-chevron-right"></span>'
       $('table.paginated').css('display', 'table');
 
-      $searchInput = $('.big-search-box')
-      strLength = $searchInput.val().length * 2
-      if strLength > 0
-        $searchInput.focus()
-        $searchInput[0].setSelectionRange(strLength, strLength)
-
 module.exports = View
--- a/SingularityUI/app/views/view.coffee
+++ b/SingularityUI/app/views/view.coffee
@@ -50,7 +50,7 @@ class View extends Backbone.View
       $('table.paginated').css('display', 'table');
 
       searchInput = $('.big-search-box')
-      strLength= searchInput.val().length * 2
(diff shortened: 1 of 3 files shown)


wsorenson:
Does this prevent it from occasionally erasing what was in there?

kwm4385:
While I wasn't able to repro that myself, I think it's due to it being overwritten as you type and the refresh happens. If that still happens after this is applied let me know.

SingularityUI/app/views/view.coffee
```diff
@@ -49,4 +49,10 @@ class View extends Backbone.View
           next: '<span class="glyphicon glyphicon-chevron-right"></span>'
       $('table.paginated').css('display', 'table');
 
+      searchInput = $('.big-search-box')
+      strLength = searchInput.val().length * 2
```
tpetr: why `* 2` here?

kwm4385: We want to make sure it always ends up at the end, there's some cases where length would fail us (Opera sometimes sees a carriage return as 2 characters)

tpetr:
This could still have issues if the cursor is in the middle of the text -- maybe we can just get the Search box to not re-render/

SingularityUI/app/views/view.coffee
```diff
@@ -49,4 +49,10 @@ class View extends Backbone.View
           next: '<span class="glyphicon glyphicon-chevron-right"></span>'
       $('table.paginated').css('display', 'table');
 
+      searchInput = $('.big-search-box')
```
benrlodge: small, but i would rename to `$searchInput` for clarity

kwm4385:
@tpetr Looking at the code it seems it won't be that easy. The table is a subview of the search, making it the top level element. If we don't re-render that, nothing else can. :/

kwm4385:
I did some more research and found a better solution for this. Now it'll save your caret position before rendering and reset it afterwards. This will also preserve any selections as well.

tpetr:
:boom: thanks

PR #795, opened 2015-12-10
Add option to also bounce when scaling

ssalinas:
screen shot 2015-12-10 at 10 12 58 am
(extra options for incremental bounce only appear when bounce is checked)

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
@@ -42,6 +42,7 @@ public class SingularityRequest {
 
   private final Optional<String> group;
   private final Optional<Set<String>> readOnlyGroups;
+  private final Optional<Boolean> bounceAfterScale;
 
   @JsonCreator
   public SingularityRequest(@JsonProperty("id") String id, @JsonProperty("requestType") RequestType requestType, @JsonProperty("owners") Optional<List<String>> owners,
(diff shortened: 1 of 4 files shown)


SingularityUI/app/models/Request.coffee
```diff
@@ -128,8 +128,14 @@ class Request extends Model
             ]
             placeholder: @get 'instances'
             callback: (data) =>
-                return if data is false
-                @scale(data).done callback
+                return unless data
+                bounce = $('.vex #bounce').is ':checked'
+                incremental = $('.vex #incremental-bounce').is ':checked'
+                @scale(data, bounce, incremental).done =>
```
tpetr: how come you're passing `bounce` and `incremental` to `@scale()`?

SingularityUI/app/templates/vex/requestScale.hbs
```diff
@@ -1,2 +1,21 @@
...
+	    <input type="radio" name="incremental" id="incremental-bounce" value="true"> Kill old tasks as new tasks become healthy
+	</label>
+</div>
+<script type="text/javascript">
+	function bounceOptionChanged() {
+	    if($('.vex #bounce').is(":checked"))
+	        $(".vex #incremental-bounce-option").show();
+	    else
+	        $(".vex #incremental-bounce-option").hide();
+	}
```
tpetr: we should figure out a way to do this in the view instead

ssalinas:
@tpetr updated form your comments, also the checkbox label just says "Bounce after scaling" now

ssalinas:
Added a filed in the request so that you can default that box to checked if you want

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
```diff
@@ -229,6 +232,10 @@ public ScheduleType getScheduleTypeSafe() {
     return readOnlyGroups;
   }
 
+  public boolean isBounceAfterScale() {
```
wsorenson: You need one that is the Json getter that just returns the actual value.

If you want to add a helper, you should @JsonIGnore it.

wsorenson:
Is the request option only used by the UI?

ssalinas:
Yes, only used by the ui. Developers wanted a way to have the checkbox default to checked for some requests but not others

tpetr:
This could be addressed in a subsequent PR, but I think it'd be worth packing the bounce + scale functionality into the backend. The "dumber" the UI is, the better.

SingularityUI/app/templates/vex/requestScale.hbs
```diff
@@ -1,2 +1,13 @@
 <p>Enter the desired number of instances to run for request:</p>
 <pre>{{ id }}</pre>
+
+<label for="bounce" id="bounce-label">
+    <input type="checkbox" id="bounce"{{#if bounceAfterScale}} checked{{/if}}>
+    Bounce after scaling
+</label>
```
tpetr: I'd consider moving the bounce checkbox to after the scale input to fit in with the "bounce after scaling" wording.

tpetr: bonus points for indenting the incremental bounce option + greying it out if the bounce checkmark is unchecked

ssalinas: right now the incremental option isn't show if bounce checkmark is unchecked and appears when you check it. would you rather see it greyed out instead of having it hidden?

ssalinas:
New look of the prompt (radio buttons are hidden when bounce isn't checked):
screen shot 2015-12-18 at 8 49 07 am

PR #796, opened 2015-12-10
make files table in task view sortable

ssalinas:


kwm4385:
@ssalinas Just tweaked an attr so we don't sort on the last column.

ssalinas:
Cool, thanks

tpetr:
- Seeing some weirdness when sorting by Size:
  screen shot 2015-12-15 at 11 45 47 am
  screen shot 2015-12-15 at 11 47 25 am
- Task refresh blows away the sort setting

ssalinas:
@tpetr believe I have fixed the odd sorting now, we had some odd stuff going on with data-value which is used by sortable. Probably due to an older way of doing sorting. Still need to fix the task refresh piece

ssalinas:
Actually, sort clearing on task refresh might be part of a larger problem, looks like sorts do that on all of our pages, not just here

tpetr:
LGTM -- let's revisit the sort clearing on refresh in a separate PR

PR #797, opened 2015-12-10
validation constraints for config classes

ssalinas:
Some are probably overkill, but adding in the extra `@NotNull` and `@NotEmpty` to be consistent with what we are doing elsewhere. Will take one more pass at this before I move it up through the branches

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
@@ -143,6 +143,7 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   @JsonProperty
   private String s3UploaderBucket;
 
+  @NotEmpty
   @JsonProperty
   private boolean useLocalDownloadService = false;
 
@@ -157,6 +158,7 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   @JsonProperty
   private String dockerPrefix = "se-";
 
+  @NotNull
   @JsonProperty
   private int dockerStopTimeout = 15;
 
@@ -186,15 +188,19 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   @JsonProperty
   public List<SingularityExecutorShellCommandDescriptor> shellCommands = Collections.emptyList();
 
+  @NotEmpty
   @JsonProperty
   public String shellCommandOutFile = "executor.commands.{TIMESTAMP}.log";
 
+  @NotEmpty
   @JsonProperty
   private String shellCommandPidPlaceholder = "{PID}";
 
+  @NotEmpty
   @JsonProperty
   private String shellCommandUserPlaceholder = "{USER}";
 
+  @NotEmpty
   @JsonProperty
   private String shellCommandPidFile = ".task-pid";
 
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
@@ -143,7 +143,6 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   @JsonProperty
   private String s3UploaderBucket;
 
-  @NotEmpty
   @JsonProperty
   private boolean useLocalDownloadService = false;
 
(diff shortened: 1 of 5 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -143,6 +143,7 @@
   @JsonProperty
   private String s3UploaderBucket;
 
+  @NotEmpty
```
tpetr: `@NotEmpty` and `@NotNull` are unnecessary for `boolean` -- can't be set to `null` since it's a primitive type

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -157,6 +158,7 @@
   @JsonProperty
   private String dockerPrefix = "se-";
 
+  @NotNull
```
tpetr: `int` can't be null, but could could default to zero -- we should decide what a good minimum is, and then use the `@Min` annotation

tpetr: There's a bunch of these below -- can you replace all `@NotNull`s with `@Min`s that make sense?

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -161,9 +161,11 @@
   private long saveStateEverySeconds = 60;
 
   @JsonProperty("sentry")
+  @Valid
```
tpetr: `sentryConfiguration` and `smtpConfiguration` can be `null` -- could you sanity check that a `null` value annotated with `@Valid` is legal? (i.e. we don't want it to fail validation unless it also has a `@NotNull`)

ssalinas: The sample/docker singularity runs with both of these as null, started up with no issues

ssalinas:
Added some sensible mins and maxes for other config classes as well

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/config/SingularityS3UploaderConfiguration.java
```diff
@@ -42,15 +43,19 @@
   @Obfuscate
   private Optional<String> s3SecretKey = Optional.absent();
 
+  @Max(5368709120L)
```
tpetr: i'm worried more about min values than max values here -- (i.e. a huge number isn't a big deal, but a negative value would probably screw things up)

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/config/SingularityS3UploaderConfiguration.java
```diff
@@ -42,15 +43,19 @@
...
   @JsonProperty
   private long maxSingleUploadSizeBytes = 5368709120L;
 
+  @Min(5242880L)
   @JsonProperty
   private long uploadPartSize = 20971520L;
 
+  @Min(500)
```
tpetr: how'd you decide on 500? probably fine, but maybe we should just make the minimum 0 so there's no arbitrary cutoff

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
tpetr: sweet

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/config/SingularityS3UploaderConfiguration.java
ssalinas: yeah min of 0 is fine there

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/config/SingularityS3UploaderConfiguration.java
ssalinas: added a Min of 5242880L here, any smaller than that and there isn't a way multipart will work since that's also the minimum part size for multipart

PR #798, opened 2015-12-10
(logfetch) make a silent mode, activate if not a tty

ssalinas:
Behave more like curl where if we realize output is being redirected, we do not output stderr/progress messages. There is a `--silent` option that will do this, and it is automatically on when output is being redirected. The only stderr that will still write is for error messages (i.e. when the stdout output would not be useful/complete anyways)

/cc @mjball


tpetr:
:+1:

PR #800, opened 2015-12-11
ability to edit a request without unpausing it

ssalinas:


SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -149,7 +149,11 @@ public SingularityRequestParent submit(@ApiParam("The Singularity request to cre
 
     final long now = System.currentTimeMillis();
 
-    requestManager.activate(newRequest, maybeOldRequest.isPresent() ? RequestHistoryType.UPDATED : RequestHistoryType.CREATED, now, JavaUtils.getUserEmail(user));
+    if (maybeOldRequestWithState.isPresent() && maybeOldRequestWithState.get().getRequest().isLongRunning()) {
```
tpetr: if i'm reading this right, paused cron jobs will still get un-paused. i'd welcome @wsorenson's 2 cents on this, but i think we should only call `activate()` if the request didn't previously exist

ssalinas: There must have been a reason for it at some point. We actually have a test that ensures that they do get activated on request update

ssalinas: Actually, not as explicitly as I remembered, `testSchedulerHandlesFinishedTasks` uses requestResource.submit to cause tasks to be pending, so slightly different

wsorenson: let's hold on this one.

wsorenson:
This is better addressed in #810

PR #801, opened 2015-12-14
Link each column section

kwm4385:
On the status page, link each section in the column to the corresponding requests/tasks page.


tpetr:
I think this might already be implemented somehow -- a build of Singularity without this PR merged in seemed to have the same links on the status page already.

tpetr:
nevermind, i misunderstood the PR

PR #804, opened 2015-12-14
Add tasks from /killed to decom badges

kwm4385:


SingularityUI/app/models/TaskKill.coffee
```diff
@@ -0,0 +1,18 @@
+Model = require './model'
+
+class TaskKill extends Model
```
tpetr: i'd be more explicit and name it `TaskKillRecord`

tpetr:
:dancer:

tpetr:
Depends on #802

PR #807, opened 2015-12-14
Bounce tests

ssalinas:
Adds in some tests for the feature implemented in #767 and some helper methods we should be able to reuse for any other tests that need more than 2 slaves

/cc @wsorenson @tpetr


tpetr:
LGTM -- @wsorenson ?

PR #810, opened 2015-12-15
Expiring actions

wsorenson:
- Allows bounce & other actions to skip healthChecks
- Allows bounce, pause, scale, and skipHealthchecks commands to expire.
- Introduces request-level skipHealthchecks setting

- [x] Send email when scale expires?
- [x] Expire / cancel bounce
- [x] Need UI for skipping healthchecks at request level


kwm4385:
@wsorenson UI should be all set now!

PR #813, opened 2015-12-17
Get task info by runId

ssalinas:
Get the `SingularityTaskIdHistory` object for a task by it's run id

/cc @jhaber this should get you back taskId, last state, and the SingularityTaskId object

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
@@ -29,6 +29,9 @@ public interface HistoryJDBI {
   @SqlQuery("SELECT bytes FROM taskHistory WHERE taskId = :taskId")
   byte[] getTaskHistoryForTask(@Bind("taskId") String taskId);
 
+  @SqlQuery("SELECT bytes FROM taskHistory WHERE runId = :runId")
+  byte[] getTaskHistoryForTaskByRunId(@Bind("runId") String runId);
+
   @SqlQuery("SELECT bytes FROM deployHistory WHERE requestId = :requestId AND deployId = :deployId")
   byte[] getDeployHistoryForDeploy(@Bind("requestId") String requestId, @Bind("deployId") String deployId);
 
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
@@ -29,8 +29,8 @@ public interface HistoryJDBI {
   @SqlQuery("SELECT bytes FROM taskHistory WHERE taskId = :taskId")
   byte[] getTaskHistoryForTask(@Bind("taskId") String taskId);
 
-  @SqlQuery("SELECT bytes FROM taskHistory WHERE runId = :runId")
-  byte[] getTaskHistoryForTaskByRunId(@Bind("runId") String runId);
+  @SqlQuery("SELECT bytes FROM taskHistory WHERE requestId = :requestId AND runId = :runId")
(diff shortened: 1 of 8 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
```diff
@@ -29,6 +29,9 @@
   @SqlQuery("SELECT bytes FROM taskHistory WHERE taskId = :taskId")
   byte[] getTaskHistoryForTask(@Bind("taskId") String taskId);
 
+  @SqlQuery("SELECT bytes FROM taskHistory WHERE runId = :runId")
+  byte[] getTaskHistoryForTaskByRunId(@Bind("runId") String runId);
```
jhaber: Shouldn't this also filter by requestId?

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -153,6 +153,18 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
   }
 
   @GET
+  @Path("/request/{requestId}/tasks/{runId}")
```
jhaber: Should this be `/request/{requestId}/run/{runId}`?

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -153,6 +153,18 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
...
+  @ApiOperation("Retrieve the history for all tasks of a specific request.")
```
jhaber: This doesn't sound right

SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
```diff
@@ -29,6 +29,9 @@
   @SqlQuery("SELECT bytes FROM taskHistory WHERE taskId = :taskId")
   byte[] getTaskHistoryForTask(@Bind("taskId") String taskId);
 
+  @SqlQuery("SELECT bytes FROM taskHistory WHERE requestId = :requestId AND runId = :runId")
+  byte[] getTaskHistoryForTaskByRunId(@Bind("requestId") String requestId, @Bind("runId") String runId);
```
wsorenson: do we index on runId?

ssalinas:
Updated a couple things, had forgotten about the description, thanks for that

SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
ssalinas: not currently, we would probably need to for this case

tpetr: yeah, we should add an index. let me know if you need a hand with this.

ssalinas: Added an index in  b63cf47 , can someone with more mysql chops double check that please?

tpetr: syncing up w/ monty about this

solomonty: Which query/queries will benefit from the added index?

ssalinas: The one on the line for this comment thread:
`SELECT bytes FROM taskHistory WHERE requestId = :requestId AND runId = :runId`

tpetr:
This PR reads like there was an assumption that `runId` would be unique, but I don't see any constraint for that in #669. Did I miss something, or was this an oversight?

PR #815, opened 2015-12-18
New fuzzy search algorithm for better results and perf

kwm4385:
Switched to a different fuzzy lib that'll give us better results and not nearly as laggy while typing.


kwm4385:
Memoizing the filter further reduced the lag from 320ms to 140ms in applicable situations. :)

PR #816, opened 2015-12-21
Allow optional message to accompany slave decomission

kwm4385:


kwm4385:
@tpetr The UI will now include a `message` field in the post to the decommission endpoint. We just need to save it on the backend and return the message when we get the list of decom'd slaves.

kwm4385:
@wsorenson Added a message field for many request actions. It should also be applicable  here.

tpetr:
This functionality is going to get folded into #810.

PR #817, opened 2015-12-21
Incremental deploys

ssalinas:
Still needs plenty of work, but opening up early for feedback.

Right now I am adding a field to the `SingularityPendingDeploy` object that will track the progress of an incremental deploy as it goes along. This new field is an object that holds information about the current target number of instances and rate at which the deploy should proceed.

Still to-do:
- [x] Better scheduling of upcoming tasks. Right now it's just keying off of numMissingInstances, but since we know the time between deploy steps, we can go ahead and schedule them
- [x] Make it play nice with load balancer add/remove. Regular deploys add everything all at once when the deploy succeeds, this will need to do it per-task
- [x] Surface more information about what is happening in the deploy, if you aren't watching the logs it's kind of mysterious right now
- [x] Option to manually move the deploy forward or backward a step (should be fairly straightforward since we have the info in the `SingularityPendingDeploy` object)
- [x] Tests, tests, tests, tests, tests....
- [x] UI

/cc @tpetr @wsorenson


SingularityBase/src/main/java/com/hubspot/singularity/DeployStyle.java
```diff
@@ -0,0 +1,5 @@
+package com.hubspot.singularity;
+
+public enum DeployStyle {
```
tpetr: I don't want to overcomplicate things, but I wonder about the merits of an enum vs. being able to define a batch size for the deploy that defaults to the total instance count (i.e. a "normal" deploy)

ssalinas: Yeah, that's fair, might keep it simpler in the end, move towards all deploys working the same, but 'regular' deploys are just everything at once

ssalinas:
The basics are working now. All deploys now use a similar system where progress is tracked in the pending deploy object, incrementing a target number of instances for each step of the deploy. By default the number of instances per deploy step is the total instances (i.e. everything at once like before) unless a `deployRate` is set that then determines how many instances to deploy in each 'step'. Load balancer actions are triggered at the end of each step.

Some info is surfaced in the ui now on the request page, but the wording/design could probably still be a bit cleaner.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
```diff
@@ -169,7 +177,10 @@ public SingularityDeployBuilder toBuilder() {
     .setEnv(copyOfMap(env))
     .setUris(copyOfList(uris))
     .setExecutorData(executorData)
-    .setLabels(labels);
+    .setLabels(labels)
+
```
tpetr: ^ can you remove the newlines? can get confusing in a builder block

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
```diff
@@ -134,6 +139,9 @@ public SingularityDeploy(@JsonProperty("requestId") String requestId,
     this.serviceBasePath = serviceBasePath;
     this.loadBalancerGroups = loadBalancerGroups;
     this.loadBalancerOptions = loadBalancerOptions;
+
+    this.deployRate = deployRate;
```
tpetr: kind of a nitpick but `deployRate` seems a little confusing to me. maybe `deployInstanceCountPerStep`? (curious what @wsorenson thinks before you make the change though)

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -50,6 +51,7 @@
 @Api(description="Manages Singularity Deploys for existing requests", value=DeployResource.PATH, position=2)
 public class DeployResource extends AbstractRequestResource {
   public static final String PATH = SingularityService.API_BASE_PATH + "/deploys";
+  public static final int DEFAULT_DEPLOY_STEP_WAIT_TIME_SECONDS = 60;
```
ssalinas: @tpetr was debating making the default wait time 0, would that make more sense? i.e. move on to the next step as soon as the previous is finished 'rolling deploy', by default

tpetr: yeah i think that jives with the incremental deploy idea

ssalinas:
This is working pretty solidly now. Each deploy for a longRunning request has a `SingularityDeployProgress` item on the pending deploy that holds information about the current target number of instances and how the deploy should proceed (timed vs manual step)

Once the deploy is in progress, you will have the option to advance the deploy or cancel it in the UI (advance button not shown if target instances is the same as request instance count):
screen shot 2016-01-19 at 2 21 45 pm


When advancing the deploy, you can then select the next target number of instances (defaults to current target + deployInstancesPerStep) allowing you to step the deploy forward or backwards:

screen shot 2016-01-19 at 2 22 21 pm

The manual advance can also override the timed advance. The deploy checker checks for the presence of a request to manually advance the deploy as well as the expiration of the wait time.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
```diff
@@ -147,29 +158,27 @@ public SingularityDeployBuilder toBuilder() {
...
     .setMetadata(copyOfMap(metadata))
     .setVersion(version)
     .setTimestamp(timestamp)
     .setEnv(copyOfMap(env))
     .setUris(copyOfList(uris))
     .setExecutorData(executorData)
-    .setLabels(labels);
+    .setLabels(labels)
+    .setDeployInstanceCountPerStep(deployInstanceCountPerStep)
+    .setDeployStepWaitTimeSeconds(deployStepWaitTimeSeconds);
```
tpetr: missing `setAutoAdvanceDeploySteps()` here

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeployBuilder.java
```diff
@@ -390,6 +425,8 @@ public String toString() {
       ", serviceBasePath=" + serviceBasePath +
       ", loadBalancerGroups=" + loadBalancerGroups +
       ", loadBalancerOptions=" + loadBalancerOptions +
+      ", deployInstanceCountPerStep=" + deployInstanceCountPerStep +
+      ", deployStepWaitTimeSeconds=" + deployStepWaitTimeSeconds +
```
tpetr: missing `autoAdvanceDeploySteps`

SingularityBase/src/main/java/com/hubspot/singularity/TaskCleanupType.java
```diff
@@ -3,7 +3,7 @@
 public enum TaskCleanupType {
 
   USER_REQUESTED(true, true), USER_REQUESTED_TASK_BOUNCE(false, false), DECOMISSIONING(false, false), SCALING_DOWN(true, false), BOUNCING(false, false), INCREMENTAL_BOUNCE(false, false),
-  DEPLOY_FAILED(true, true), NEW_DEPLOY_SUCCEEDED(true, false), DEPLOY_CANCELED(true, true), UNHEALTHY_NEW_TASK(true, true), OVERDUE_NEW_TASK(true, true);
+  DEPLOY_FAILED(true, true), DEPLOY_STEP_FINISHED(true, false), DEPLOY_CANCELED(true, true), UNHEALTHY_NEW_TASK(true, true), OVERDUE_NEW_TASK(true, true);
```
tpetr: we need to treat enums as append-only in order to maintain backwards compatibility -- removing or renaming values can cause exceptions when deserializing from ZK

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -50,6 +52,7 @@
 @Api(description="Manages Singularity Deploys for existing requests", value=DeployResource.PATH, position=2)
 public class DeployResource extends AbstractRequestResource {
   public static final String PATH = SingularityService.API_BASE_PATH + "/deploys";
+  public static final int DEFAULT_DEPLOY_STEP_WAIT_TIME_SECONDS = 0;
```
tpetr: i'd make this a field in `SingularityConfiguration`

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -134,4 +149,33 @@ public SingularityRequestParent cancelDeploy(
 
     return fillEntireRequest(requestWithState);
   }
+
+  @POST
+  @Path("/deploy/{deployId}/request/{requestId}")
```
tpetr: ordering here feels a little awkward -- i understand that you're following the pattern of our other resources, but it might make more sense flip it to `/deploys/request/{requestId}/deploy/{deployId}`

tpetr: or just do `/deploys/update` since both `requestId` and `deployId` are fields in `SingularityUpdatePendingDeployRequest`

ssalinas:
@tpetr updated from your comments

Also, note to myself, need to add the update endpoint to the SingularityClient

ssalinas:
@wsorenson any thoughts on the PR before it goes further?

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -134,4 +151,30 @@ public SingularityRequestParent cancelDeploy(
...
+
+    Optional<SingularityRequestDeployState> deployState = deployManager.getRequestDeployState(requestWithState.getRequest().getId());
+
+    if (!deployState.isPresent() || !deployState.get().getPendingDeploy().isPresent() || !deployState.get().getPendingDeploy().get().getDeployId().equals(updateRequest.getDeployId())) {
+      throw badRequest("Request %s does not have a pending deploy %s", updateRequest.getRequestId(), updateRequest.getDeployId());
+    }
+
+    if (updateRequest.getTargetActiveInstances() > requestWithState.getRequest().getInstancesSafe()) {
+      throw badRequest("Cannot update pending deploy to have more instances (%s) than instances set for request (%s)");
+    }
```
tpetr: we should also check to make sure nobody is setting `targetActiveInstances` to something less than zero

tpetr: we also have a static method that simplifies this a little more: `checkBadRequest(condition, message, args...)`

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -346,15 +360,17 @@ private SingularityDeployResult fromLbState(DeployState state, SingularityLoadBa
   }
 
   private boolean shouldCheckLbState(final SingularityPendingDeploy pendingDeploy) {
-    return pendingDeploy.getLastLoadBalancerUpdate().isPresent() && (pendingDeploy.getLastLoadBalancerUpdate().get().getLoadBalancerState() != BaragonRequestState.UNKNOWN);
+    return pendingDeploy.getLastLoadBalancerUpdate().isPresent()
+        && getLoadBalancerRequestId(pendingDeploy).getId().equals(pendingDeploy.getLastLoadBalancerUpdate().get().getLoadBalancerRequestId().getId())
```
tpetr: can you clarify what this comparison is for? (i assume related to the incremental deploy, but having a hard time understanding it)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -346,15 +360,17 @@ private SingularityDeployResult fromLbState(DeployState state, SingularityLoadBa
...
+        && (pendingDeploy.getLastLoadBalancerUpdate().get().getLoadBalancerState() != BaragonRequestState.UNKNOWN);
   }
 
-  private LoadBalancerRequestId getLoadBalancerRequestId(SingularityDeployMarker deployMarker) {
-    return new LoadBalancerRequestId(String.format("%s-%s", deployMarker.getRequestId(), deployMarker.getDeployId()), LoadBalancerRequestType.DEPLOY, Optional.<Integer> absent());
+  private LoadBalancerRequestId getLoadBalancerRequestId(SingularityPendingDeploy pendingDeploy) {
+    return new LoadBalancerRequestId(String.format("%s-%s-%s", pendingDeploy.getDeployMarker().getRequestId(), pendingDeploy.getDeployMarker().getDeployId(), pendingDeploy.getDeployProgress().get().getTargetActiveInstances()), LoadBalancerRequestType.DEPLOY, Optional.<Integer> absent());
```
tpetr: could we run into issues if there are LB updates going on while we deploy this out? would it make sense to keep the old request ID format for non-incremental deploys?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
ssalinas: Since there can be multiple load balancer requests for a single deploy, checking that one is present isn't quite enough. Need to check that it is the one for the current step we are on

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -370,74 +386,211 @@ private SingularityDeployResult getDeployResult(final SingularityRequest request
...
-
-      updatePendingDeploy(pendingDeploy, lbUpdate, deployState);
-
-      if (deployState != DeployState.WAITING) {
-        return fromLbState(deployState, lbUpdate);
-      }
+    if (pendingDeploy.getDeployProgress().isPresent()) {
+      return checkDeployProgress(request, cancelRequest, pendingDeploy, updatePendingDeployRequest, deploy, deployActiveTasks, otherActiveTasks);
+    } else {
+      return new SingularityDeployResult(DeployState.FAILED, "No deploy progress present for deploy of long running request");
```
tpetr: This means that any pending deploy(s) when this change is rolled out will fail. It might be worth making a ZK migration for this (but I can also see the argument where that'd probably be overkill). A compromise might be to make this error message a little more specific (since folks would only hit it in this edge case, right?) -- maybe something like `No deploy progress data present in ZooKeeper. Please reattempt your deploy.`?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
ssalinas: Mentioning here what we said over slack. Rather than supporting old an new deploy formats with and without deploy progress data, deploys with missing deploy progress will be rolled back with a message about reattempting the deploy. Updated the messaging from the comment mentioned below

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeployProgress.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class SingularityDeployProgress {
+  private final int targetActiveInstances;
+  private final int deployInstanceCountPerStep;
+  private final long deployStepWaitTimeSeconds;
+  private final boolean stepComplete;
+  private final boolean  autoAdvanceDeploySteps;
+  private final long timestamp;
+
+  public SingularityDeployProgress(@JsonProperty("targetActiveInstances") int targetActiveInstances,
```
wsorenson: <code>@JsonCreator</code>?

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -81,6 +81,8 @@
 
   private boolean defaultValueForKillTasksOfPausedRequests = true;
 
+  private int defaultDeployStepWaitTimeSeconds = 0;
```
wsorenson: I've learned my lesson with seconds. Always make millis so you can test it easier and be consistent.

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -92,7 +97,19 @@ public SingularityRequestParent deploy(@ApiParam(required=true) SingularityDeplo
...
+
+    Optional<SingularityDeployProgress> deployProgress = Optional.absent();
+    if (request.isLongRunning()) {
+      deployProgress = Optional.of(new SingularityDeployProgress(
+          Math.min(deploy.getDeployInstanceCountPerStep().or(request.getInstancesSafe()), request.getInstancesSafe()),
+          deploy.getDeployInstanceCountPerStep().or(request.getInstancesSafe()),
+          deploy.getDeployStepWaitTimeSeconds().or(configuration.getDefaultDeployStepWaitTimeSeconds()),
+          false,
+          deploy.getAutoAdvanceDeploySteps().or(true),
+          System.currentTimeMillis()));
```
wsorenson: use now so timestamps align

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeployProgress.java
```diff
@@ -0,0 +1,62 @@
+package com.hubspot.singularity;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SingularityDeployProgress {
+  private final int targetActiveInstances;
+  private final int deployInstanceCountPerStep;
+  private final long deployStepWaitTimeSeconds;
+  private final boolean stepComplete;
+  private final boolean  autoAdvanceDeploySteps;
```
wsorenson: extra space here.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskCleanup.java
```diff
@@ -5,7 +5,6 @@
 import com.google.common.base.Optional;
 
 public class SingularityTaskCleanup {
-
```
wsorenson: y?

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUpdatePendingDeployRequest.java
```diff
@@ -0,0 +1,37 @@
...
+    return deployId;
+  }
+
+  public int getTargetActiveInstances() {
+    return targetActiveInstances;
+  }
+
+  @Override public String toString() {
+    return "SingularityDeployStepUpdate{" +
+      "requestId='" + requestId + '\'' +
```
wsorenson: why the quotes?

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -92,7 +97,19 @@ public SingularityRequestParent deploy(@ApiParam(required=true) SingularityDeplo
...
     SingularityDeployMarker deployMarker = new SingularityDeployMarker(requestId, deploy.getId(), now, deployUser, deployRequest.getMessage());
-    SingularityPendingDeploy pendingDeployObj = new SingularityPendingDeploy(deployMarker, Optional.<SingularityLoadBalancerUpdate> absent(), DeployState.WAITING);
+
+    Optional<SingularityDeployProgress> deployProgress = Optional.absent();
+    if (request.isLongRunning()) {
+      deployProgress = Optional.of(new SingularityDeployProgress(
+          Math.min(deploy.getDeployInstanceCountPerStep().or(request.getInstancesSafe()), request.getInstancesSafe()),
+          deploy.getDeployInstanceCountPerStep().or(request.getInstancesSafe()),
+          deploy.getDeployStepWaitTimeSeconds().or(configuration.getDefaultDeployStepWaitTimeSeconds()),
+          false,
```
wsorenson: is this a case where using an enum would be better than a boolean?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -92,10 +98,15 @@ public int checkDeploys() {
       LOG.debug("Removing cancel deploy request {} - {}", cancelDeploy, deleteResult);
     }
 
+    for(SingularityUpdatePendingDeployRequest updateRequest : updateRequests) {
```
wsorenson: space after for

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -370,74 +386,211 @@ private SingularityDeployResult getDeployResult(final SingularityRequest request
...
+              return markStepFinished(pendingDeploy, deploy, otherActiveTasks, request, updatePendingDeployRequest);
+            }
+          case UNHEALTHY:
+          default:
+            return new SingularityDeployResult(DeployState.FAILED, "Not all tasks for deploy were healthy");
+        }
+      } else {
+        maybeUpdatePendingRequest(pendingDeploy, deploy, request, updatePendingDeployRequest);
+        return checkOverdue(deploy, deployActiveTasks, isDeployOverdue);
       }
```
wsorenson: This method is too long now. Refactor code into methods that are named appropriately

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUpdatePendingDeployRequest.java
tpetr: i think the default intellij `toString()` template includes quotes around strings

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -569,19 +584,33 @@ private boolean shouldRetryImmediately(SingularityRequest request, SingularityDe
...
 
-  private int getNumMissingInstances(List<SingularityTaskId> matchingTaskIds, SingularityRequest request, SingularityPendingRequest pendingRequest) {
+  private int getNumMissingInstances(List<SingularityTaskId> matchingTaskIds, SingularityRequest request, SingularityPendingRequest pendingRequest, Optional<SingularityPendingDeploy> maybePendingDeploy) {
     if (request.isOneOff() && pendingRequest.getPendingType() == PendingType.ONEOFF) {
       return 1;
     }
 
-    final int numInstances = request.getInstancesSafe();
+    final int numInstances;
+    if (maybePendingDeploy.isPresent() && !(maybePendingDeploy.get().getCurrentDeployState() == DeployState.CANCELED) && maybePendingDeploy.get().getDeployProgress().isPresent()) {
```
wsorenson: a good indicator of a place where you can make a method is when you have:
final XYZ object;

and then a bunch of if code - turn that if code into the method

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -370,74 +386,211 @@ private SingularityDeployResult getDeployResult(final SingularityRequest request
...
+  private void maybeUpdatePendingRequest(SingularityPendingDeploy pendingDeploy, Optional<SingularityDeploy> deploy, SingularityRequest request, Optional<SingularityUpdatePendingDeployRequest> updatePendingDeployRequest) {
+    maybeUpdatePendingRequest(pendingDeploy, deploy, request, updatePendingDeployRequest, Optional.<SingularityLoadBalancerUpdate> absent());
+  }
 
-    if (pendingDeploy.getLastLoadBalancerUpdate().isPresent()) {
-      return new SingularityDeployResult(DeployState.WAITING, Optional.of("Waiting on load balancer API"), pendingDeploy.getLastLoadBalancerUpdate(), System.currentTimeMillis());
+  private void maybeUpdatePendingRequest(SingularityPendingDeploy pendingDeploy, Optional<SingularityDeploy> deploy, SingularityRequest request, Optional<SingularityUpdatePendingDeployRequest> updatePendingDeployRequest, Optional<SingularityLoadBalancerUpdate> lbUpdate) {
+    if (updatePendingDeployRequest.isPresent() && pendingDeploy.getDeployProgress().isPresent()) {
+      SingularityDeployProgress deployProgress = pendingDeploy.getDeployProgress().get();
+      SingularityDeployProgress newProgress = new SingularityDeployProgress(
```
wsorenson: i'm wondering if there might be a way to write a method here, or inside this class, to make these transitions take up less space here in the critical logic code

wsorenson: Like deployProgress.transform/progress/nextStep/clone(pass in changed parameters)

Not sure what the right name is.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -370,74 +386,211 @@ private SingularityDeployResult getDeployResult(final SingularityRequest request
...
-      if (isCancelRequestPresent) {
-        LOG.info("Canceling a deploy {} due to cancel request {}", pendingDeploy, cancelRequest.get());
-        return new SingularityDeployResult(DeployState.CANCELED, String.format("Canceled due to request by %s at %s", cancelRequest.get().getUser(), cancelRequest.get().getTimestamp()));
+        final DeployHealth deployHealth = deployHealthHelper.getDeployHealth(request, deploy, deployActiveTasks, true);
+        switch (deployHealth) {
+          case WAITING:
+            maybeUpdatePendingRequest(pendingDeploy, deploy, request, updatePendingDeployRequest);
+            return checkOverdue(deploy, deployActiveTasks, isDeployOverdue);
+          case HEALTHY:
+            if (request.isLoadBalanced()) {
```
wsorenson: if you can find ways to reduce the massive IF / ELSE indentation on here, great. 

1) through adding additional method calls
2) to inverting - handle the ! case which simply returns something, and then the next code doesn't have to be indented.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUpdatePendingDeployRequest.java
ssalinas: yeah it does, I'll remove them in next commit

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -247,7 +280,16 @@ private boolean isDeployOverdue(SingularityPendingDeploy pendingDeploy, Optional
       return false;
     }
 
-    final long startTime = pendingDeploy.getDeployMarker().getTimestamp();
+    final long startTime;
```
wsorenson: another great method candidate

ssalinas:
@tpetr @wsorenson finished a bunch of refactoring on this to slim down the methods. Also added more tests for canceled/failed deploy scenarios. Let me know if you see anything else in here that's needs updating

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUpdatePendingDeployRequest.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public String getDeployId() {
+    return deployId;
+  }
+
+  public int getTargetActiveInstances() {
+    return targetActiveInstances;
+  }
+
+  @Override public String toString() {
```
tpetr: this is a recent intellij thing that started happening, might not be a setting that the eclipse formatter file sets. i think it'd be good to keep all our `@Override`'s on their own line

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -126,23 +144,21 @@ private void checkDeploy(final SingularityPendingDeploy pendingDeploy, final Lis
...
-
         deleteObsoletePendingTasks(pendingDeploy);
         finishDeploy(requestWithState, deploy, pendingDeploy, allOtherMatchingTasks, deployResult);
         return;
       } else {
         LOG.warn("Failing deploy {} because it failed to save deploy state", pendingDeployMarker);
-        deployResult = new SingularityDeployResult(DeployState.FAILED_INTERNAL_STATE, Optional.of(String.format("Deploy had state %s but failed to persist it correctly", deployResult.getDeployState())), deployResult.getLbUpdate(), deployResult.getTimestamp());
+        deployResult =
+          new SingularityDeployResult(DeployState.FAILED_INTERNAL_STATE, Optional.of(String.format("Deploy had state %s but failed to persist it correctly", deployResult.getDeployState())),
+            deployResult.getLbUpdate(), deployResult.getTimestamp());
```
tpetr: this is obviously a nitpick / not super important, but is this really the kind of formatting we want? this his happening due to a max column width setting?

ssalinas: if we don't care as much about the line length pieces, we should probably update the eclipse formatter settings that are in the repo.

PR #824, opened 2015-12-29
be sure to transfer labels to the task info

ssalinas:
Had thought I wired this all up before, apparently not. This gets the task `labels` from our deploy into the mesos `TaskInfo`


tpetr:
LGTM

PR #827, opened 2015-12-30
be sure to check deploy health for incremental bounce

ssalinas:
Thanks @wsorenson for pointing this out. I reworked/reordered the if statements so that we will check the health of the matching tasks for incremental deploys as well before deciding if we can shut a task down

/cc @tpetr


wsorenson:
You should add a test which fails pre-PR and succeeds post-PR.

ssalinas:
@wsorenson turns out the previous test I wrote for incremental bounce served that pretty well with some modifications. I extended it so that there is an explicit check that a old tasks will not shut down from a task in `TASK_RUNNING` but without a passing healthcheck.

PR #829, opened 2015-12-30
Updated to allow post with params. Support runId when running an ON_DEMAND request

gchomatas:
/cc @tpetr


wsorenson:
This is obsoleted by #810

PR #830, opened 2015-12-30
Updated client to support bounce options

gchomatas:
/cc @tpetr


wsorenson:
You should open a PR onto the branch that introduces the change rather than hs-staging.

PR #832, opened 2015-12-31
automatically run DB migrations at Singularity startup

tpetr:


tpetr:
Going to take another stab at this in the future -- we should be running these migrations when the history persister boots up, not at Singularity startup time.

tpetr:
Closing this for now -- I don't anticipate us getting to this anytime soon.

PR #833, opened 2016-01-04
Incremental bounce task health checking

ssalinas:
- Rework the methods in the deploy health helper so we can use it to get the number of currently healthy tasks for a deploy
- Compare against the number of healthy tasks to determine if we can shut down a cleaning task instead of falling to the deployHealth switch/case like before
- Add a test to check that a task in `TASK_STARTING` does not cause a cleaning task to be killed, but that moving that same task into `TASK_RUNNING` (making it healthy) does cause the cleaning task to be killed

/cc @wsorenson


ssalinas:
This is now in hs_qa. Going to merge into expiring actions and they can be moved to hs_stable together

PR #835, opened 2016-01-05
make rack sensitive placement more even

ssalinas:
Had an idea for a (possibly) simple way to ensure more even distribution across different racks, based on two things:
- We sort tasks to be scheduled such that lower instance number tasks will be scheduled first
- On a scale down we kill highest numbered instances first

This PR will calculate the number allowed on a certain rack so that we make sure we are evenly distributed among all racks before adding additional instances to a rack. For example, if we have 3 active racks and a request with 5 instances. It will first ensure that each rack gets 1 instance before allowing any rack to get 2.

Then, when scaling down, because we added them in this order, we will still have even distribution when killing highest instance numbers first.

The one drawback is it might take more runs of the scheduler to schedule all of the tasks for a request.

@tpetr @wsorenson would love some opinions on this


ssalinas:
Bah... found a hole in this logic where rack placement is uneven if we kill a random instance then scale down. I'll write a test for it and see if I can't update this logic to cover that case as well

ssalinas:
need to rethink this one, closing for now

PR #836, opened 2016-01-06
Bounce task

wsorenson:
- Acts like a bounce for a single task
- Launches replacement task before killing task


wsorenson:
Making it part of #810

PR #837, opened 2016-01-07
make sure literal host ports are processed correctly

ssalinas:
A few things to fix ports/params with docker:
- Make sure any port with a `LITERAL` mapping also gets added to the mesos Resources when building the task
- Verify during `doesOfferMatchResources` that any requested literal host ports are present in the offer
- Add a scheduler test for port resources
- Make sure docker `params` make it onto the mesos task definition and add a MesosUtilsTest assertion for this and the literal ports

/fyi @chrisbrewer @wsorenson


ssalinas:
@tpetr this should be gtg, thoughts on merging?

SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
```diff
@@ -134,27 +131,45 @@ public static Resource getPortsResource(int numPorts, List<Resource> resources)
...
-      portsSoFar += (rangeEndSelection - rangeStartSelection) + 1;
+        portsSoFar += (rangeEndSelection - rangeStartSelection) + 1;
 
-      if (portsSoFar == numPorts) {
-        break;
+        List<Long> toRemove = new ArrayList<>();
+        for (long port : requestedPorts) {
+          if (rangeStartSelection >= port && rangeEndSelection <= port) {
+            toRemove.add(port);
+            portsSoFar --;
```
tpetr: i'd remove the space here for clarity

ssalinas: sry, missed this comment before, fixed

ssalinas:
@tpetr added the JsonIgnore, this should be good now

tpetr:
LGTM

PR #838, opened 2016-01-11
bump handlebars-brunch dep to 1.10.1

tpetr:


tpetr:
Forget this -- looks like the dep had already been bumped in the past.

PR #839, opened 2016-01-11
Task search

wsorenson:
- [x] needs UX


SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -141,7 +143,13 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
   @Path("/request/{requestId}/tasks")
   @ApiOperation("Retrieve the history for all tasks of a specific request.")
   public List<SingularityTaskIdHistory> getTaskHistoryForRequest(
-      @ApiParam("Request ID to look up") @PathParam("requestId") String requestId,
+      @ApiParam("Request ID to match") @PathParam("requestId") String requestId,
+      @ApiParam("Optional deploy ID to match") @QueryParam("deployId") Optional<String> deployId,
+      @ApiParam("Optional host to match") @QueryParam("deployId") Optional<String> host,
```
ssalinas: just noticed this one, @QueryParam is still `deployId` here instead of `host`

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -141,7 +143,13 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
...
+      @ApiParam("Optional last task status to match") @QueryParam("lastTaskStatus") Optional<ExtendedTaskState> lastTaskStatus,
+      @ApiParam("Optionally match only tasks started after") @QueryParam("startedAfter") Optional<Long> startedAfter,
+      @ApiParam("Optionally match only tasks started before") @QueryParam("startedBefore") Optional<Long> startedBefore,
+      @ApiParam("Sort direction") @QueryParam("startBefore") Optional<OrderDirection> orderDirection,
```
ssalinas: Same with this @QueryParam (`orderDirection`)

wsorenson: :100: for code review

mysql/migrations.sql
```diff
@@ -89,3 +89,19 @@ ALTER TABLE `taskHistory` ADD KEY `runId` (`runId`, `requestId`);
 ALTER TABLE `requestHistory` ADD COLUMN message VARCHAR(280) NULL;
 
 ALTER TABLE `deployHistory` ADD COLUMN message VARCHAR(280) NULL;
+
+--changeset wsorenson:10 dbms:mysql
+ALTER TABLE `taskHistory`
+  ADD COLUMN `host` VARCHAR(100) CHARACTER SET ASCII NULL,
+  ADD COLUMN `startedAt` TIMESTAMP NULL,
```
tpetr: we lose precision here (milliseconds --> seconds) -- thoughts on storing this as an integer instead?

tpetr: though I guess if we're just using it for searching it's not that big of a deal

wsorenson: Aren't we using this column type elsewhere?

tpetr: Yep, we are. I was thinking about this in the context of pagination -- `LIMIT :offset, :count` isn't great the further in you go and since the offset isn't a concrete point in time, runs of the persister will make the results randomly update on you. Paging based on `startedAt` would be nice if it was more accurate. Maybe we could look into an auto_increment id? This isn't a super high priority thing, of course, but it might be worth thinking about some more in the future.

Docs/features/task-search.md
```diff
@@ -0,0 +1,23 @@
...
+
+- `requestId`: Optional request id to match (only for `/api/history/tasks` endpoint as it is already specified in the path for `/request/{requestId}/tasks`)
+- `deployId`: Optional deploy id to match
+- `host`: Optional host (slave host name) to match
+- `lastTaskStatus`: Optional [`ExtendedTaskState`](../reference/api-docs/models.md#model-ExtendedTaskState) to match
+- `startedAfter`: Optionally match only tasks started after this time (13 digit unix timestamp)
+- `startedBefore`: Optionally match only tasks started before this time (13 digit unix timestamp)
+- `orderDirection`: Sort direction (by `startedAt`), can be ASC or DESC, defaults to DESC (newest tasks first)
+- `count`: Maximum number of items to return
+- `page`: Page of items to view (e.g. page 1 is the first `count` items, page 2 is the next `count` items)
```
Calvinp: Count and page are optional too. Page defaults to 1, and I believe count defaults to infinite but I don't remember exactly.

ssalinas: ah, yeah, good to include defaults, I'll add them in

Docs/features/task-search.md
```diff
@@ -0,0 +1,23 @@
+#### Task Search
+
+As of 0.4.12, Singularity has better support for searching historical tasks. A global task search endpoint was added:
+
+`/api/history/tasks` -> Retrieve the history sorted by startedAt for all inactive tasks.
+
+... and the above endpoint as well `/api/history/request/{requestId}/tasks` now take additonal query parameters:
```
Calvinp: This line is kind of confusing to me. Maybe 'The above endpoint as well as `/api/history/request/{requestId}/tasks` now take additonal query parameters'

Docs/features/task-search.md
Calvinp: The defaults would be nice to have in api.md too, though it looks like space is limited there. (I can't add line notes to api.md because the diff is so big Github won't show it to me)

ssalinas: Yeah, we need to do some fooling with swagger at some point, that will probably end up being a separate PR some time

Calvinp:
:clap:

PR #841, opened 2016-01-12
Request type zk migration

tpetr:
- bring back deprecated `daemon` field
- add ZK migration to ensure that all requests have `requestType` set
- add null check to `SingularityValidator` for `requestType`

/cc @wsorenson

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/transcoders/JsonTranscoder.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/transcoders/JsonTranscoder.java
@@ -14,7 +14,7 @@ public class JsonTranscoder<T> implements Transcoder<T> {
   private final ObjectMapper objectMapper;
   private final Class<T> clazz;
 
-  JsonTranscoder(final ObjectMapper objectMapper, final Class<T> clazz) {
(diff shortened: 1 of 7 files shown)


wsorenson:
In order for this migration to work in the future, don't you need to drop daemon and reconstruct the object using raw json?

We should add some simple non null checks to request constructor.

wsorenson:
Or copy / paste the actual former request object into the mgiration class itself.

tpetr:
yep you're right, i'll copy in the old object into the migration, and add null checks to the constructor

tpetr:
^ still want to do a little more testing, but this should be more like it

SingularityService/src/main/java/com/hubspot/singularity/data/transcoders/JsonTranscoder.java
```diff
@@ -14,7 +14,7 @@
   private final ObjectMapper objectMapper;
   private final Class<T> clazz;
 
-  JsonTranscoder(final ObjectMapper objectMapper, final Class<T> clazz) {
+  public JsonTranscoder(final ObjectMapper objectMapper, final Class<T> clazz) {
```
tpetr: wasn't sure if this was OK -- i assume it is, but wasn't sure if it was package private on purpose...

PR #842, opened 2016-01-14
Task is cleaning default msg fix

Calvinp:
Pull request to solve an issue where the message field for a bounced task would be blank.

Added a default message which will appear when a task is being cleaned up after a bounce

### DIFF
--- a/SingularityUI/app/templates/taskDetail/taskOverview.hbs
+++ b/SingularityUI/app/templates/taskDetail/taskOverview.hbs
@@ -2,7 +2,10 @@
     {{! If there's a TaskCleanup object associated with this task, display its info }}
     {{#if data.cleanup}}
         <div class="alert alert-warning" role="alert">
-            <b>Task is cleaning:</b> {{data.cleanup.message}}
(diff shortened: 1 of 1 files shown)


ssalinas:
Diff will be better once expiring_Actions is merged in, quick link to diff for this branch
https://github.com/HubSpot/Singularity/compare/expiring_actions...task-is-cleaning-default-msg-fix?expand=1

tpetr:
Thanks for the PR, Calvin!

A few notes:
- It looks like some old `expiring_actions` commits bled into this branch. Now that the branch has been merged to master, could you rebase?
- I think the original issue was unclear, sorry. When the new `message` field was added, the `Task is cleaning` banner was updated to only display the message value (which wasn't ideal, and is causing the blank "Task is cleaning:" banner you saw). What we want to do instead is for the banner to display `cleanup.cleanupType`, and the message in `( )`'s only if there's one set. (examples: `Task is cleaining: BOUNCING` and `Task is cleaning: BOUNCING (my custom message here)`)
- With the new display mentioned above, it's not necessary to set a default message for bounces anymore.

PR #849, opened 2016-01-19
Include message with emails

Calvinp:
Now that Singularity has the capability to accept optional user messages for tasks, these messages can be included in any emails sent because of user actions with messages. This commit includes that ability.
It also includes the ability to set a variable, LOG_EMAILS_FOR_DEBUG - If this variable is set to true, all emails sent will be logged in the debug logs. This allows local testing of emails without an SMTP server.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
@@ -73,6 +73,10 @@ public class SingularityMailer implements Managed {
 
   private static final Pattern TASK_STATUS_BY_PATTERN = Pattern.compile("(\\w+) by \\w+");
 
+  /// Set this to true to log emails being sent. This allows testing locally without setting up an SMTP server
+  /// To find the email html, grep the logs: "grep TheMail -A 10" (change 10 depending on how many lines you need)
+  private static final Boolean LOG_EMAILS_FOR_DEBUG = false;
+
   @Inject
   public SingularityMailer(
       SingularitySmtpSender smtpSender,
@@ -287,6 +291,10 @@ public class SingularityMailer implements Managed {
 
     final String body = Jade4J.render(taskTemplate, templateProperties);
 
+    if (LOG_EMAILS_FOR_DEBUG) {
+      LOG.debug("TheMail: " + body);
+    }
+
     final Optional<String> user = task.isPresent() ? task.get().getTaskRequest().getPendingTask().getUser() : Optional.<String> absent();
 
     queueMail(emailDestination, request, emailType, user, subject, body);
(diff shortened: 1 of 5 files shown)


SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
```diff
@@ -287,6 +291,10 @@ private void prepareTaskMail(Optional<SingularityTask> task, SingularityTaskId t
 
     final String body = Jade4J.render(taskTemplate, templateProperties);
 
+    if (LOG_EMAILS_FOR_DEBUG) {
+      LOG.debug("TheMail: " + body);
```
tpetr: I'd suggest logging this at the `LOG.trace()` level instead, so that people can more easily control whether or not the email should be logged

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularityMailer.java
```diff
@@ -597,6 +625,10 @@ private void queueMail(final Collection<SingularityEmailDestination> destination
       }
     }
 
+    if (LOG_EMAILS_FOR_DEBUG) {
+      LOG.debug("TheMail: " + body);
+    }
+
```
tpetr: We should minimize the number of times the email body is logged -- i think just having this one here will suffice

Calvinp:
These two issues have been fixed and the branch pushed again.

Calvinp:
Whoops, I lied before, forgot about making it trace level. Now they're both fixed.

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularitySmtpSender.java
```diff
@@ -62,6 +68,10 @@ public void stop() {
   }
 
   public void queueMail(final List<String> toList, final List<String> ccList, final String subject, final String body) {
+    if (LOG_EMAILS_FOR_DEBUG) {
+      LOG.trace("TheMail: " + body);
```
ssalinas: We can do without the `LOG_EMAILS_FOR_DEBUG`. We can control the logging level for each class in the configuration for Singularity, so just putting it at trace should be enough.

Also for logging, you can so something more like this rather that using "string" + "string"
```java
LOG.trace("Sending email: {}, to: {}, cc: {}", body, toList, ccList)
```

jhaber: For readability I've found it to be easier to put the recipients first and the body last so you're not scrolling through a wall of HTML looking for the recipient at the end

Calvinp:
Fixed

SingularityService/src/main/java/com/hubspot/singularity/smtp/SingularitySmtpSender.java
```diff
@@ -30,6 +30,8 @@
 
 import io.dropwizard.lifecycle.Managed;
 
+
+
```
ssalinas: super nit-picky, but can we get rid of these extra new lines here and down by the LOG.trace

Calvinp:
Not sure how those got there in the first place, but they're gone now.

PR #852, opened 2016-01-20
Page Now Automatically Refreshes Even After Invalid Duration Entered

Calvinp:
This could potentially create a merge conflict with #851 - if so, resolve by adding the callback argument to any @_validateDuration calls.


ssalinas:
looks good, thanks @Calvinp

PR #854, opened 2016-01-21
Add methods to client for unpause and scale requests

bacchus123:


ssalinas:
Thanks for the PR @bacchus123 ! LGTM

SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
```diff
@@ -323,9 +327,25 @@ private void addCredentials(HttpRequest.Builder requestBuilder) {
     return Optional.absent();
   }
 
+  private <T> Optional<T> put(String uri, String type, Optional<?> body, Optional<Class<T>> clazz) {
```
tpetr: it looks like this method is unused, and is causing findbugs to barf. could you remove this method to make the build pass?

tpetr:
Pulled out of `hs_staging` due to failing build via https://github.com/HubSpot/Singularity/commit/ba6bf5f85e8c3fd4c0e689f7449c5bfe34cc4146

bacchus123:
this builds locally with 
mvn clean install fine so I'm not sure what I need to fix??

tpetr:
build with `mvn clean verify` to ensure that checkstyle, findbug, etc. all run

bacchus123:
I ran `mvn clean verify` and all projects completed successfully

ssalinas:
Thanks for fixing @bacchus123 , I put this back into staging

PR #855, opened 2016-01-21
Update ui.md

Calvinp:
Fix the Javascript console command for changing the API Root Override.
localStorage.set() is not a function, it's localStorage.setItem()


ssalinas:
Thanks! LGTM

PR #856, opened 2016-01-22
Upgrade Horizon

jhaber:
The new version of Horizon makes the async client use daemon threads which helps to ensure the JVM shuts down gracefully

@tpetr @ssalinas @wsorenson

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ git:
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


jhaber:
I upgraded to the new basepom as part of this

ssalinas:
trying this in hs_staging now

PR #860, opened 2016-01-26
Fix the sorting in the wrong order bug.

Calvinp:
When more than 10 files were present in a task, the sort order became very wrong in Chrome. Manually overriding the sort order would cause files to jump around.
This fixes the wrong initial sort order. Unfortunately files will still jump around, but that can be avoided by not manually sorting the table.

### DIFF
--- a/SingularityUI/app/collections/TaskFiles.coffee
+++ b/SingularityUI/app/collections/TaskFiles.coffee
@@ -58,7 +58,7 @@ class TaskFiles extends Collection
             return -1
         else if not a.get('isDirectory') and b.get('isDirectory')
             return 1
-        return a.get('path') - b.get('path')
+        return a.get('name') > b.get('name')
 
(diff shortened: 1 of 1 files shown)


wsorenson:
Why do files still jump around?

Calvinp:
The page refreshes fairly often (I think whenever it becomes in focus), so if you manually override the sort order it will refresh back to the default sort order.

wsorenson:
- Is there a way to pull the elements without refreshing the page and changing client interactions like sort order?

- Is there a way to persist changes like sort order after refresh?

Curious if this is related to standard backbone issues and thus there should be a well-known workaround.  I'd check with other FE resources.

Calvinp:
I am/will look into it. 
Either way, this PR can be advanced separately from that. It will fix the pain in a fair portion of cases on its own.

PR #861, opened 2016-01-26
make sure request history always in correct order

ssalinas:


ssalinas:
added a small tweak to the ui so that we won't show the deleted message unless we are really deleted

PR #865, opened 2016-01-26
better handling of Docker timeouts

ssalinas:
@wsorenson this was my addition to #864

`DockerUtils` has a single threaded `ExecutorService` (we only need one call at a time right now anyways) that will execute each with a timeout. So, anything docker is not executed on the main thread that needs to keep going in order for the Executor to communicate with mesos.

Also adds a default thread pool of 5 for the docker client. The initial client was meant for managing large numbers of containers concurrently, thus the limit of 100.

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
@@ -65,7 +65,7 @@ public class SingularityExecutorTaskProcessBuilder implements Callable<ProcessBu
       try {
         dockerUtils.pull(task.getTaskInfo().getContainer().getDocker().getImage());
       } catch (DockerException e) {
-        throw new ProcessFailedException(String.format("Could not pull docker image due to error: %s", e));
+        throw new ProcessFailedException("Could not pull docker image", e);
       }
     }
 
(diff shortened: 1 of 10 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
```diff
@@ -63,9 +63,9 @@ public ProcessBuilder call() throws Exception {
     if (task.getTaskInfo().hasContainer() && task.getTaskInfo().getContainer().hasDocker()) {
       executorUtils.sendStatusUpdate(task.getDriver(), task.getTaskInfo(), TaskState.TASK_STARTING, String.format("Pulling image... (executor pid: %s)", executorPid), task.getLog());
       try {
-        dockerClient.pull(task.getTaskInfo().getContainer().getDocker().getImage());
-      } catch (UncheckedTimeoutException te) {
-        throw new ProcessFailedException(String.format("Timed out trying to reach docker daemon after %s seconds", configuration.getDockerClientTimeLimitSeconds()));
+        dockerUtils.pull(task.getTaskInfo().getContainer().getDocker().getImage());
+      } catch (DockerException e) {
+        throw new ProcessFailedException(String.format("Could not pull docker image due to error: %s", e));
```
wsorenson: Don't add the exception to an error message, include the exception as a cause argument.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/utils/DockerUtils.java
```diff
@@ -0,0 +1,125 @@
...
+    Callable<ContainerInfo> callable = new Callable<ContainerInfo>() {
+      @Override public ContainerInfo call() throws Exception {
+        return dockerClient.inspectContainer(containerName);
+      }
+    };
+
+    try {
+      return callWithTimeout(callable);
+    } catch (Exception e) {
+      LOG.error("Caught exception while getting container status", e);
```
wsorenson: I would only log all of these if you don't control the code that is going to handle the thrown exception - meaning, you need to ensure it gets logged here because you don't know if it will be logged when caught.

PR #866, opened 2016-01-28
add a default bounce expiration

ssalinas:
Don't allow a bounce that will go on forever. Defaults to 60 minutes and shows in the ui as:
screen shot 2016-01-28 at 9 14 55 am

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
@@ -8,7 +8,6 @@ import static com.hubspot.singularity.WebExceptions.checkNotNullBadRequest;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.TimeUnit;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
@@ -152,7 +151,7 @@ public class RequestResource extends AbstractRequestResource {
   @ApiOperation(value="Bounce a specific Singularity request. A bounce launches replacement task(s), and then kills the original task(s) if the replacement(s) are healthy.",
   response=SingularityRequestParent.class)
   public SingularityRequestParent bounce(@ApiParam("The request ID to bounce") @PathParam("requestId") String requestId,
-      @ApiParam("Bounce request options") Optional<SingularityBounceRequest> maybeBounceRequest) {
+      @ApiParam("Bounce request options") Optional<SingularityBounceRequest> bounceRequest) {
     SingularityRequestWithState requestWithState = fetchRequestWithState(requestId);
 
     authorizationHelper.checkForAuthorization(requestWithState.getRequest(), user, SingularityAuthorizationScope.WRITE);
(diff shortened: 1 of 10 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -171,34 +190,18 @@ public SingularityRequestParent bounce(@ApiParam("The request ID to bounce") @Pa
...
-    final Optional<Boolean> skipHealthchecks = bounceRequest.isPresent() ? bounceRequest.get().getSkipHealthchecks() : Optional.<Boolean> absent();
-
-    Optional<String> message = Optional.absent();
-    Optional<String> actionId = Optional.absent();
-
-    if (bounceRequest.isPresent()) {
-      actionId = bounceRequest.get().getActionId();
-      message = bounceRequest.get().getMessage();
-
-      if (bounceRequest.get().getDurationMillis().isPresent() && !actionId.isPresent()) {
```
wsorenson: This logic was changed

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -162,7 +163,25 @@ public SingularityRequestParent bounce(@ApiParam("The request ID to bounce") @Pa
...
+        maybeBounceRequest.get().getIncremental(),
+        maybeBounceRequest.get().getSkipHealthchecks(),
+        maybeBounceRequest.get().getDurationMillis().or(Optional.of(TimeUnit.MINUTES.toMillis(configuration.getDefaultBounceExpirationMinutes()))),
+        maybeBounceRequest.get().getActionId().or(Optional.of(UUID.randomUUID().toString())),
+        maybeBounceRequest.get().getMessage());
+    } else {
+      bounceRequest = new SingularityBounceRequest(
+        Optional.<Boolean>absent(),
+        Optional.<Boolean>absent(),
+        Optional.of(TimeUnit.MINUTES.toMillis(configuration.getDefaultBounceExpirationMinutes())),
```
wsorenson: Instead of storing this in the object, why not check it later? Recreating the object like this is gross.

wsorenson: Also, as a general rule, late-binding of configuration properties makes testing easier.

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
ssalinas: Thanks, I'll check against staging to see what changed

wsorenson: Ah, now that I look again - are we now saying it's impossible to have unlimited bounce?

If so, what will you do if someone enters a negative value for this configuration? Need to have validation on it or allow people to set it to 0/negative to indicate that they don't want to use this feature.

In which case, we'd only set actionId if there was a duration set.

ssalinas: Do you think we should still create an expiring object either way? One thing I realized in writing this is that if we don't create the expiring object, there isn't actually a way to cancel the bounce (without doing a janky pause/unpause)

wsorenson: Good point, maybe just setup the config to validate that this must be non-zero, positive, then it's not an issue. I think you can do that with annotations. Someone could set it really high.

ssalinas:
Ok, reworked this a bit. Not reconstructing the object anymore, just getting the `durationMillis` slightly differently for bounces in `SingularityExpiringUserActionPoller` so we can use the default and added a `@Min` for the config value. Tweaked the UI to show default value as well if there isn't a duration present for bounce

PR #868, opened 2016-02-01
bump to brunch 2

ssalinas:
Should also fix the handlebars-brunch travis build issue we were seeing


ssalinas:
brunch piece of the build is working on travis, although it seems we have another flaky test in there that needs fixing

PR #870, opened 2016-02-01
allow s3 log retrieval even if tasks have been deleted

wsorenson:


SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
```diff
@@ -98,23 +98,32 @@ public S3LogResource(RequestManager requestManager, HistoryManager historyManage
...
+
+    if (history.isPresent()) {
+      SimplifiedTaskState taskState = SingularityTaskHistoryUpdate.getCurrentState(history.get().getTaskUpdates());
+
+      if (taskState == SimplifiedTaskState.DONE) {
+        end = Iterables.getLast(history.get().getTaskUpdates()).getTimestamp();
+      }
+    } else if (configuration.isPresent()) {
+      end = start + configuration.get().getMissingTaskDefaultS3SearchPeriodMillis();
+    }
```
jhaber: So if the history is missing and there is no S3Configuration it searches from task start until now? Should it return an error instead?

wsorenson: That would probably be more correct. As is, it's impossible due to the checkS3 call made before this method is called.

PR #871, opened 2016-02-02
add support for override defaults file

tpetr:
This PR adds support for pulling in an additional Singularity configuration file which overrides values from the default one. This is useful if you need to lock down S3 / database / LDAP credentials but would like the other settings to have looser read/write permissions.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/MergingSourceProvider.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/MergingSourceProvider.java
@@ -16,24 +16,24 @@ import io.dropwizard.configuration.ConfigurationSourceProvider;
 
 public class MergingSourceProvider implements ConfigurationSourceProvider {
     private final ConfigurationSourceProvider delegate;
-    private final String overridePath;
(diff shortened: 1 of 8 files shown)


SingularityService/src/main/java/com/hubspot/singularity/config/MergingSourceProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+        while (newFieldNames.hasNext()) {
+            String newFieldName = newFieldNames.next();
+            JsonNode oldVal = to.get(newFieldName);
+            JsonNode newVal = from.get(newFieldName);
+
+            if (oldVal == null || oldVal.isNull()) {
+                to.put(newFieldName, newVal);
+            } else if (oldVal.isArray() && newVal.isArray()) {
+                ((ArrayNode) oldVal).addAll((ArrayNode) newVal);
```
jhaber: For this use-case do you think the behavior should be to append or to overwrite?

tpetr: good eye -- probably overwrite

PR #879, opened 2016-02-04
Add ability to post task metadata to a task

wsorenson:


wsorenson:
Will need to add this to:

- [ ] UI
- [ ] Email messages

Open question of whether we should provide a en enum for type (less extensible) but more explicit for the UI.

Still need to delay task email & persistence so that post-task metadata can be sent.

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -322,6 +323,27 @@ public SingularityTaskCleanup killTask(@PathParam("taskId") String taskId, Optio
...
+  })
+  @Consumes({ MediaType.APPLICATION_JSON })
+  public void postTaskMetadata(@PathParam("taskId") String taskId, final SingularityTaskMetadata taskMetadata) {
+    SingularityTaskId taskIdObj = getTaskIdFromStr(taskId);
+
+    authorizationHelper.checkForAuthorizationByTaskId(taskId, user, SingularityAuthorizationScope.WRITE);
+
+    WebExceptions.checkBadRequest(taskMetadata.getTaskId().equals(taskIdObj), "Task metadata taskId %s didn't match API path %s", taskMetadata.getTaskId(), taskId);
+
+    SingularityCreateResult result = taskManager.saveTaskMetadata(taskMetadata);
```
ssalinas: If the task has already been persisted to history and removed from zk, we won't be able to update the task's metadata. Do we want to be able to?

wsorenson: In the ideal world, sure, but I don't think it's realistic for now, which is why we're delaying task persistence for this case.

ssalinas: Sounds good, just realized you mentioned that at the end of the other comment

Calvinp:
The UI and email portions of this are complete and merged in here. From my perspective, this is good to push along.

PR #880, opened 2016-02-04
Healthcheck message fix

Calvinp:
This fixes the two healthcheck messages.

On the yellow notification bar at the top of the task view, there will be a message stating how many times the healthchecks were retried if they were tried more times than allowed. If not, instead it will state the number of seconds the checks were configured to run before timeout.

In the healthchecks subview, now the retry limit is explicitly stated next to the timeout limit.


SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
```diff
@@ -27,6 +29,9 @@ class taskHealthcheckNotificationSubview extends View
         lastHealthcheckFailed: @model.get('healthcheckResults')?.length > 0 and @model.get('healthcheckResults')[0].statusCode isnt 200
         synced:           @model.synced
         config:           config
+        tooManyRetries: @model.get('healthcheckResults').length >= maxRetries
```
ssalinas: Small thing, number of retries doesn't include the first try, meaning we have exhausted our retries if there are `maxRetries + 1` failed health checks present. Currently, this may unintentionally come up with exhausted retries when it is actually a timeout.

Related, you are using numberFailed as the number of failed retries in the message, but defining it as the number of failed total. Either way is fine, just needs to be consistent in messaging

Calvinp: Good point; This is fixed with the new commit.

ssalinas:
:+1:

ssalinas:
Thanks for all the changes and small tweaks to this @Calvinp , LGTM

PR #881, opened 2016-02-05
Logrotate and S3 upload tweaks

tpetr:
- support file extensions for extra logrotated files
- support uploading files to a separate S3 bucket

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
@@ -1,0 +1,36 @@
+package com.hubspot.singularity.executor.config;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Optional;
+
+public class SingularityExecutorLogrotateAdditionalFile {
+    private final String filename;
+    private final Optional<String> extension;
+    private final Optional<String> dateformat;
+
+    public static SingularityExecutorLogrotateAdditionalFile fromString(String value) {
+        return new SingularityExecutorLogrotateAdditionalFile(value, Optional.<String>absent(), Optional.<String>absent());
+    }
+
+    @JsonCreator
+    public SingularityExecutorLogrotateAdditionalFile(@JsonProperty("filename") String filename,
+        @JsonProperty("extension") Optional<String> extension,
+        @JsonProperty("dateformat") Optional<String> dateformat) {
+        this.filename = filename;
+        this.extension = extension;
+        this.dateformat = dateformat;
+    }
+
+    public String getFilename() {
+        return filename;
+    }
+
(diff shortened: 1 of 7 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Optional;
+
+public class SingularityExecutorLogrotateAdditionalFile {
+    private final String filename;
+    private final Optional<String> extension;
+    private final Optional<String> dateformat;
+
+    public static SingularityExecutorLogrotateAdditionalFile fromString(String value) {
```
ssalinas: Need an `@JsonCreator` for this one?

tpetr: Dang -- I forgot the annotation but Jackson is magically making it work (based on the method name & signature i assume). I'll add the annotation.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskLogManager.java
```diff
@@ -58,7 +56,15 @@ private boolean writeS3MetadataFileForRotatedFiles(boolean finished) {
     final Path serviceLogParent = serviceLogOutPath.getParent();
     final Path logrotateDirectory = serviceLogParent.resolve(configuration.getLogrotateToDirectory());
 
-    return writeS3MetadataFile("default", logrotateDirectory, getS3GlobForLogRotatedFiles(), finished);
+    boolean result = writeS3MetadataFile("default", logrotateDirectory, String.format("%s*.gz*", taskDefinition.getServiceLogOutPath().getFileName()), Optional.<String>absent(), Optional.<String>absent(), finished);
+
+    int index = 1;
+    for (SingularityExecutorS3UploaderAdditionalFile additionalFile : configuration.getS3UploaderAdditionalFiles()) {
+      result = result && writeS3MetadataFile(additionalFile.getS3UploaderFilenameHint().or(String.format("extra%d", index)), logrotateDirectory, String.format("%s*.gz*", additionalFile.getFilename()), additionalFile.getS3UploaderBucket(), additionalFile.getS3UploaderKeyPattern(), finished);
```
ssalinas: I think this is forcing that all files to upload are gzipped, do we want to enforce that if it's possible to use this an a folder for uploading arbitrary files?

tpetr: Yep (and that's how it was operating before this PR: https://github.com/HubSpot/Singularity/pull/881/files#diff-e8314245f66e3ac2fc51f17320d6f1f6L197). Arbitrary file upload via a "drop" folder will come in a separate PR, if we attempt it

PR #885, opened 2016-02-08
Allow destination targets to be specified

wsorenson:
FYI @jhaber @tpetr 

We should probably stop relying on embedded artifacts "defaulting" to the app directory - that feels off to me given this capability.

Part 2 of this will allow batching, but I don't think it's necessary for MVP.


tpetr:
:+1:

jhaber:
cool lgtm

PR #887, opened 2016-02-09
allow deploy to specify port index for healthchecks and LB api

### LINKED ISSUE #886: Support load balancing and healthchecks on other ports
Singularity currently uses the first allocated port for healthchecks and load balancing, but sometimes greater flexibility is needed.

ssalinas:
On the deploy you can specify `loadBalancerPortIndex` or `healthcheckPortIndex` (e.g. `0` for first allocated port, `1` for second allocated port). Defaults to 0 if not specified

/fixes #886

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
@@ -46,6 +46,7 @@ public class SingularityDeploy {
   private final Optional<String> healthcheckUri;
   private final Optional<Long> healthcheckIntervalSeconds;
   private final Optional<Long> healthcheckTimeoutSeconds;
+  private final Optional<Integer> healthcheckPortIndex;
   private final Optional<Boolean> skipHealthchecksOnDeploy;
   private final Optional<HealthcheckProtocol> healthcheckProtocol;
 
@@ -58,6 +59,7 @@ public class SingularityDeploy {
 
   private final Optional<String> serviceBasePath;
   private final Optional<Set<String>> loadBalancerGroups;
+  private final Optional<Integer> loadBalancerPortIndex;
   private final Optional<Map<String, Object>> loadBalancerOptions;
 
   public static SingularityDeployBuilder newBuilder(String requestId, String id) {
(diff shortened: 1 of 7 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
```diff
@@ -281,6 +289,11 @@ public String getId() {
     return healthcheckTimeoutSeconds;
   }
 
+  @ApiModelProperty(required=false, value="Perform healthcheck on this dynamically allocated port (e.g. 0 for first port)")
```
tpetr: should probably indicate that it defaults to the first port

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityHealthchecker.java
```diff
@@ -167,9 +169,9 @@ public void reEnqueueOrAbort(SingularityTask task) {
 
     final String hostname = task.getOffer().getHostname();
 
-    Optional<Long> firstPort = task.getFirstPort();
+    Optional<Long> healthecheckPort = task.getPortByIndex(task.getTaskRequest().getDeploy().getHealthcheckPortIndex().or(0));
```
tpetr: typo

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTask.java
```diff
@@ -45,16 +47,13 @@ public TaskInfo getMesosTask() {
...
-  public Optional<Long> getFirstPort() {
-    for (Resource resource : mesosTask.getResourcesList()) {
-      if (resource.getName().equals(MesosUtils.PORTS)) {
-        for (Range range : resource.getRanges().getRangeList()) {
-          return Optional.of(range.getBegin());
-        }
-      }
+  public Optional<Long> getPortByIndex(int index) {
+    List<Long> ports = MesosUtils.getAllPorts(mesosTask.getResourcesList());
+    if (index >= ports.size()) {
```
tpetr: should check for `index < 0` too

tpetr:
Can you also add checks for these new fields to `SingularityValidator.checkDeploy()`?

tpetr:
Worth adding and/or updating tests to verify this works

tpetr:
(LGTM once all this is addressed)

ssalinas:
Added validator pieces, just need to add tests

tpetr:
LGTM, thanks

Issue #874, opened 2016-02-03
Tasks can fail repeatedly without entering cooldown if they fail too early in the launching process


stevenschlansker:
We run the standard Docker executor.  Unfortunately an administrator deleted one of the Docker images of an active task.
This caused Singularity to get in an endless loop restarting the task.
Apparently, if the active task fails abnormally before it begins, this does not count against the cooldown limit.  So it was left running overnight.

This filled up our ZooKepeer cluster with tens of thousands of tasks, which caused the ZK cluster to go south.

* Any task failure should count towards cooldown
* There should be some hard limit of how many tasks are kept in history, even if they haven't expired.

```
INFO  [2016-02-02 07:07:20,963] com.hubspot.singularity.mesos.SingularityMesosScheduler: Launching task prod-umami-config-server-teamcity.2015.08.26T01.30.04-1454396840963-2-mesos_slave6_prod_sc.otsql.opentable.com-FIXME slot on slave 20151217-222149-3943699466-5050-19261-S4 (mesos-slave6-prod-sc.otsql.opentable.com)
INFO  [2016-02-02 07:07:20,978] com.hubspot.singularity.mesos.SingularityMesosScheduler: 1 tasks ([prod-umami-config-server-teamcity.2015.08.26T01.30.04-1454396840963-2-mesos_slave6_prod_sc.otsql.opentable.com-FIXME]) launched with status DRIVER_RUNNING
DEBUG [2016-02-02 07:07:21,188] com.hubspot.singularity.mesos.SingularityMesosScheduler: Task prod-umami-config-server-teamcity.2015.08.26T01.30.04-1454396840963-2-mesos_slave6_prod_sc.otsql.opentable.com-FIXME is now TASK_FAILED (Failed to launch container: Failed to 'docker pull docker-prod-sc.otenv.com/umami-config-server:23': exit status = exited with status 1 stderr = Error: Status 400 trying to pull repository umami-config-server: "{\n  \"errors\" : [ {\n    \"status\" : 400,\n    \"message\" : \"Unsupported docker v1 repository request for 'docker-v2'\"\n  } ]\n}"
) at 1454396841177 
[...]

PR #889, opened 2016-02-09
add support for max numbers of objects in ZK when no database is configured

tpetr:
re: #874

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityDeployHistoryPersister.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityDeployHistoryPersister.java
@@ -1,8 +1,6 @@
 package com.hubspot.singularity.data.history;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityDeployHistoryPersister.java
```diff
@@ -56,16 +59,22 @@ public void runActionOnPoll() {
...
-        continue;
       }
+    }
 
-      if (moveToHistoryOrCheckForPurge(deployHistory.get())) {
-        numTransferred++;
-      }
+    for (Collection<SingularityDeployHistory> deployHistoryForRequest : deployHistoryByRequestId.asMap().values()) {
+      int i=0;
+      for (SingularityDeployHistory deployHistory : deployHistoryForRequest) {
```
ssalinas: Do we need to sort the `Collection<SingularityDeployHistory>` first here? I saw where we were sorting for tasks and for requests, but did not see any sort for deploy history.

tpetr: The values in a `TreeMultimap<>` are sorted by default

tpetr:
FYI @stevenschlansker

PR #890, opened 2016-02-09
Task Search UI

Calvinp:
This Implements the UI for the new global and request task search APIs.
It is written in React, which has only been used for one other Singularity component (the aggregate log tailer).
It allows viewing up to 50 tasks per page, clearing search parameters, modifying search order, paging through search options, and returning to request.
It also includes some components, in a 'common' folder, that may be useful for others using react later on.


SingularityUI/app/collections/TaskSearchResults.coffee
```diff
@@ -0,0 +1,63 @@
...
+    			base += '&'
+    		base += 'count=' + @count
+    		anyParams = true
+    	if @page
+    		if anyParams
+    			base += '&'
+    		base += 'page=' + @page
+    		anyParams = true
+    	return '' if not anyParams
+    	return base
```
ssalinas: We can build the params a bit cleaner here. Probably easier to build up an array of params (i.e. if @attribute then add to array), then return paramsArray.join('&')

SingularityUI/app/collections/TaskSearchResults.coffee
```diff
@@ -0,0 +1,63 @@
+Collection = require './collection'
+
+TaskSearchResult = require '../models/TaskSearchResult'
+
+class TaskSearchResults extends Collection
+
+    model: TaskSearchResult
+
+    params: ->
```
ssalinas: because a number of our backbone collections and other classes have an @params of some sort, it might be easier to rename this method to something like `queryParams` or `getQueryParams`

SingularityUI/app/collections/TaskSearchResults.coffee
```diff
@@ -0,0 +1,63 @@
...
+    		anyParams = true
+    	return '' if not anyParams
+    	return base
+
+    url: -> 
+    	if @requestId
+    		base = "#{ config.apiRoot }/history/request/#{ @requestId }/tasks"
+    	else
+    		base = "#{ config.apiRoot }/history/tasks"
+    	return base + @params()
```
ssalinas: You can do this with string interpolation rather than adding strings, e.g.

```coffee
if @requestId
	"#{ config.apiRoot }/history/request/#{ @requestId }/tasks?#{ @queryParams() }"
else
 	"#{ config.apiRoot }/history/tasks?#{ @queryParams() }"
```

SingularityUI/app/collections/TaskSearchResults.coffee
```diff
@@ -0,0 +1,63 @@
...
+
+class TaskSearchResults extends Collection
+
+    model: TaskSearchResult
+
+    params: ->
+    	anyParams = false
+    	base = '?'
+    	if @deployId
+    		base += 'deployId=' + @deployId
```
ssalinas: Let's make sure these are all spaces not tabs, it's burned us earlier when testing this out. Comment applies for all files

SingularityUI/app/collections/TaskSearchResults.coffee
tpetr: an even better solution would be to build an Object of all the query params to set, and then use `$.param()` to convert the object into a legit query string

SingularityUI/app/components/common/Link.cjsx
```diff
@@ -0,0 +1,13 @@
+Link = React.createClass
+
+    getLink: ->
+        return window.config.appRoot + "/task/" + @props.taskId
```
ssalinas: can just use `config` here, don't need it to be `window.config`

also, another case for the string interpolation here
```coffee
"#{ config.apiRoot }/task/#{ @props.taskId }"
```

Calvinp: Actually, that function isn't even used - it's leftover from when I was doing this a different way, and I guess I forgot to delete it.

SingularityUI/app/collections/TaskSearchResults.coffee
ssalinas: TIL... thanks @tpetr :smiley:

SingularityUI/app/components/taskSearch/TaskSearch.cjsx
```diff
@@ -0,0 +1,198 @@
...
+            startedAfter: @props.initialStartedAfter
+            sortDirection: @props.initialSortDirection or 'ASC'
+            pageNumber: 1
+            count: @props.initialCount or @countDefault
+            showForm: true
+        }
+
+    handleSubmit: (event) ->
+        event.preventDefault()
+        @setState({
```
ssalinas: more of a space-saving/readability thing we can do with coffeescript
```
@setState showForm: false
```
is the same as
```
 @setState({
      showForm: false
})
```

SingularityUI/app/controllers/TaskSearch.coffee
```diff
@@ -0,0 +1,29 @@
+Controller = require './Controller'
+
+TaskSearchView = require '../views/taskSearch'
+TaskSearchResults = require '../collections/TaskSearchResults'
```
ssalinas: Doesn't look like this is used in the controller here

SingularityUI/app/collections/TaskSearchResults.coffee
```diff
@@ -0,0 +1,63 @@
+Collection = require './collection'
```
ssalinas: Looking at our other collections/models, it seems we are doing some double work with this and the `RequestHistoricalTasks` collection. Both are the same url, this one has the extra params, the other has a few more methods. I would argue that we merge the two together (maybe into `RequestHistoricalTasks` for naming). `RequestHistoricalTasks` also already has a `TaskHistoryItem` defined that we can use in place of the `TaskSearchResult` model

SingularityUI/app/components/common/Table.cjsx
```diff
@@ -0,0 +1,47 @@
...
+
+    renderTableRow: (elements) ->
+        data = []
+        key = 0
+        for element in elements
+            data.push(<td key={key}>{element}</td>)
+            key++
+        return data
+
+    # Note: The data in @props.tableRows should be the JSX you would like to render
```
zdhickman: If you prefer, you can instead use the pattern where you pass the name of a specific row component as a prop into this `Table` component, as well as a prop that contains raw data (instead of straight `jsx`). You can call the row component by name:

```
TableRow = @props.rowComponent # e.g. @props.rowComponent = "MyTableRowComponent"

...

<TableRow # is replaced with "MyTableRowComponent"
  prop={@props.prop}
  ...
/>
```

SingularityUI/app/components/taskSearch/QueryParam.cjsx
```diff
@@ -0,0 +1,14 @@
+QueryParam = React.createClass
```
zdhickman: This is a helpful component - have you considered defining an additional component to contain a bunch of `QueryParam`s, instead of throwing them into a JS array/object?

SingularityUI/app/components/common/input/DateEntry.cjsx
```diff
@@ -0,0 +1,46 @@
+FormField = require './FormField'
+
+DateEntry = React.createClass
+
+    componentDidMount: ->
+        id = '#' + @props.id
+        changeFn = @props.updateFn
+        $ -> $(id).datetimepicker({
+                sideBySide: true
+                format: "ddd MMM DD YYYY HH:mm:ss [UTC]ZZ"
```
zdhickman: Drag out the string to be a constant somewhere

SingularityUI/app/components/common/input/DateEntry.cjsx
```diff
@@ -0,0 +1,46 @@
...
+                    placeholder = {@props.title}
+                    type = {@props.inputType}
+                    disabled = {@props.disabled}
+                    size = {@props.size}
+                    value = {@getValue()}
+                    updateFn = {@props.updateFn} 
+                    id = {@props.id}
+                />
+                <span className="input-group-addon">
+                    <span className="glyphicon glyphicon-calendar"></span>
```
zdhickman: You can make an `Icon` or `Glyphicon` component for this and the others

SingularityUI/app/components/common/input/FormField.cjsx
```diff
@@ -0,0 +1,30 @@
...
+    defaultSize: 50
+
+    getSize: ->
+        if @props.size
+            return @props.size
+        else
+            return @defaultSize
+
+    getDisabled: ->
+        if @props.disabled
```
zdhickman: just return `@props.disabled` (or don't create a function at all)

SingularityUI/app/components/taskSearch/DisplayResults.cjsx
```diff
@@ -0,0 +1,242 @@
...
+                </ul>
+            </nav>
+        </div>
+
+
+    renderTasks: ->
+        taskTableColumns = ["Name", "Last State", "Deploy", "Started", "Updated"]
+        taskTableData = []
+        for task in @collection.models
+            taskTableData.push([<Link
```
zdhickman: Another instance where creating a container component would be useful

SingularityUI/app/components/taskSearch/DisplayResults.cjsx
```diff
@@ -0,0 +1,242 @@
...
+                                />, 
+                                <TimeStamp 
+                                    timestamp={task.taskId.startedAt} 
+                                    display='timeStampFromNow'} 
+                                />, 
+                                <TimeStamp 
+                                    timestamp={task.updatedAt} 
+                                    display='timeStampFromNow'} 
+                                />])
+        return <StripedTable
```
zdhickman: Probably should have a wrapper component that generates the `StripedTable` and all of the data in `taskTableData`

SingularityUI/app/components/taskSearch/TaskSearchForm.cjsx
```diff
@@ -0,0 +1,100 @@
...
+    updateCount: (event) ->
+        @props.updateCount(event.target.value)
+
+    render: ->
+        <div className='col-xs-5'>
+            <Header
+                global = @props.global
+                requestId = @props.requestId
+            />
+            <form role="form" onSubmit={@props.handleSubmit}>
```
zdhickman: You have a table component to generate tables for you - have you considered doing the same for forms? Even creating a wrapper component for each form group that contains a label and a `FormField` would be helpful

SingularityUI/app/components/common/input/FormField.cjsx
Calvinp: This function used to do something different, then I changed around some functionality and forgot about it. That is a very good point.

SingularityUI/app/components/common/input/DateEntry.cjsx
Calvinp: This actually already existed as a constant, I just forgot about it. Pulling it out.

SingularityUI/app/collections/TaskSearchResults.coffee
```diff
@@ -0,0 +1,21 @@
...
+
+TaskHistoryItem = require '../models/TaskHistoryItem'
+
+class TaskSearchResults extends Collection
+
+    model: TaskHistoryItem
+
+    getQueryParams: ->
+        return $.param(_.pick(@params, _.identity))
```
TrevorBurnham: Isn't `_.pick(@params, _.identity)` just a copy of `@params`?

ssalinas: We wanted to exclude null items, things like deployId=&requestId=somevalue were messing with the optional query params on the backend, making them optionals of empty strings instead of optional absents

SingularityUI/app/components/common/Table.cjsx
```diff
@@ -0,0 +1,47 @@
...
+
+Table = React.createClass
+
+    renderTableHeader: ->
+        header = []
+        key = 0
+        for columnName in @props.columnNames
+            header.push(<th key={key}>{columnName}</th>)
+            key++
+        return header
```
TrevorBurnham: You can express these `for` loops in a more JSX-y way:

```coffee
renderTableHeader: ->
  {@props.columnNames.map (columnName, i) =>
    <th key={i}>{columnName}</th>
  }
```

SingularityUI/app/components/common/Table.cjsx
```diff
@@ -0,0 +1,47 @@
...
+        key = 0
+        for tableRow in @props.tableRows
+            tableRows.push(<tr key={key}>{@renderTableRow tableRow}</tr>)
+            key++
+        return tableRows
+
+
+    # Note: Use @props.tableClassOpts to declare things like striped or bordered
+    getClassName: ->
+        return "table " + @props.tableClassOpts
```
TrevorBurnham: There's a handy `classnames` function in [react-utils](https://git.hubteam.com/HubSpot/react-utils). I highly recommend using it instead of string concatenation, and using a `className` prop instead of `tableClassOpts`. So this would become:

```coffee
getClassName: ->
  classNames('table', @props.className)
```

SingularityUI/app/components/taskSearch/DisplayResults.cjsx
```diff
@@ -0,0 +1,242 @@
...
+                            <label htmlFor="pageNumber" className="sr-only">Jump To Page:</label>
+                            <FormField 
+                                value = @state.pageNumberEntered 
+                                inputType = 'number'
+                                id = 'pageNumber'
+                                title = "Jump to Page"
+                                updateFn = @updatePageNumber />
+                        </div>
+                        <button type="submit" className="btn btn-default">Jump!</button>
+                        &nbsp;&nbsp;
```
TrevorBurnham: Usually it's preferable to add margin or padding with CSS rather than adding non-breaking spaces in the markup, since it makes restyling easier.

SingularityUI/app/components/common/Table.cjsx
Calvinp: Ooh it's really cool that that works. I didn't even think of doing it.

SingularityUI/app/components/taskSearch/QueryParam.cjsx
```diff
@@ -0,0 +1,14 @@
...
+    getClassName: ->
+        name="list-group-item"
+        if @props.cantClear
+            name += " disabled"
+        return name
+
+    render: ->
+        return  <li className={@getClassName()}>
+                        <b>{@props.paramName}:</b> {@props.paramValue} 
+                        {<span className="remove-query-param glyphicon glyphicon-remove pull-right" onClick={@props.onClick} /> unless @props.cantClear}
```
TrevorBurnham: As a rule, anything clickable should should a `<button>` tag. A clickable `<span>` can't be selected via the keyboard, and doesn't indicate its clickability when you hover over it with the mouse. If SingularityUI doesn't already have an "icon button" style class, I would add one.

SingularityUI/app/styles/taskSearch.styl
```diff
@@ -0,0 +1,4 @@
+@import colors
+
+.remove-query-param
+    color $red
```
TrevorBurnham: image

Please turn on your text editor's "ensure newline at end of file" feature to avoid superfluous diffs in the future. :rocket:

SingularityUI/app/components/common/Table.cjsx
Calvinp: It looks like react-utils is an internal HubSpot library, not public.
Singularity is open source, so I don't think I can use that.

TrevorBurnham: Ah, right. You'll want [JedWatson/classnames](https://github.com/JedWatson/classnames). It's [officially recommended](https://facebook.github.io/react/docs/class-name-manipulation.html)!

SingularityUI/app/components/taskSearch/QueryParam.cjsx
Calvinp: This is now implemented.

PR #891, opened 2016-02-10
use pushState for log line links

tpetr:
no longer opens a new tab when you click on a permalink to a log line

### DIFF
--- a/SingularityUI/app/router.coffee
+++ b/SingularityUI/app/router.coffee
@@ -89,7 +89,7 @@ class Router extends Backbone.Router
         app.bootstrapController new TaskDetailController {taskId, filePath}
 
     tail: (taskId, path = '') ->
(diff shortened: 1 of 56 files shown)


SingularityUI/app/router.coffee
```diff
@@ -105,7 +105,7 @@ class Router extends Backbone.Router
         app.bootstrapController new DeployDetailController {requestId, deployId}
 
     aggregateTail: (requestId, path = '') ->
-        offset = window.location.hash.substr(1) || null
+        offset = parseInt(window.location.hash.substr(1)) || null
```
Friss: Should pass a radix to `parseInt` to make sure its always parsed the same `parseInt(window.location.hash.substr(1),10)`

tpetr: thanks, done

PR #894, opened 2016-02-12
Task kill checkbox label

Calvinp:
The message for when you killed a task was not a label, so could not be clicked to select the corresponding checkbox.
On bounce, the label that should have been for incremental bounce = false was applying to incremental bounce = true.
This fixes both of those bugs.


ssalinas:
LGTM

tpetr:
LGTM

PR #901, opened 2016-02-17
Lock brunch to version 2.2.x

Calvinp:
The new version released this morning breaks the build, so until we can bring the build in line with the new version's change we must lock brunch to the old version.


ssalinas:
Thanks @Calvinp LGTM

PR #902, opened 2016-02-17
Fix time display emails

Calvinp:
In pause emails, when an expiration time was set:
- The timestamp displayed in a confusing way (24h clock always UTC). Now the format and timezone are configurable, and it defaults to 12h clock (but still UTC).
- A space was missing after the timestamp. This is fixed.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
@@ -77,7 +77,7 @@ public class SingularityConfiguration extends Configuration {
   @Min(value = 1, message = "Must be positive and non-zero")
   private int defaultBounceExpirationMinutes = 60;
 
-  private String defaultDatePattern = "MMM dd h:mm:ss a zzz";
+  private String mailerDatePattern = "MMM dd h:mm:ss a zzz";
 
   @NotNull
   private SlavePlacement defaultSlavePlacement = SlavePlacement.GREEDY;
@@ -188,7 +188,7 @@ public class SingularityConfiguration extends Configuration {
 
   private long startNewReconcileEverySeconds = TimeUnit.MINUTES.toSeconds(10);
 
-  private TimeZone timeZone = TimeZone.getTimeZone("UTC");
+  private TimeZone mailerTimeZone = TimeZone.getTimeZone("UTC");
 
   @JsonProperty("ui")
   @Valid
@@ -961,19 +961,19 @@ public class SingularityConfiguration extends Configuration {
     this.maxStaleTasksPerRequestInZkWhenNoDatabase = maxStaleTasksPerRequestInZkWhenNoDatabase;
   }
 
-  public String getDefaultDatePattern() {
-    return defaultDatePattern;
+  public String getMailerDatePattern() {
+    return mailerDatePattern;
   }
 
-  public void setDefaultDatePattern(String defaultDatePattern) {
-    this.defaultDatePattern = defaultDatePattern;
+  public void setMailerDatePattern(String mailerDatePattern) {
+    this.mailerDatePattern = mailerDatePattern;
   }
 
-  public TimeZone getTimeZone() {
-    return timeZone;
+  public TimeZone getMailerTimeZone() {
+    return mailerTimeZone;
   }
 
-  public void setTimeZone(TimeZone timeZone) {
-    this.timeZone = timeZone;
+  public void setMailerTimeZone(TimeZone mailerTimeZone) {
+    this.mailerTimeZone = mailerTimeZone;
   }
 }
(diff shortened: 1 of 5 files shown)


SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -38,9 +38,13 @@
 
   private final Optional<String> uiBaseUrl;
   private final Optional<SMTPConfiguration> smtpConfiguration;
+  private String taskDatePattern;
```
wsorenson: make this final

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -76,6 +77,8 @@
   @Min(value = 1, message = "Must be positive and non-zero")
   private int defaultBounceExpirationMinutes = 60;
 
+  private String defaultDatePattern = "MMM dd h:mm:ss a zzz";
```
wsorenson: If these config properties are going to be used for mail only, they should indicate in the name.

Calvinp: I'm not sure if they're going to be used for mail only or not - it's plausible that we would want to use them for something else. I don't know what that thing would be though.

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
Calvinp: Done

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -38,9 +38,13 @@
 
   private final Optional<String> uiBaseUrl;
   private final Optional<SMTPConfiguration> smtpConfiguration;
+  private final String taskDatePattern;
+  private final TimeZone timeZone;
 
   @Inject
   public MailTemplateHelpers(SandboxManager sandboxManager, SingularityConfiguration singularityConfiguration) {
+    taskDatePattern = singularityConfiguration.getDefaultDatePattern();
+    timeZone = singularityConfiguration.getTimeZone();
```
ssalinas: lets follow the format of the ones below with this.x = ...

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
ssalinas: I think we should still make it more specific. At the moment, it is only used for mail, so let's make the name reflect that. Also it isn't really a default that we are falling back on, it is just a setting that will be used by everything. So maybe just calling it `mailerDatePattern` ?

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -185,6 +188,8 @@
 
   private long startNewReconcileEverySeconds = TimeUnit.MINUTES.toSeconds(10);
 
+  private TimeZone timeZone = TimeZone.getTimeZone("UTC");
```
ssalinas: Will this initialize correctly from a String? (i.e. reading it from a yaml field set to 'EST' or 'UTC') It very well might, but something we should probably test to be sure

Calvinp: Yes, it is still able to start up.
I renamed the field to mailerTimeZone, and putting a top level line 'mailerTimeZone: UTC' into the yaml doesn't break the build.

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -76,6 +77,8 @@
   @Min(value = 1, message = "Must be positive and non-zero")
   private int defaultBounceExpirationMinutes = 60;
 
+  private String mailerDatePattern = "MMM dd h:mm:ss a zzz";
```
tpetr: we should move these mailer* config fields into `SMTPConfiguration`

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -3,9 +3,11 @@
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.TimeZone;
 
 import org.apache.commons.lang3.text.WordUtils;
 import org.apache.commons.lang3.time.DateFormatUtils;
+import org.apache.commons.lang3.tuple.Pair;
```
tpetr: i dont think this import is necessary

PR #903, opened 2016-02-17
ensure run once tasks don't get launched at startup

tpetr:
- Update SingularityStartup logic to ignore both `ON_DEMAND` and `RUN_ONCE` requests.
- Add unit test to assert that `RUN_ONCE` requests aren't launched at startup

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
@@ -206,7 +206,7 @@ public class SingularityRequest {
 
   @JsonIgnore
   public boolean isOneOff() {
-    return requestType == RequestType.ON_DEMAND || requestType == RequestType.RUN_ONCE;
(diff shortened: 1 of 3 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
```diff
@@ -206,7 +206,7 @@ public boolean isAlwaysRunning() {
 
   @JsonIgnore
   public boolean isOneOff() {
-    return requestType == RequestType.ON_DEMAND;
+    return requestType == RequestType.ON_DEMAND || requestType == RequestType.RUN_ONCE;
```
tpetr: @wsorenson do you agree with this approach, or would you rather I update the check in [SingularityStartup](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java#L138)?

PR #906, opened 2016-02-19
Even number across racks

Calvinp:
When scaling a rackSensitive request, it is best to scale to a number of instances that is a multiple of the number of racks to evenly spread load across each rack.

With this change, if scaling is selected but the target number of instances is not a multiple of the number of racks, a new dialog will pop up notifying the user of this fact. It has radio buttons to choose between the number they entered, the multiple one lower than the number they entered (but 0 is never suggested), and the multiple one higher than the one they entered. 
So, if there are 3 racks and the user tries to scale to 5 instances, they will be prompted to choose between 3, 5 and 6 instances.
The radio button corresponding to the number the user entered is selected by default, so if the user fat fingers the ok button they will still get what they originally wanted.

### DIFF
--- a/SingularityUI/app/models/Request.coffee
+++ b/SingularityUI/app/models/Request.coffee
@@ -245,18 +245,17 @@ class Request extends Model
             else
                 callback()
 
-    promptScaleEvenNumberRacks: (data, mod, bounce, incremental, message, duration, callback) =>
+    promptScaleEvenNumberRacks: (scaleData) =>
         vex.dialog.open
             message: scaleEvenNumbersTemplate
-                instances: parseInt(data.instances)
-                notOneInstance: parseInt(data.instances) != 1
+                instances: parseInt(scaleData.data.instances)
+                notOneInstance: parseInt(scaleData.data.instances) != 1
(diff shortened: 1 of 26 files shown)


SingularityUI/app/models/Request.coffee
```diff
@@ -235,7 +238,46 @@ class Request extends Model
...
+            callback: (data) =>
+                return unless data
+                scaleData.data.instances = data.instances
+                @callScale scaleData.data, scaleData.bounce, scaleData.incremental, scaleData.message, scaleData.duration, scaleData.callback
+
+    checkScaleEvenNumberRacks: (data, bounce, incremental, message, duration, callback) =>
+        return 0 unless @attributes.request.rackSensitive
+        if not @racks
+            @racks = new Racks []
+            @racks.fetch { async: false }
```
ssalinas: We should be sure to catch errors here. If the call to fetch the number of racks fails for some reason, we should allow the user to continue without the extra prompt

ssalinas:
This should be gtg now that we can optionally opt out of the messages, thanks @Calvinp

PR #908, opened 2016-02-19
Fuzzy sort better adjusted score

Calvinp:
Requests and tasks that were searched from the requests (or tasks) page would sometimes filter to the bottom of search results despite containing the search term. This is because the fuzzy search algorithm only favored requests that started with the term (and not even that for tasks).
This fixes that, giving the highest score to requests or tasks that start with the search term, a medium score to requests or tasks that contain the search term, and a low score to others.

### DIFF
--- a/SingularityUI/app/utils.coffee
+++ b/SingularityUI/app/utils.coffee
@@ -198,5 +198,13 @@ class Utils
         else
           {}
 
+    @fuzzyAdjustScore: (filter, fuzzyObject) ->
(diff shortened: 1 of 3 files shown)


ssalinas:
logic looks good. Now that this is replicated in two places, can we move the score adjusting logic maybe into `utils.coffee`? Then we just have one place to update if we want to change it

Calvinp:
The score adjusting logic is moved into utils.

PR #913, opened 2016-02-19
Healthcheck notification won't display if task state is TASK_FAILED

Calvinp:
The notification that the task failed due to no passing healthchecks was being displayed even for tasks that failed on their own for other reasons.

This fixes that.

### DIFF
--- a/SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
+++ b/SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
@@ -14,6 +14,7 @@ class taskHealthcheckNotificationSubview extends View
 
     render: =>
         return if not @model.synced
+        return if _.last(@model.attributes.taskUpdates).taskState is "TASK_FAILED" # This is unfortunately the only place last state is stored.
         @$el.html @template @renderData()
 
     renderData: =>


SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
```diff
@@ -14,6 +14,7 @@ class taskHealthcheckNotificationSubview extends View
 
     render: =>
         return if not @model.synced
+        return if _.last(@model.attributes.taskUpdates).taskState is "TASK_FAILED" # This is unfortunately the only place last state is stored.
```
ssalinas: We are pulling the task state in the same way in other places as well, no need for the comment here. Alternatively, if you want it cleaner here, you could add a line in `parse` on the model for the task and set an attribute called `currentState` using the code above. Then we have it for reference later on on each task model that we look at

Calvinp: Changed.

ssalinas:
:+1:

SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
```diff
@@ -14,6 +14,7 @@ class taskHealthcheckNotificationSubview extends View
 
     render: =>
         return if not @model.synced
+        return if @model.attributes.lastKnownState is "TASK_FAILED"
```
zdhickman: Single quotes for this?

PR #916, opened 2016-02-22
New task checker healthcheck retries

ssalinas:
Based off deploy failures branch (relevant commits 7d7474c 7e68809)

uses the existing code from the deploy health helper to check if a task is has passed health checks or not


ssalinas:
This has been stable and working well, going to merge :balloon:

PR #920, opened 2016-02-24
Less ui build magic

tpetr:
This PR addresses build issues we've experienced with Bower + Brunch lately:
- Remove sketchy postinstall actions from `package.json`
- Update `maven-frontend-plugin` config to explicitly launch `npm install`, `bower install`, `brunch build`
- Launch brunch via a gulpfile. This is necessary due to how the `maven-frontend-plugin` works (no Brunch support), and also makes it easier for us to transition off of Brunch in the future.
- Explicitly clean assets folder before each UI build.

### DIFF
--- a/SingularityService/pom.xml
+++ b/SingularityService/pom.xml
@@ -539,9 +539,18 @@
                 <goals>
                   <goal>npm</goal>
                 </goals>
-                <configuration>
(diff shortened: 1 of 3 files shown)


tpetr:
Closing in favor of #923

PR #923, opened 2016-02-25
Die brunch

tpetr:
- Fully replace Brunch and Bower with Gulp
- Explicitly require all modules instead of assuming some will be in the global scope
- Copy `index.mustache` over to SingularityService instead of depending on a symlink
- Replace ZeroClipboard with Clipboard.js

*drops mic*


kwm4385:
:bow:

PR #931, opened 2016-03-02
RFC 5545 recur schedule support

grepsr:
To get over cron/quartzCron limitations for my own projects, I tried to add RFC5545 support.
https://tools.ietf.org/html/rfc5545#section-3.3.10

I've been running this on my testing server with great success.

The scheduling is much more easy to configure and dynamic with RFC5545 (things like start date, end date, odd schedules like every 37 hours or every 93 days works with this)

Have not had the time to work on the UI to make the new ScheduleType appear; but once this is reviewed, adding that would be trivial.


SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -32,6 +33,7 @@
 import com.hubspot.singularity.WebExceptions;
 import com.hubspot.singularity.config.SingularityConfiguration;
 import com.hubspot.singularity.data.history.DeployHistoryHelper;
+import org.dmfs.rfc5545.recur.RecurrenceRule;
```
tpetr: can you sort these imports? if you're using Eclipse or IntelliJ, we have a formatter config here: https://github.com/HubSpot/Singularity/blob/master/eclipse/singularity_eclipse_formatter.xml

grepsr: Code re-formatted.

grepsr: Wow that re-formatted everything, perhaps I should just re-format the sections I added?

grepsr:
Not sure why the travis check keeps failing, it compiles file on my machine. Any idea how to go from here?

tpetr:
@grepsr the Travis output makes it look like there's a findbugs issue. How are you building it locally? Try running `mvn clean verify` to compile, test, package, and analyzes the code for possible bugs or style issues.

grepsr:
Thanks, I'm using mvn clean package

I'll try that.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
tpetr: Yeah, just your sections sounds good. Thanks!

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
```diff
@@ -28,110 +37,161 @@
...
+                        } else {
+                            nextRunAtTimestamp = 0;
+                        }
+                    }
+                    if (nextRunAtTimestamp <= 0) {
+                        String msg = String.format("No next run date found for %s (%s)", taskId, request.getRequest().getScheduleTypeSafe());
+                        LOG.warn(msg);
+                        exceptionNotifier.notify(msg, ImmutableMap.of("taskId", taskId.toString()));
+                        return Optional.absent();
+                    }
```
tpetr: could we refactor all this RFC5545 logic (lines 128-166) into its own class? it'd be nice to expose an interface similar to `CronExpression.getNextValidTimeAfter()` below

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
```diff
@@ -28,110 +37,161 @@
...
+            }
+            long nextRunAtTimestamp = 0;
+            final long now = System.currentTimeMillis();
+            try {
+                if (request.getRequest().getScheduleTypeSafe() == ScheduleType.RFC5545) {
+                    final String schedule = request.getRequest().getSchedule().get();
+                    final RecurrenceRule recurrenceRule = new RecurrenceRule(schedule);
+                    if (recurrenceRule.isInfinite()) {
+                        // set limit at 2100-01-01 00:00:00
+                        recurrenceRule.setUntil(new DateTime(2100, 1, 1, 0, 0, 0));
```
tpetr: is a limit necessary?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
grepsr: I actually thought about doing that too. I'll take care of it and commit

I'm not a Java guy, can you tell me where I should place the new RFC5545Schedule.java file into the folder structure? Should I put it inside com.hubspot.singularity folder inside the SingularityService module?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
grepsr: Not really necessary, I put that just in case there was an infinite loop some how due to user input.

Got a better idea?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
tpetr: No worries -- I'd suggest putting it under `com.hubspot.singularity.helpers`

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -662,21 +671,67 @@ private int numInstancesExpected(SingularityRequest request, SingularityPendingR
...
+              dtStart = formatter.parseDateTime(matcher.group(1));
+            }
+            else{
+              dtStart = org.joda.time.DateTime.now();
+              dtStart = dtStart.withSecondOfMinute(0);
+            }
+
+            startDateTime = new DateTime(dtStart.getYear(),(dtStart.getMonthOfYear() - 1), dtStart.getDayOfMonth(),
+                    dtStart.getHourOfDay(), dtStart.getMinuteOfHour(), dtStart.getSecondOfMinute());
+            RecurrenceRuleIterator timeIterator = recurrenceRule.iterator(startDateTime);
```
tpetr: i think we can eliminate this loop if we select the more recent of `now()` vs. `DTSTART`, what do you think?

grepsr: I think we can't do that, because `DTSTART` is used to loop by the library. So if DTSTART is a date in the past, `org.dmfs.rfc5545.recur.RecurrenceRule` will start looping from that day to get the next best date & time that is in the future.
I'm almost done refactoring this into a class, committing it in a bit.

grepsr:
I just refactored the code and moved everything into a helper class, much cleaner now.

grepsr:
This finally worked on JDK8 on Travis, but still fails on JDK7 - not sure I understand why, any thoughts? Thanks

grepsr:
Just wanted to check back on this and see if I could get any help getting this to work on JDK7? 
JDK8 compiles great - really want to help move this feature into production branch.

grepsr:
I just re-comitted the code without any change and it builds in both JDK7 and JDK8! I guess it was a Travis issue.

Where do we go from here?

tpetr:
Awesome! Taking another look at this today. We'll need to add some unit tests since this change affects the operation of the scheduler.

grepsr:
Great, please let me know if you need any help.

Also, I'm using this on production for the past 10 days and this is working great! I have about 200+ tasks scheduled and it schedules everything right on time.

grepsr:
A note about this feature. Lately I've realised that the tasks are usually overdue by 5 minutes before running. I have a feeling its because of some default setting that triggers the check

Is there some global setting that can help here? Perhaps deltaAfterWhichTasksAreLateMillis?

ssalinas:
So there are two things to check here (second one is most likely the issue since you mentioned things are overdue):

- Are the tasks being put in the pending queue with the correct time? You can check the tasks page in the ui and filter for scheduled. These should have the time they were expected to run at. If those times are not what you expect, that would indicate an error in the scheduler (i.e. choosing the wrong time based on the changes in this PR).
- Are you getting offers with enough resources? While the tasks are put in the pending queue with an expected time to run, Singularity can only start those tasks that are due to run when an offer comes in that has the appropriate resources. So, if you have a scheduled job that is due, but you aren't getting offers, or you are getting offers that don't have enough resources, that would cause the tasks to be overdue until an appropriate offer came in. If this is the case, you may need more resources in your cluster, or another running framework in the cluster may be hogging all the offers
- Only other thing I can think of is if you have `maxTasksPerOffer` set too low. This will limit the number of tasks Singularity can schedule from a single offer, meaning you then have to wait for more offers to schedule the next tasks. By default, there is no limit, so if you don't have it set in your config this isn't the issue

Hope that helps

grepsr:
Hi there, we've fixed the merge conflicts for this, and pushed to latest branch.

ssalinas:
closing in favor of #1071

PR #932, opened 2016-03-02
(WIP) task search efficiency

ssalinas:
Working on a few improvements:
- Speed up `getInactiveTaskIds` by finding them directly instead of finding active tasks then removing from an array of all tasks. We make the same number of zk calls either way, might as well have the results we want from them.
- batch together calls for getting `SingularityTaskHistoryUpdate`s
- Store `SingularityTaskIdHistory`s in the zk history path so we can access them more easily
/cc @tpetr @Calvinp


SingularityService/src/main/java/com/hubspot/singularity/data/CuratorAsyncManager.java
```diff
@@ -217,6 +220,51 @@ public void processResult(CuratorFramework client, CuratorEvent event) throws Ex
...
+
+    final CountDownLatch latch = new CountDownLatch(pathsMap.size());
+
+    final BackgroundCallback callback = new BackgroundCallback() {
+
+      @Override
+      public void processResult(CuratorFramework client, CuratorEvent event) throws Exception {
+        if (event.getStat() == null) {
+          objects.add(pathsMap.get(event.getPath()));
+          latch.countDown();
```
wsorenson: what's the point of these 2 statements? latch.countDown(); return; --> latch.countDown()

ssalinas: ah, yeah that could just be outside the if and without the return, will fix

jhaber: I may be paranoid but I usually do `try { x; } finally { latch.countDown(); }` just in case an exception is thrown the latch still gets decremented

wsorenson: good point

ssalinas:
@wsorenson Consolidated some of the methods in CuratorAsync so we aren't repeating as much
@jhaber put all the latch count downs in finally blocks as well

SingularityService/src/main/java/com/hubspot/singularity/data/CuratorAsyncManager.java
```diff
@@ -241,5 +282,60 @@ public void processResult(CuratorFramework client, CuratorEvent event) throws Ex
     }
   }
 
-}
+  protected <T, Q> Map<T, List<Q>> getAsyncAsMapThrows(final String pathNameForLogs, final Map<String, T> parentPathsMap, final String subpath, final Transcoder<Q> transcoder) throws Exception {
```
tpetr: method name seems kinda vauge

ssalinas:
Summary of changes:
- implement `notExists` in curator async to get inactive task ids without having to get active task ids then filter
- batch together any calls for task history updates using new `getAsyncAsMap` in curator async
- begin storing current `SingularityTaskIdHistory` for a task in the `/tasks/history/(requestId)/(taskId)` node, and write a migration to backfill. This will update on each status update for the task

SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
```diff
@@ -396,10 +399,42 @@ public boolean taskHistoryUpdateExists(SingularityTaskHistoryUpdate taskHistoryU
...
-  public SingularityCreateResult saveTaskHistoryUpdate(SingularityTaskHistoryUpdate taskHistoryUpdate) {
+  public SingularityCreateResult saveTaskHistoryUpdate(SingularityTaskHistoryUpdate taskHistoryUpdate, Optional<String> runId) {
     singularityEventListener.taskHistoryUpdateEvent(taskHistoryUpdate);
 
-    return create(getUpdatePath(taskHistoryUpdate.getTaskId(), taskHistoryUpdate.getTaskState()), taskHistoryUpdate, taskHistoryUpdateTranscoder);
+    try {
+      String updatePath = getUpdatePath(taskHistoryUpdate.getTaskId(), taskHistoryUpdate.getTaskState());
+      String updatesPath = getUpdatesPath(taskHistoryUpdate.getTaskId());
+      String historyPath = getHistoryPath(taskHistoryUpdate.getTaskId());
+      String reqeustPath = getRequestPath(taskHistoryUpdate.getTaskId().getRequestId());
```
tpetr: typo

ssalinas:
@wsorenson had to undo some of the consolidation of those async methods, turns out having the callbacks com from other methods, I was getting ConcurrentModification and related exceptions when trying to run things.

Also, implemented what I talked about with @tpetr and started storing the json of the `SingularityTaskIdHistory` object in the `/tasks/history/(requestId)/(taskId)` node. The benefit of this ebing that we can now just use one getAsync call to gather up many id histories instead of having to build each from multiple calls. At the moment, it will update this json on each status update so that the task is history has the most recent task state. Thoughts on that strategy?

wsorenson:
I'm not sure it's worth all that effort to duplicate data and re-save data on updates just to make admin task search more efficient.

ssalinas:
@wsorenson did some refactoring, zk migration is taken out, and methods in curator async each build their own callback then call a shared method to actually execute the query, watch the latch, and return results

PR #933, opened 2016-03-03
Tail Logs From Bottom of File or Find First Error

Calvinp:
Working on tailing the log from the bottom of the file instead of the top.


ssalinas:
Hey @Calvinp seems this has a bunch of extra diff in there, can we merge master into here so the run now prompt diff isn't in this as well? Should be good to merge soon, just want to check the changes again

Calvinp:
Done

ssalinas:
This looks gtg, thanks @Calvinp

PR #934, opened 2016-03-04
Don't display healthcheck message for tasks in state TASK_LOST or TASK_FINISHED

Calvinp:
Missed these potential task states - now if a task is lost or finishes before having a successful healthcheck, it will not give a misleading healthcheck message.


ssalinas:
LGTM

PR #936, opened 2016-03-07
Add ability to search through tasks table by request type

Calvinp:
This will break bookmarks and other URLs into the task table.


ssalinas:
Now that we are using the `requestsTable/requestsFilter` template both in requests and in tasks, can we move it up a directory, and maybe rename it to `requestTypeFilter.hbs`? Bit more descriptive, and moves it out of a folder specific to requests since it's now shared. Other than that looks good

Calvinp:
@ssalinas Done

ssalinas:
LGTM

PR #937, opened 2016-03-07
Refactor HistoricalTasks Collections into One

Calvinp:
The RequestHistoricalTasks, DeployHistoricalTasks and TaskSearch collections were all almost exactly the same (except they hit slightly different endpoints). This refactors them to actually be the same collection.


SingularityUI/app/collections/RequestHistoricalTasks.coffee
```diff
@@ -1,25 +0,0 @@
...
-
-    comparator: (task0, task1) =>
-        -(task0.get("updatedAt") - task1.get("updatedAt"))
-
-    parse: (data) ->
-        for task in data
-            task.id = task.taskId.id
-        data
-
-    getTasksForDeploy: (deployId) ->
```
ssalinas: didn't see it anywhere, but assuming this method is unused since you didn't reimplement it, is that correct?

Calvinp: Correct.

PR #938, opened 2016-03-07
Fix Duplicates in Fuzzy Search

Calvinp:
Fuzzy Search gave duplicates in certain situations. This is because there are two searches done, one based on task id and one based on host. Fuzzy returns an object containing the string it searched on. If both the task id and host match the filter, two non-equal fuzzy objects representing the same task are returned, and _.union doesn't see them as the same thing.

This fixes that problem by calling _.uniq after extracting the original object.


Calvinp:
The duplicate 'p' in the branch name was a typo, though I wish it was an intentional pun.

ssalinas:
Thanks for the fix @Calvinp , LGTM

PR #940, opened 2016-03-11
Task Metadata UI and Email

Calvinp:
This adds task metadata to the UI and task emails.
Error level metadata will be shown in a red alert at the top of the task page, showing at most 5 (with page buttons to see more if there are more). Warn level metadata will be shown in an orange alert, showing at most 1 (with page buttons to see more if there are more). Info level metadata will be shown in a table at the bottom of the task.


Calvinp:
Note that as of now, the level field is not yet implemented so tasks aren't actually sorted by level. This will change soon.

SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -21,6 +24,9 @@ Handlebars.registerHelper "ifAll", (conditions..., options)->
         return options.inverse @ unless condition?
     options.fn @
 
+Handlebars.registerHelper 'ifNotEmptyArray', (arr, options) ->
```
ssalinas: is this helper being used anywhere?

Calvinp: Nope! I originally was using it but forgot to delete it when I realized that it's not at all necessary to have a function like that.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskMetadata.java
```diff
@@ -48,11 +51,13 @@ public String getTitle() {
...
   }
 
+  public MetadataLevel getLevel() { return level; }
+
   @Override
   public int compareTo(SingularityTaskMetadata o) {
     return ComparisonChain.start()
         .compare(timestamp, o.getTimestamp())
-        .compare(type, o.getType())
+        .compare(type, o.getType()).compare(level, o.getLevel())
```
ssalinas: nit, to keep with styling let's put this on a new line

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -46,15 +47,32 @@ public MailTemplateHelpers(SandboxManager sandboxManager, SingularityConfigurati
...
     List<SingularityMailTaskHistoryUpdate> output = Lists.newArrayListWithCapacity(taskHistory.size());
 
     for (SingularityTaskHistoryUpdate taskUpdate : taskHistory) {
       output.add(
-          new SingularityMailTaskHistoryUpdate(
-              DateFormatUtils.formatUTC(taskUpdate.getTimestamp(), TASK_DATE_PATTERN),
-              WordUtils.capitalize(taskUpdate.getTaskState().getDisplayName()),
-              taskUpdate.getStatusMessage().or("")));
+              new SingularityMailTaskHistoryUpdate(
+                      DateFormatUtils.formatUTC(taskUpdate.getTimestamp(), TASK_DATE_PATTERN),
```
ssalinas: might want to double check your editor settings for indentations. Not a huge deal, but easier to look at the correct diffs if there aren't also indentation changes

SingularityUI/app/templates/taskDetail/taskMetadataTable.hbs
```diff
@@ -0,0 +1,76 @@
...
+                    {{else}}
+                        view
+                    {{/if}}
+                </a>
+            </small>
+        </h2>
+    </div>
+    <div class='{{#unless expanded}}hide{{/unless}}'>
+        <div class="table-container">
+            <table class="table table-striped">
```
ssalinas: We should check to see what some of the longer/more detailed metadata items look like in a table. Since we had to reformat the error/warn ones to look nicer in the panels at the top, might be worth the same investigation for the info ones

Calvinp: It's still kind of ugly, but it's at the bottom of the task page and starts collapsed. You wouldn't see it unless you want to find it, and even then it's very out of the way. So I think it's ok for now, we can iterate on it later if we want to.
screenshot 2016-03-11 14 23 09

Calvinp:
Something is wrong with emailing tasks with metadata at the moment. I'm looking into what exactly went wrong. Please do not merge this right now.

Calvinp:
The problem with emailing task metadata is fixed in this PR:
https://github.com/HubSpot/Singularity/pull/961

PR #941, opened 2016-03-11
Use gitbook for docs

ssalinas:
Reorganizing docs and displaying using gitbook. Eventual goal is be able to replace our github pages page with gitbook generated site.

ToDo:
- [x] Setup page with details for each release (like we have for 0.4.9 on the wiki). Fill out relevant details back to at least 0.4.5
- [x] Write more examples and docs for use of specific features, organize these in a folder so they can be pulled in in other locations by the gitbook
- [x] Update the ui page with more recent screenshots
- [x] Double check examples page and update to use an org repo
- [x] One more pass over all docs to be sure everything is up to date (examples are working, etc)

if you want to view the current state of the docs, on this branch you can `npm install -g gitbook` and run `gitbook serve`. Docs should be available on localhost:4000


ssalinas:
Good draft of this is done now, with a bunch of docs added for a few features, releases, updated ui images, etc. Let me know what you think

ssalinas:
Some notes on this:
- Changes to existing docs are just typo fixes, updating links, and a small bit of rewording. Most changes to the actual docs are additions of new docs
- Created an intro.md which the gitbook uses at its landing page. This is almost the same as the README.md, but I wanted to have the flexibility of the gitbook homepage and README being different
- Added a script that will build the latests swagger docs and split them up for the gitbook, build the gitbook, replace the favicon, and publish to gh-pages branch so that it is easy to update. This should only include the relevant files since the .bookignore excludes the actual source code

PR #943, opened 2016-03-11
Don't capture ctrl-t

Calvinp:
Capturing ctrl-t prevents someone on a Linux or Windows machine from using the common keyboard shortcut to open a new tab.


ssalinas:
LGTM

PR #944, opened 2016-03-11
More Descriptive Error Message On Http 200 But No JSON

Calvinp:
When a user's session expires, we send an http 302 redirect to the login page. However, the browser catches this and follows it without passing it off to the application. Then the application receives html when it was expecting JSON. This handles that problem by advising the user of the situation and that one possible cause is an expired session.

### DIFF
--- a/SingularityUI/app/application.coffee
+++ b/SingularityUI/app/application.coffee
@@ -116,13 +116,12 @@ class Application
                     serverMessage = JSON.parse(jqxhr.responseText).message or jqxhr.responseText
                 catch
                     if jqxhr.status is 200
-                        console.error jqxhr
+                        console.error jqxhr.responseText
                         Messenger().error
-                            message:    '''
-                                            <p>Expected JSON but received something else (possibly html). The response has been saved to your js console.</p>
-                                            <p>One possible cause is an http redirect to an html web page.</p>
-                                        '''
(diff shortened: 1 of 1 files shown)


SingularityUI/app/application.coffee
```diff
@@ -113,9 +113,17 @@ class Application
...
-                  serverMessage = JSON.parse(jqxhr.responseText).message or jqxhr.responseText
+                    serverMessage = JSON.parse(jqxhr.responseText).message or jqxhr.responseText
                 catch
-                  serverMessage = jqxhr.responseText
+                    if jqxhr.status is 200
+                        console.error jqxhr
+                        Messenger().error
+                            message:    '''
+                                            <p>Expected JSON but received something else (possibly html). The response has been saved to your js console.</p>
+                                            <p>One possible cause is an http redirect to an html web page.</p>
```
ssalinas: I still think we can be a bit smarter about the messaging here. We can know if it is html by checking if the string starts with `<!DOCTYPE html>`. Then we can either say '...but received html' or '...but received something else' if we don't know.

Also, The one possible cause is very specific to our own use case. If there is no way for us to know that that might be the error, we shouldn't be mentioning possible causes.

ssalinas:
LGTM

PR #949, opened 2016-03-14
Opt out of even number across racks.

Calvinp:
Adds a parameter to SingularityRequest to opt out of suggesting even numbers across racks, and the UI listens to this.
Also when creating or updating requests through the UI, there is a checkbox to toggle this.

### DIFF
--- a/SingularityUI/app/views/requestFormBase.coffee
+++ b/SingularityUI/app/views/requestFormBase.coffee
@@ -97,8 +97,6 @@ class RequestFormBase extends FormBaseView
             
             requestObject.rackAffinity = @getSelect2Val "#rackAffinity-#{ type }"
 
-            debugger
-
         if type in ['SCHEDULED', 'ON_DEMAND', 'RUN_ONCE']
             killOld = parseInt @$("#killOldNRL-#{ type }").val()
             requestObject.killOldNonLongRunningTasksAfterMillis = killOld if killOld
--- a/SingularityUI/app/views/requestFormBase.coffee
+++ b/SingularityUI/app/views/requestFormBase.coffee
@@ -92,8 +92,12 @@ class RequestFormBase extends FormBaseView
             requestObject.waitAtLeastMillisAfterTaskFinishesForReschedule = waitAtLeast if waitAtLeast
 
             requestObject.rackSensitive = @$("#rack-sensitive-#{ type }").is ':checked'
+
+            requestObject.hideEvenNumberAcrossRacksHint = @$("#hide-distribute-across-racks-hint-#{ type }").is ':checked'
             
             requestObject.rackAffinity = @getSelect2Val "#rackAffinity-#{ type }"
+
+            debugger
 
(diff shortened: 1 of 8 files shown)


Calvinp:
Possible improvements in future iterations:
- Add a configuration for whether requests will have the hint on or off by default (when the param isn't supplied)
- Make the checkbox not exist when 'rack sensitive' isn't checked
- Add a checkbox to the hint box to disable future hints for this request

ssalinas:
I think the last of those nice to haves (checkbox to ignore in the future) we should implement in this PR. Should just have to get the current request json, add the field, and POST it

Calvinp:
Working on the functionality, but how's this wording?

screenshot 2016-03-14 15 54 44

Note the fact that there are two '3' options is not a bug, it's because I cheated to get it to work with the one rack I have locally.

SingularityUI/app/views/requestFormBase.coffee
```diff
@@ -92,9 +92,13 @@ class RequestFormBase extends FormBaseView
             requestObject.waitAtLeastMillisAfterTaskFinishesForReschedule = waitAtLeast if waitAtLeast
 
             requestObject.rackSensitive = @$("#rack-sensitive-#{ type }").is ':checked'
+
+            requestObject.hideEvenNumberAcrossRacksHint = @$("#hide-distribute-across-racks-hint-#{ type }").is ':checked'
             
             requestObject.rackAffinity = @getSelect2Val "#rackAffinity-#{ type }"
 
+            debugger
```
tpetr: this shouldn't be committed

ssalinas:
Maybe just a simpler 'Don't show this prompt again for {requestId}' in terms of messaging

SingularityUI/app/views/requestFormBase.coffee
Calvinp: You are right, editing that out.

Calvinp:
Hmm I feel like it's important to specify that this applies to all users, not just the person who's clicking the button.

ssalinas:
That's fair, can add a 'for any user' to the end of the messaging, should also see if we can separate it just a bit more so it doesn't look like part of the radio buttons, the difference between the circle/square is pretty slight.

Calvinp:
I can get the text onto one line if I don't include the specific request id (keeping it as 'this request'). Which seems fairly reasonable because the request id is in the background on this screen:

screenshot 2016-03-14 16 15 36

Calvinp:
More separation:

screenshot 2016-03-14 16 18 12

ssalinas:
:+1:

Calvinp:
This is now implemented. Sending this request, plus the scale request, plus the bounce request (when scaling with bounce), and ensuring none of them conflicted caused a bit of trouble, but this is resolved now.

ssalinas:
Thanks for this @Calvinp , going to merge it into the the other PR branch since they go together

PR #950, opened 2016-03-14
allow multiple on demand requests to be queued up

ssalinas:
/cc @jhaber @jonathanwgoodwin

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
@@ -24,6 +24,7 @@ import com.hubspot.singularity.SingularityCreateResult;
 import com.hubspot.singularity.SingularityDeleteResult;
 import com.hubspot.singularity.SingularityDeployKey;
 import com.hubspot.singularity.SingularityPendingRequest;
+import com.hubspot.singularity.SingularityPendingRequest.PendingType;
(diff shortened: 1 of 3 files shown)


ssalinas:
@wsorenson @tpetr  traced back through the different `PendingTypes` we could possibly get in combination with the `ONEOFF` type. Those are: `RETRY`, `DECOMISSIONED_SLAVE_OR_RACK`, `UPDATED_REQUEST`, or `TASK_DONE`. No others are currently created for an on demand request.

`RETRY` -> create another identical oneoff, there is no conflict here as both kinds just schedule a single task with the relevant options
`DECOMISSIONED_SLAVE_OR_RACK` - > no problem there, no new tasks will be scheduled for a oneoff
`TASK_DONE` -> Similar, we do not schedule new tasks for a oneoff on task done
`UPDATED_REQUEST` -> This was the only possible relevant one (for values like `killOldNonLongRunningTasksAfterMillis` which I don't believe we wired up for on demand anyway). However, when we encounter this pending type the update to request data is already finished and we do not schedule any new oneoff tasks for a `UPDATED_REQUEST` pending type

ssalinas:
quick addition, misread an if statement, we do not get a `UPDATED_REQUEST` pending type for on demand requests

jhaber:
There are no words to express my excitement :dancers:

ssalinas:
This has been performing well with no issues, going to merge

PR #951, opened 2016-03-15
After submitting or editing a request, redirect to that request's page

Calvinp:
Rather than simply display a note saying the request was created, for most use cases it would probably be better to redirect back to that request's page. This implements that.


ssalinas:
LGTM, thanks @Calvinp

PR #952, opened 2016-03-15
Don't Lose Currently Typed Owner/Rack Affinity When Losing Focus

Calvinp:
This fixes the bug where, on the request edit or create form an owner or rack (for rack affinity) is being typed, then the user clicks or tabs causing the form to lose focus, the in-progress owner or rack would be lost.


Calvinp:
Note - to test this in QA (or, presumably, production), open the js console and run the command 'config.hideNewRequestButton = false'

ssalinas:
LGTM

PR #954, opened 2016-03-16
Linkify text in tags with 'linkify' class

Calvinp:
Linkify all text in <p> tags.


Calvinp:
There's a bunch of commits in here because it includes the die-brunch branch - otherwise there would be a bunch of merge conflicts when it is merged in.

ssalinas:
Let's get all those node_modules folders/files for linkify out of source control, other than that, should be good

Calvinp:
We changed this to linkify elements in tags with the class 'linkify' rather than all p tags.
The branch name is now misleading, but I edited the PR name.

PR #955, opened 2016-03-17
optional redirect in ui when receiving a 401 from the api

ssalinas:
based on brunch -> gulp changes to avoid merge conflicts, last commit is the relevant one


SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
```diff
@@ -77,6 +77,9 @@ public static RootUrlMode parse(String value) {
   @NotEmpty
   private String timestampWithSecondsFormat = "lll:ss";
 
+  @JsonProperty
+  private String redirectOnUnauthorizedUrl = "";
```
tpetr: we should use `Optional<String>` instead

SingularityUI/app/application.coffee
```diff
@@ -107,6 +107,8 @@ class Application
             if jqxhr.status is 502
                 Messenger().info
                     message:   "Singularity is deploying, your requests cannot be handled. Things should resolve in a few seconds so just hang tight!"
+            else if jqxhr.status is 401 and config.redirectOnUnauthorizedURL
+                window.location.href = config.redirectOnUnauthorizedURL
```
tpetr: we should have some sort of support for redirecting back to the original page once the user has logged in... something along the lines of `.replace('${URL}', window.location.href)` could work

PR #959, opened 2016-03-18
Thread check types

### LINKED ISSUE #741: Failed to parse cgroup information
I am trying to get singularity executor working on a mesos slave node.  Whenever I try to run a task using the signularity executor it fails to read the cgroup information.  The /prod/{PID}/cgroup directory exist but it contains no data.
I am running singularity executor 0.4.5 on  Ubuntu 12.04(Precise) and kernel 3.2.0-67-virtual on a amazon EC2 instance.

java.lang.RuntimeException: Unable to parse cgroup container from /proc/25758/cgroup
	at com.hubspot.singularity.executor.SingularityExecutorThreadChecker.getNumUsedThreads(SingularityExecutorThreadChecker.java:131) [SingularityExecutor-0.4.5-shaded.jar:0.4.5]
	at com.hubspot.singularity.executor.SingularityExecutorThreadChecker.checkThreads(SingularityExecutorThreadChecker.java:86) [SingularityExecutor-0.4.5-shaded.jar:0.4.5]
	at com.hubspot.singularity.executor.SingularityExecutorThreadChecker.access$000(SingularityExecutorThreadChecker.java:32) [SingularityExecutor-0.4.5-shaded.jar:0.4.5]
	at com.hubspot.singularity.executor.SingularityExecutorThreadChecker$1.run(SingularityExecutorThreadChecker.java:65) [SingularityExecutor-0.4.5-shaded.jar:0.4.5]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471) [na:1.7.0_07]
	at java.util.concurrent.FutureTask$Sync.innerRunAndReset(FutureTask.java:351) [na:1.7.0_07]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:178) [na:1.7.0_07]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:178) [na:1.7.0_07]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293) [na:1.7.0_07]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1110) [na:1.7.0_07]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:603) [na:1.7.0_07]
	at java.lang.Thread.run(Thread.java:722) [na:1.7.0_07]

ssalinas:
Although cgroup isolation is common, posix is the default. If we are not using cgroup isolation we will not find a cpu cgroup to parse and this will error every time. This PR does a few things:
- Make two types of thread checks, `CGROUP` and `PSTREE`
- Configurably fall back to `PSTREE` when `CGROUP` fails (defaults to false)

/fixes #741

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorThreadChecker.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorThreadChecker.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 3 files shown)


tpetr:
I like the idea of removing the cgroups requirement for thread limiting, but our original pstree implementation wasn't great. For example:

```bash
$ pstree 91309 -p
daemon-runner(91309)───su(91330)───daemon-runner(91338)─┬─python2.6(91408)
                                                        ├─tee(91339)
                                                        └─tee(91340)
```

The thread checker would conclude that process 91309 is using 3 threads, when it's really using 6. Are there any other UNIX facilities we can use to figure out the thread count? Do we know what `pstree` is using under the hood in order to generate this info?

ssalinas:
Good, point. Let me look and see what else we could use.

ssalinas:
Updated to have three possible thread checking types:
- cgroup
- ps
- /proc/<pid>/status

You can now configure the primary (defaults to cgroup) and a fallback (optional). If the fallback is set, it will be used in the case that the primary type does not result in a valid value

PR #961, opened 2016-03-18
Fix stray '-' that was preventing task email from sending when it had…

Calvinp:
This fixes an issue with task metadata caused by a stray non-escaped dash character.


Calvinp:
There are also changes to bring the metadata section formatting in line with the formatting of the rest of the Jade template.

ssalinas:
:+1:

Issue #897, opened 2016-02-16
Docker parameters in UI


scalp42:
Would like to open this issue regarding the lack of UI on the tasks side.

It'd be great if Singularity could allow a "write" access by adding the the API resources in the new request UI bits:

![](https://i.imgur.com/QQBChE6.png)

Using Singularity 0.4.9, it's really limited (even more than in v0.4.5).

Maybe keep the current UI but allow an attribute to be specified in the YAML to allow adding tasks parameters like Marathon:

![](https://i.imgur.com/vRKfsSY.png)

Thoughts ?

ssalinas:
@scalp42 most of what you are looking for is on the new deploy page. On the page for the request you just created, click on deploy and you will get a set of fields very similar to that. The request holds over-arching information for all deploys (number of instances, type of request, etc), while the individual deploy has information about the resources, docker image, ports, ...

scalp42:
@ssalinas found it, but it's still more or less the same as in you can't pass a command for example to override the container one (you have to rely on the container one).

ssalinas:
fixed by #962

PR #962, opened 2016-03-18
Some ui fixes

### LINKED ISSUE #317: (ui) On New Request page, Instances form field and label are too narrow (Firefox Only)
Actually the whole data entry portion of the page seems to be to have its width incorrectly constrained.

### LINKED ISSUE #321: (ui) The "form" tag on the "New deploy" page has no "action" property
As a result it is doing a GET request to the API instead of the desired POST.

### LINKED ISSUE #945: No easy way to pass multiple args to ON_DEMAND jobs from the UI
When I've setup a docker ON_DEMAND job and tried passing in args from the UI, I see that the command line input is not split on whitespace.  Here's the relevant bits of doing a `docker inspect` after the fact on a container that was launched by singularity in this way:

```
$ sudo docker inspect 54f8c357cd4f
[
    {
        "Id": "54f8c357cd4fa214fefae11bbf27ab3b5851c49861f2fa3fb280ed716100d613",
        "Created": "2016-03-11T19:34:39.312828156Z",
        "Path": "/usr/bin/node",
        "Args": [
            "/app/lib/echo.js hi"
        ],
```

The command line I passed in the UI was: `/app/lib/echo.js hi`.  the --entrypoint I specified for the container was `/usr/bin/node`.  What I was hoping for was that the command line input I specified in the singularity UI would be passed as separate args (split on whitespace).  It seems like the optional command line input you specify in the UI should either a) split that input on whitespace or b) the UI should allow you to specify multiple arguments (one textfield for each).

fwiw - There is a similar (multiple textfields) capability over in the ENV var setup for new container deploys.

ssalinas:
- fix a missing argument for handling an event /fixes #321
- allow mutliple command line inputs and clean up the code around keeping history/rerunning /fixes #945 

/cc @dtabwhite @eherot

### DIFF
--- a/SingularityUI/app/models/Request.coffee
+++ b/SingularityUI/app/models/Request.coffee
@@ -357,7 +357,7 @@ class Request extends Model
             lastCommands = JSON.parse(localStorage.getItem(@localStorageCommandLineInputKeyPrefix + @id))
         catch e
             console.error('Could not parse previous commands JSON')
-            lastCommands = [lastCommands.length - 1]
+            lastCommands = []
         vex.dialog.prompt
             message: "<h3>Run Task</h3>"
             input: runTemplate
@@ -376,7 +376,7 @@ class Request extends Model
                 fileName = @data.filename.trim()
                 message = @data.message
 
-                if fileName.length is 0 and @data.autoTail is 'on'
+                if fileName and fileName.length is 0 and @data.autoTail is 'on'
                     $(window.noFilenameError).removeClass('hide')
                     return false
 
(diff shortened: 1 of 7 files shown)


ssalinas:
WIP => when done also /fixes #897

ssalinas:
Finished changes now, overall:
/fixes #897
/fixes #321 
/fixes #317 
/fixes #945

bradjorgensen:
Does this pull request add support for properties like requiredSlaveAttributes on the request create/edit forms?

ssalinas:
@bradjorgensen it is not in this PR, we can add it into another one

PR #965, opened 2016-03-18
warning if singularity has no leader or leader has no mesos connection

### LINKED ISSUE #776: Show big visual indicator if Singularity loses connectivity with Mesos
If Singularity has issues (re)registering with Mesos, we'll only know about it via the `mesos running:` part of log lines like this:

```
TRACE [2015-11-23 23:50:32,160] com.hubspot.singularity.scheduler.SingularityLeaderOnlyPoller: Skipping SingularitySchedulerPoller (period: 00:05.000) (leadership: true, mesos running: false, enabled: true)
```

We need to:
- Expose this value in the state endpoint
- Make an obnoxious visual indicator in the UI so that users know something is up
- Link to documention on how to fix

/cc @scalp42

ssalinas:
An alert message will show at the top of the status page, a `Connected` column also appears in the table of scheduler hosts. Let me know if you have any inputs on the warning message itself.
/fixes #776 

screen shot 2016-03-18 at 3 23 55 pm

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosScheduler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosScheduler.java
@@ -428,4 +428,8 @@ public class SingularityMesosScheduler implements Scheduler {
     LOG.warn("Error from mesos: {}", message);
   }
 
+  public boolean isConnected() {
+    return schedulerDriverSupplier.get().isPresent();
+  }
+
 }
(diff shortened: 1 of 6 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosScheduler.java
```diff
@@ -428,4 +428,8 @@ public void error(SchedulerDriver driver, String message) {
     LOG.warn("Error from mesos: {}", message);
   }
 
+  public boolean isConnected() {
+    return schedulerDriverSupplier.get().isPresent();
```
tpetr: need to handle absent case

tpetr:
LGTM

PR #966, opened 2016-03-21
Consistent terminology and more ovbious deploy failure on task page

Calvinp:
This includes several UI improvements around task and deploy failures:

On the task page:
- Healthcheck failed notification now states that the task was killed, rather than task failed
- If the healthcheck failed because a connection was refused, the healthcheck failed notification will state that it was probably a problem with the app, not the cluster
- The healthcheck notification will not display if the deploy failed and killed the task after one healthcheck ran and failed (but not enough to kill the task) 
- If this task caused the deploy to fail, it will have a danger notification saying this (and linking to other causes)
- If this task's deploy failed, but this task was not a cause of that failure, then it will have a warn notification. This notification will state if the task was killed because the deploy failed, and it will link to failure causes.

On the deploy page:
- Rather than say (n) tasks failed, now the message states how many causes of deploy failure there were. Those two numbers don't always match and caused some confusion.
- The links to the tasks that failed will now have the taskIds bolded to make it more clear that they are links to the task pages.

### DIFF
--- a/SingularityUI/app/templates/taskDetail/taskDeployFailureNotification.hbs
+++ b/SingularityUI/app/templates/taskDetail/taskDeployFailureNotification.hbs
@@ -1,0 +1,46 @@
+{{#ifEqual deploy.deployResult.deployState "FAILED"}}
+    {{#ifCauseOfFailure data deploy }}
+        <div class="alert alert-danger" role="alert">
+            <p>This task casued <a href='{{appRoot}}/request/{{deploy.requestId}}/deploy/{{deploy.deployId}}'>Deploy {{deploy.deployId}}</a> to fail. Cause: {{causeOfDeployFailure data deploy }}</p>
+            {{#ifGT deploy.deployResult.deployFailures.length 1}}
+                <p>The deploy failure was also caused by:</p>
+                <ul>
+                    {{#each deploy.deployResult.deployFailures}}
+                        {{#if taskId}}
+                            {{#ifNotEqual taskId.id ../data.taskId}}
+                                <li><a href='{{appRoot}}/task/{{taskId.id}}'>{{taskId.id}}</a>: {{humanizeText reason}} {{#if message}}({{ message }}){{/if}}</li>
+                            {{/ifNotEqual}}
+                        {{else}}
+                            <li>{{humanizeText reason}} {{#if message}}({{ message }}){{/if}}</li>
+                        {{/if}}
+                    {{/each}}
+                </ul>
+            {{/ifGT}}
+        </div>
+    {{else}}
+        <div class="alert alert-warning" role="alert">
+            <p>
+                <a href='{{appRoot}}/request/{{deploy.requestId}}/deploy/{{deploy.deployId}}'>Deploy {{deploy.deployId}}</a> failed.
+                {{#ifDeployFailureCausedTaskToBeKilled data}}
+                    This task was killed as a result of the failing deploy.
+                {{/ifDeployFailureCausedTaskToBeKilled}}
+                {{#if deploy.deployResult.deployFailures.length}}
+                    The deploy failure was caused by:
+                {{else}}
+                    The reason for the deploy failure is unknown.
+                {{/if}}
+            </p>
+            {{#if deploy.deployResult.deployFailures.length}}
+                <ul>
+                    {{#each deploy.deployResult.deployFailures}}
+                        {{#if taskId}}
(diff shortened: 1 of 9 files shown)


SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
```diff
@@ -18,12 +18,24 @@ class taskHealthcheckNotificationSubview extends View
         @listenTo @model, 'sync', @render
         @listenTo @pendingDeploys, 'sync', @render
 
+    deployFailureKilledTask: =>
```
ssalinas: Doesn't look like this method is used. You can probably eliminate the section below where it is calculated in `renderData` and just use this

SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
```diff
@@ -45,11 +57,19 @@ class taskHealthcheckNotificationSubview extends View
...
         lastHealthcheckFailed: @model.get('healthcheckResults')?.length > 0 and @model.get('healthcheckResults')[0].statusCode isnt 200
+        healthcheckFailureReasonMessage: @healthcheckFailureReasonMessage()
         synced:           @model.synced
         config:           config
         tooManyRetries: @model.get('healthcheckResults').length > maxRetries and maxRetries != 0
         numberFailed: @model.get('healthcheckResults').length
         secondsElapsed: healthTimeoutSeconds
+        doNotDisplayHealthcheckNotification: deployFailureKilledTask
+
+    healthcheckFailureReasonMessage: () ->
```
ssalinas: A few things here. First, the method name makes me think we are calculating reasons for any healthcheck failure, when in fact we are only looking for connection refused.

Second thing, the connection refused message isn't very clear. The general confusion is that users see 'connection refused to (slave host):(port)' and assume it's a problem with the slave and don't think to check their own app. We need to clarify that the connection refused was to the app (i.e. the app did not start or wasn't listening on the port). Maybe something more like:

'Healthcheck failed due to a refused connection. It is possible your app did not start properly or was not listening on the anticipated port'

SingularityUI/app/templates/taskDetail/taskDeployFailureNotification.hbs
```diff
@@ -0,0 +1,44 @@
+{{#ifEqual deploy.deployResult.deployState "FAILED"}}
```
ssalinas: if our task 404s you are not fetching the deploy info and are instead setting it to '', this if statement will throw 'has no attribute deployResult ...' type errors. Probably want to check that deploy is defined first

SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
Calvinp: I was getting errors using the function at first, but it seems to be working now so I'll use the function.

SingularityUI/app/views/taskHealthcheckNotificationSubview.coffee
Calvinp: I named that healthcheckFailureReasonMessage because I figured if we wanted to add something else in there later we could do so in that function rather than have the logic spread around. For now, yes, it only looks for connection refused.

I will modify the message to make it more clear.

SingularityUI/app/templates/taskDetail/taskDeployFailureNotification.hbs
Calvinp: Fixed

Calvinp:
[This task](https://private.hubteamqa.com/singularity/task/ORION_TestService_incremental_deploy-web-490_11_11-1455804940908-1-twilight_penguin.iad03.hubspot_networks.net-us_east_1a)'s failed healthchecks caused its deploy to fail. There weren't enough failed healthchecks to fail the task, but there were enough to fail the deploy. This then caused the task to be killed.

It is probably appropriate to show the healthcheck notification in this edge case. It is not being shown right now because this PR doesn't show the notification when the failing deploy caused the task to fail. Maybe if this task is one of the deploy failure causes, show the healthcheck notification anyway. That might cause other edge case failures (task triggers a deploy failure for some reason other than failing healthchecks -> deploy fails -> task killed), but I'm not sure how to detect and account for all possibilities.

ssalinas:
This has been looking good, messaging is much more clear, thanks @Calvinp

PR #968, opened 2016-03-23
Better Slave Decommissioned / Offline Communication

Calvinp:
- The alert for a slave decommissioning is now at warn level, not danger level
- The wording of the alert now suggests that this is a normal Singularity action, nothing is on fire, your task has been replaced, and you can continue about your day without (much) worry
- If a slave is offline, the scary ajax error stating so is now caught if it appears when trying to fetch files
- Instead of the scary ajax error, the files table now says that the slave is offline so we don't have files.


ssalinas:
Thanks for this @Calvinp , messaging is much clearer (and less alarming!) now

PR #970, opened 2016-03-24
Task badge color for TASK_KILLED is now default (grey)

Calvinp:
The color for TASK_KILLED badges is now grey, the bootstrap default color, to distinguish it from TASK_FAILED.

The two getLabelClass functions were combined into one helper function in utils to avert confusion when changing these colors in the future.


ssalinas:
Much easier to distinguish failed/lost from the killed tasks now. And +1 for consolidating the colors all to one place for these

Calvinp:
Thanks for the +1, though I'm not sure how well deserved it is on my part since I'm the one who split the colors in the first place.

PR #971, opened 2016-03-24
ability to update request data upon successful deploy

ssalinas:
Currently, if you want to change the scale of a service (or any other request-level setting) as well as deploy new code, you need to do this in two separate steps. This PR adds an optional `newRequestData` field to the `SingularityDeployRequest`. If present, it will be added to the pending deploy so that the scheduler and deploy checker will use this new request data for the new deploy. If the deploy succeeds the request is then updated. No other updates to request data are allowed if there is currently a pending deploy that also updates request data.

@tpetr @wsorenson


SingularityBase/src/main/java/com/hubspot/singularity/SingularityPendingDeploy.java
```diff
@@ -10,14 +10,16 @@
   private final Optional<SingularityLoadBalancerUpdate> lastLoadBalancerUpdate;
   private final DeployState currentDeployState;
   private final Optional<SingularityDeployProgress> deployProgress;
+  private final Optional<SingularityRequest> newRequestData;
```
tpetr: naming seems a little awkward in my opinion -- `updatedRequest` might make more sense

ssalinas: yeah, went back and forth on the name like 10 times already. Open to changing for sure, `updatedRequest` sounds good

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityDeployRequest.java
```diff
@@ -12,15 +13,22 @@
...
   private final Optional<String> message;
+  private final Optional<SingularityRequest> newRequestData;
 
   @JsonCreator
   public SingularityDeployRequest(
       @JsonProperty("deploy") SingularityDeploy deploy,
       @JsonProperty("unpauseOnSuccessfulDeploy") Optional<Boolean> unpauseOnSuccessfulDeploy,
-      @JsonProperty("message") Optional<String> message) {
+      @JsonProperty("message") Optional<String> message,
+      @JsonProperty("newRequestData") Optional<SingularityRequest> newRequestData) {
```
tpetr: is there a reason this should be here, as opposed to on the `SingularityDeploy` object?

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -87,25 +87,31 @@ public SingularityRequestParent deploy(@ApiParam(required=true) SingularityDeplo
     final String requestId = checkNotNullBadRequest(deploy.getRequestId(), "DeployRequest must have a non-null requestId");
 
     SingularityRequestWithState requestWithState = fetchRequestWithState(requestId);
-    SingularityRequest request = requestWithState.getRequest();
 
     authorizationHelper.checkForAuthorization(requestWithState.getRequest(), user, SingularityAuthorizationScope.WRITE);
 
+    SingularityRequest newRequestData = requestWithState.getRequest();
+    boolean hasUpdatedRequest = false;
```
tpetr: i think you can short-circuit this to `final boolean hasUpdatedRequest = deployRequest.getNewRequestData().isPresent()`, since an invalid updated request will cause an exception

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityDeployRequest.java
ssalinas: I had it in the deploy initially. But, the usage of this new field would only be at deploy time. After the deploy fails or succeeds we are back to using the normal request data. Having request data that stuck around with the deploy seemed like it could be more confusing. Made more sense to put it outside the deploy so it didn't stick around

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
ssalinas: good point, I'll update

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -87,25 +87,31 @@ public SingularityRequestParent deploy(@ApiParam(required=true) SingularityDeplo
     final String requestId = checkNotNullBadRequest(deploy.getRequestId(), "DeployRequest must have a non-null requestId");
 
     SingularityRequestWithState requestWithState = fetchRequestWithState(requestId);
-    SingularityRequest request = requestWithState.getRequest();
 
     authorizationHelper.checkForAuthorization(requestWithState.getRequest(), user, SingularityAuthorizationScope.WRITE);
 
+    SingularityRequest newRequestData = requestWithState.getRequest();
```
tpetr: naming seems awkward here since it's not always a new request

ssalinas:
@tpetr updated the field names and such here. Took another look and I didn't see any more places where we would get conflicts between the two sources of request data.
- deploy checker -> uses the new request data if it is set when checking the deploy
- scheduler -> uses the pending deploy request data if it is set and if the pending request has a matching deploy id
- cleaner -> relevant cases only deal with the active deploy, so it always should use the current request data not pending deploy one

Would like to get @wsorenson 's thoughts as well to make sure I'm not missing anything.

tpetr:
What are your thoughts on moving the `updatedRequest` field from `SingularityDeployParent` to `SingularityDeploy`? I'm trying to debug a possible issue with this PR and it's a pain to not be able to get the updated request data over the API.

ssalinas:
My initial concern was that we then have possibly stale request data that hangs around and isn't correct. The deploy data is saved in history, but if the request is updated, we have out of date data hanging around. The deploy parent data is only around during the time the deploy is actually running as part of the pending deploy.

ssalinas:
@tpetr after the update we made back in April this has been stable, good to merge?

tpetr:
LGTM

PR #975, opened 2016-03-28
Remove cleanup after bounce expire

### LINKED ISSUE #972: Task bounce expires but no new task scheduled
Today we experienced a task which was commanded to bounce, but the new task was unable to launch for some reason.  This led to a state where the existing task is in the `CLEANING` state waiting for a new task to come up, but there was no scheduled task ever to be seen.  Seen on 0.4.11

[Singularity logs for the task](https://gist.github.com/stevenschlansker/ef9ef4f29ce54e0e51d6)

screenshot_2016-03-24_16-18-56
screenshot_2016-03-24_16-19-17

ssalinas:
@tpetr initially did this so that it /fixes #972 . This will:
- check that no other task cleanups are present (i.e. if task should still be in cleaning state)
- if not, find the task history updates and if the last one is `TASK_CLEANING` delete it
- re-fire an event for the previous state the task was in (most likely running)
- make sure a pending request is put in the queue to take care of any other pending tasks or launched tasks that might need to be taken care of

In addition I also did the same thing for the case where a decommissioned slave is reactivated (handled by SingularityCleaner), since right now reactivating the slave still requires all of the cleaning tasks to be properly cleaned and repalced, even though their slave is no longer deocmmissioning.

Thoughts on re-firing an event for the previous state vs creating a new event type for history updates that are deleted?


ssalinas:
@tpetr are you cool with merging this?

ssalinas:
Found one thing here, seems that we can accidentally reuse instance number in this case. IF a task is cleaning we will currently reuse that instance number. If it then comes back out of cleaning, we will have a duplicate instance number. Not sure if that's an issue or not, will revisit this

ssalinas:
updated to not worry about the decommission/reactivation for now, causes more headaches than it's worth at the moment. Bounce expire was the real issue to start with anyways.

ssalinas:
@tpetr are you good with merging this one? causes a rather annoying problem for anyone without the change when canceling a bounce

tpetr:
LGTM

PR #977, opened 2016-03-29
Run now dialog improvements

Calvinp:
This makes big changes to the run now dialog UX, and even bigger changes behind the scenes to enable this.
UX improvements
-
- Fix a bug in which the run now button didn't work when clicked from the pending tasks page
- The checkbox to automatically tail logs has been converted into radio buttons. The options are stay on the request's page (previous default behavior), tail logs, or browse to the task's page (new default option and what will be selected for you the first time you refresh after this is merged)
- These radio buttons will remember your previous choice and default you to that choice
- If you don't have automatically tail logs selected, the input box to specify a file to tail will be disabled
- The input box to specify a file to tail will have 'service.log' filled in for you. If you write in something else, it will from then on fill that in for you instead. It uses local storage for this.
- If you're automatically tailing logs, and the poller detects that the task you're polling is in a terminal state but the log you're looking for doesn't exist yet, it will give a more descriptive message in the popup.
- Fix a bug in which the global constant defining the task poller timeout was completely ignored, and the timeout was instead set to one minute. But the user was told that the constant (5 minutes) was used, so it appeared to the user as if they had somehow lost 4 minutes of their life watching the poller.
- Fix a bug in which the command line args weren't populated from localStorage
- Change the command line args to populate from the last task run that request instead of localStorage
- screenshot 2016-03-29 14 20 50

Behind the scenes UI changes
-
- The code for the run now dialog was a huge mess riddled with bugs. This makes most of that better.
- The promptRerun function in the request model has been vanquished. Instead, promptRun takes an optional task parameter, and if that parameter is specified it reruns the given task.
- The logic to start the poller to automatically tail logs is now implemented by promptRun. It used to be that the caller would need to open the autotailer, and several callers didn't. So the autotailer wouldn't work if the task run button was pressed in certain places.
- The poller to automatically detect when a task exists and its files exist totail logs has been enhanced. Now it takes a pollingType variable. If this type is 'autoTail', it does what it used to do. If it's 'browse-to-sandbox', it will instead direct you to a task's page when the task starts. It has been renamed from AutoTailer to TaskPoller, which is a much more accurate name for what it does. Several of its supporting templates have also been renamed. Many internal methods and variables have also been renamed. Unfortunately, with all the renamed variables, the diff is so different that Git wasn't able to detect that this is a rename - instead believing that these are all new files.
- The poller also now uses runId to track a task rather than watching for a reset event, which sometimes missed certain tasks.

Backend change
-
- @ssalinas added an API for retrieving running tasks by run id.


SingularityUI/app/models/Request.coffee
```diff
@@ -427,18 +431,22 @@ class Request extends Model
...
+        else
+            try
+                lastCommands = JSON.parse(localStorage.getItem(@localStorageCommandLineInputKeyPrefix + @id))
+            catch e
+                console.error('Could not parse previous commands JSON')
+                lastCommands = []
+            commands = if lastCommands? then lastCommands[lastCommands.length - 1] else []
         vex.dialog.prompt
-            message: "<h3>Run Task</h3>"
+            message: "<h3>#{if task then 'Rer' else 'R'}un Task</h3>"
```
zdhickman: nothing wrong with this logically, but for the sake of clarity in the code, maybe:

`#{if task then 'Rerun' else 'Run'} Task`

SingularityUI/app/models/Request.coffee
```diff
@@ -466,17 +474,48 @@ class Request extends Model
...
+                        callback( @data )
+
+                    @run( @data.commandLineInput, message, @data.runId ).done doneFn
                     return true
 
             afterOpen: =>
-                $('#filename').val localStorage.getItem('taskRunRedirectFilename')
-                $('#autoTail').prop 'checked', (localStorage.getItem('taskRunAutoTail') is 'on')
+                taskRunAfterStart = localStorage.getItem('taskRunAfterStart')
+                $('#filename').val localStorage.getItem('taskRunRedirectFilename') or 'service.log'
```
ssalinas: service.log is a very internal thing for us, if not using our executor the log file will be stdout or some other file. I believe we have this already somewhere, but there should be a configurable default log file we can put here instead

Calvinp: It remembers your previous choice and puts it in there, so it might be ok to not actually have any default.

SingularityUI/app/templates/vex/requestRun.hbs
```diff
@@ -24,9 +24,13 @@
     <input name="message" id="run-message" type="text" />
 </div>
 <hr class='border-color-light-gray'>
-
-<p><label class='label-light'><input id="autoTail" name="autoTail" type="checkbox"> Wait for task to start, then start tailing:</label></p>
+<strong>After triggering the run:</strong>
+<p>
+    <label class='label-light'><input id="stay-on-request-page" name="afterStart" type="radio" value="stay-on-request-page"> Stay on this request's page</label>
+    <label class='label-light'><input id="browse-to-sandbox" name="afterStart" type="radio" value="browse-to-sandbox"> Wait for task to start, then browse to its sandbox</label>
```
ssalinas: 'browse to its sandbox' is awkward wording. Maybe take the 'to' out of that

SingularityUI/app/models/Request.coffee
Calvinp: Oh, I got that from the placeholder text, which is currently 'e.g. service.log'. But that's a constant, not configurable atm.

Calvinp: Ok, the configuration exists and is now being used.

SingularityUI/app/templates/vex/requestRun.hbs
```diff
@@ -24,9 +24,13 @@
     <input name="message" id="run-message" type="text" />
 </div>
 <hr class='border-color-light-gray'>
-
-<p><label class='label-light'><input id="autoTail" name="autoTail" type="checkbox"> Wait for task to start, then start tailing:</label></p>
+<strong>After triggering the run:</strong>
+<p>
+    <label class='label-light'><input id="stay-on-request-page" name="afterStart" type="radio" value="stay-on-request-page"> Stay on this request's page</label>
```
Calvinp: I'm going to take out the word "request's" here because the dialog can be accessed from pages other than just a request page.

SingularityUI/app/models/Request.coffee
```diff
@@ -427,69 +434,120 @@ class Request extends Model
...
+        showDialog = () =>
+            if task
+                commands = task.attributes.task.taskRequest.pendingTask.cmdLineArgsList
+            else
+                commands = @commands
+            vex.dialog.prompt
+                message: "<h3>#{if task then 'Rerun' else 'Run'} Task</h3>"
+                input: runTemplate
+                    id: @get "id"
+                    prefix: @localStorageCommandLineInputKeyPrefix
```
ssalinas: do we still need this prefix if we aren't using localStorage for populating args anymore?

Calvinp: Nope! It's gone.

SingularityUI/app/views/request.coffee
```diff
@@ -78,38 +77,22 @@ class RequestView extends View
...
-
-            else
+    runRequest: (e, taskId) => # If taskId is provided, rerun the task. Else run the task.
+        callback = (data) =>
+            unless data.afterStart in ['browse-to-sandbox', 'autoTail']
                 @trigger 'refreshrequest'
                 setTimeout ( => @trigger 'refreshrequest'), 2500
+        if taskId
+            task = new TaskHistory {taskId}
+            task.fetch()
```
ssalinas: we already have the task history fetched on this page, is there not a way we can use the data we already have?

SingularityUI/app/views/TaskPoller.coffee
```diff
@@ -0,0 +1,133 @@
...
+
+
+    # Start polling for task changes, and check
+    # Task History changes in case we need
+    # to back out of the file redirect
+    startTaskPolling: ->
+        @showTaskPollWaitingDialog()
+        @stopTaskPolling()
+
+        @listenTo @history, 'reset', @handleHistoryReset
```
ssalinas: Now that we are assigning a runId to everything, there is in fact an endpoint to get a task by runId. We can probably use that to be a bit more exact/not get data we aren't going to use

SingularityUI/app/views/request.coffee
Calvinp: Unfortunately, that's not inherently true. While it is true if you're running from the request's page, you may also be running the task from the tasks page or the requests page.

SingularityUI/app/models/TaskHistoryItem.coffee
```diff
@@ -6,7 +6,19 @@ class TaskHistoryItem extends Model
 
     ignoreAttributes: ['id', 'canBeRunNow']
 
-    initialize: ({ @taskId, @updatedAt, @lastTaskState }) ->
+    url: () ->
+        if @requestId and @runId
+            "#{ config.apiRoot }/history/request/#{ @requestId }/run/#{ @runId }"
+        else
+            # Currently the above URL is the ONLY place to fetch this model.
+            # If you don't have access to request ID and run ID use the TaskHistory
```
ssalinas: once we have the taskId, we can also fetch this using that. I know the main use case with be by runId, but just wanted to clarify

Calvinp: When you fetch by taskId, you get a different model as a response. That gives back a SingularityTaskHistory (which is the TaskHistory model in the UI), wheras searching by runId returns a SingularityTaskIdHistory (Which is this model in the UI). In a roundabout way, you can use Task Search to search out this model by taskId, but then you'd be fetching the collection rather than the model.

screenshot 2016-03-31 09 07 55
screenshot 2016-03-31 09 10 06

SingularityUI/app/views/TaskPoller.coffee
```diff
@@ -0,0 +1,142 @@
...
+taskPollingFailureTemplate = require 'templates/vex/taskPollingFailure'
+
+Utils = require '../utils'
+
+vex = require 'vex.dialog'
+
+interval = (a, b) -> setInterval(b, a)  # f u javascript
+timeout = (a, b) -> setTimeout(b, a)
+
+TIMEOUT_MINUTES = 1 # Modify this
```
ssalinas: Might be easier to get this to millis using something like `moment.duration(1, 'minutes').asMilliseconds()`, and gets rid of the need for using comments like 'Don't modify this'

PR #979, opened 2016-04-01
Email find first error in logs

Calvinp:
This improves upon the previously existing logic to show the bottom few lines of a file when sending the task complete email.
Now, instead, the _first_ error will be searched out. What is shown is the number of bytes we are allowed to show starting at where the error is found.
It is not necessarily easy to define what an error is, so this choice is left up to the user using a tiered system of regexes:
- If the user is lazy and doesn't specify anything anywhere, a default regex of "ERROR" is used. I'm open to discussion on exactly what this default regex should be.
- The user may override this default by providing a regex named taskLogErrorRegex in the smtp config.
- The user may override the smtp config in an individual request (very useful if that request will be running in a different language than most, with different logging conventions - but also useful for any number of other reasons that I can't even guess at)

If no error is found in the logs (which would happen if the task is killed by singularity, for example, or if it completed successfully) then the bottom of the log file is shown instead.

The motivation for choosing "ERROR" as the default regex is that in many common Java logging schemes, logging at the error level will produce a line starting with "ERROR". It is probably best to keep the regex as simple as possible to avoid situations where the word error or exception is logged for some expected reason (eg. "Set up exception handling"), then found instead of an actual error. Therefore using "ERROR" will catch as many errors as possible without catching too many things that aren't errors.


SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -142,21 +158,42 @@ public MailTemplateHelpers(SandboxManager sandboxManager, SingularityConfigurati
     }
   }
 
-  private Optional<Long> getMaybeOffset(String slaveHostname, String fullPath, String filename, Optional<String> directory) {
+  // Searches through the file,
+  private Optional<Long> getMaybeOffset(final String slaveHostname, final String fullPath, final Long logLength, Optional<SingularityTask> task) {
+    long offset = 0;
+    String regex = getLogErrorRegex(task);
+    long length = logLength + regex.length(); // Get extra so that we can be sure to find the error
+    Pattern pattern = Pattern.compile(regex);
```
ssalinas: What happens if an invalid pattern is supplied? We should probably catch this and default to returning the tail of the log file (like old behavior)

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -34,6 +36,7 @@
   private static final String REQUEST_LINK_FORMAT = "%s/request/%s";
   private static final String LOG_LINK_FORMAT = "%s/task/%s/tail/%s";
   private static final String DEFAULT_TIMESTAMP_FORMAT = "MMM dd h:mm:ss a zzz";
+  private static final String DEFAULT_TASK_LOG_ERROR_REGEX = "ERROR";
```
ssalinas: Going back and forth on if the default should be to search for errors in the logs, or if that's something the operator should need to actively configure. @tpetr any thoughts there on a sensible default behavior?

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -142,21 +158,42 @@ public MailTemplateHelpers(SandboxManager sandboxManager, SingularityConfigurati
...
-      return Optional.absent();
-    }
-    if (logChunkObject.isPresent()) {
-      return Optional.of(logChunkObject.get().getOffset());
-    } else {
-      LOG.error("Failed to get offset for log file {}", fullPath);
-      return Optional.absent();
+    Optional<MesosFileChunkObject> previous = Optional.absent();
+
+    while (true) { //Continue until we reach the end of the file, at which point we return from within the while loop
```
ssalinas: Two things for this loop:
- In the case that the file is massive (for example, a wrongly configured service running for a few hours can spew GBs upon GBs of logs into a file), should we limit the amount of log file we will search?
- `while true`  always makes me a bit uneasy that we could be stuck here forever. Maybe we should check the ending offset of the log before starting the loop, so that even if something goes wrong here, we can still have a defined end and no chance of ever getting stuck here?

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -142,21 +158,42 @@ public MailTemplateHelpers(SandboxManager sandboxManager, SingularityConfigurati
...
+        LOG.error("Sandboxmanager failed to read {} on slave {}", fullPath, slaveHostname, e);
+        return Optional.absent();
+      }
+      if (logChunkObject.isPresent()) {
+        if (logChunkObject.get().getData().equals("")) { // Passed end of file
+          if (previous.isPresent()) { // If there was any log, get the bottom bytes of it
+            long end = previous.get().getOffset() + previous.get().getData().length();
+            return Optional.of(end - logLength);
+          }
+          return Optional.absent();
```
ssalinas: returning an absent here would revert us to the previous behavior of returning the beginning portion of the file (absent offset). If we can't find anything, wouldn't we still want to see the tail like we implemented in the previous PR?

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -142,21 +158,42 @@ public MailTemplateHelpers(SandboxManager sandboxManager, SingularityConfigurati
     }
   }
 
-  private Optional<Long> getMaybeOffset(String slaveHostname, String fullPath, String filename, Optional<String> directory) {
+  // Searches through the file,
+  private Optional<Long> getMaybeOffset(final String slaveHostname, final String fullPath, final Long logLength, Optional<SingularityTask> task) {
+    long offset = 0;
+    String regex = getLogErrorRegex(task);
+    long length = logLength + regex.length(); // Get extra so that we can be sure to find the error
```
ssalinas: if we are looping through the file why do we need this extra length?

Calvinp: If we fetch just the amount we're supposed to, then it's possible that the match will be half in one fetched chunk and half in the next, in which case we wouldn't match it.
For example, we're looking for 'ERROR'. If we fetch 10 bytes, and the log file is 'xxxxxxxxERRORxxxxxxx', we would fetch 'xxxxxxxxER', not match, then fetch 'RORxxxxxxx', and still not match. Whereas with the extra length we'd fetch 'xxxxxxxxERRORxx' and correctly find the match.

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
Calvinp: Unless I missed something, this absent will only be returned if the log file is empty. Otherwise previous will be present and it'll return end-logLength, which will be a pointer to the end of the file.

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -34,6 +37,8 @@
   private static final String REQUEST_LINK_FORMAT = "%s/request/%s";
   private static final String LOG_LINK_FORMAT = "%s/task/%s/tail/%s";
   private static final String DEFAULT_TIMESTAMP_FORMAT = "MMM dd h:mm:ss a zzz";
+  private static final Optional<String> DEFAULT_TASK_LOG_ERROR_REGEX = Optional.of("ERROR");
+  private static final Long MAX_LOG_SEARCH_OFFSET = 1000000L;
```
Calvinp: I am very open to ideas on the best value for this constant (the maximum number of bytes to search through for an error before giving up and showing the bottom of the file) as well. 1,000,000 bytes was completely arbitrary and isn't necessarily the best value. We can put this into the smtp config too if we would like, but we'll still need to choose a default.

SingularityService/src/main/java/com/hubspot/singularity/config/SMTPConfiguration.java
```diff
@@ -70,6 +70,10 @@
   @JsonProperty
   private List<String> taskEmailTailFiles = Arrays.asList("stdout", "stderr");
 
+  @NotNull
+  @JsonProperty
+  private String taskLogErrorRegex = "";
```
tpetr: we should make this an `Optional<String>` (i.e. make this functionality opt-in)

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
tpetr: Yep, this should be configured in `SMTPConfiguration`. 100kb is a good default.

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -34,6 +37,8 @@
   private static final String REQUEST_LINK_FORMAT = "%s/request/%s";
   private static final String LOG_LINK_FORMAT = "%s/task/%s/tail/%s";
   private static final String DEFAULT_TIMESTAMP_FORMAT = "MMM dd h:mm:ss a zzz";
+  private static final Optional<String> DEFAULT_TASK_LOG_ERROR_REGEX = Optional.of("ERROR");
```
tpetr: Any defaults for configuration items should live in the configuration class (i.e. https://github.com/HubSpot/Singularity/pull/979/files#diff-c39ca42e7478ecb6adcf0bf05ac23373R75)

SingularityService/src/main/java/com/hubspot/singularity/smtp/MailTemplateHelpers.java
```diff
@@ -142,16 +160,69 @@ public MailTemplateHelpers(SandboxManager sandboxManager, SingularityConfigurati
     }
   }
 
-  private Optional<Long> getMaybeOffset(String slaveHostname, String fullPath, String filename, Optional<String> directory) {
+  // Searches through the file, looking for first error
+  private Optional<Long> getMaybeOffset(final String slaveHostname, final String fullPath, final Long logLength, Optional<SingularityTask> task) {
```
tpetr: can you give this method a clearer name? what exactly does it do?

PR #981, opened 2016-04-05
Improvements to Task Overview on Task page

Calvinp:
- **Removes taskId header on task page**
- Moves buttons (JSON, Kill task) to be in line with the breadcrumbs containing requestId and deployId
- Adds instance number to those breadcrumbs. 
- Adds hostName to taskOverview well. 
- Changes task status in taskOverview well into a label. 
- Puts exit reason on the taskOverview well on its own line


ssalinas:
Since it is no longer up top, can we get the rackId in another box down on the info section of the page? Unescaped version should be available in task.RackId . Other than that, looks good

ssalinas:
Thanks for working with me on this @Calvinp , new task header section looks good

PR #986, opened 2016-04-06
Fix and Rename getQuartzScheduleSafe() Method

Calvinp:
This refactors `getQuartzScheduleSafe()`. It is now named `getQuartzScheduleOrSchedule()`, and returns an optional instead of a string.
Uses of this method are updated to be able to handle an optional instead of a string.

If there are any potential side effects of doing this that I didn't think of, discuss here.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
@@ -112,10 +112,20 @@ public class SingularityScheduledJobPoller extends SingularityLeaderOnlyPoller {
 
       final CronExpression cronExpression;
 
+      Optional<String> maybeSchedule = request.getRequest().getQuartzScheduleOrSchedule();
+      if (!maybeSchedule.isPresent()) {
+        String msg = String.format("Cron doesn't exist for {}", taskId.toString());
+        LOG.warn(msg);
+        exceptionNotifier.notify(msg, ImmutableMap.of("taskId", taskId.toString()));
+        return Optional.absent();
+      }
+
+      String schedule = maybeSchedule.get();
+
       try {
-        cronExpression = new CronExpression(request.getRequest().getQuartzScheduleSafe());
+        cronExpression = new CronExpression(schedule);
       } catch (ParseException e) {
-        LOG.warn("Unable to parse cron for {} ({})", taskId, request.getRequest().getQuartzScheduleSafe(), e);
+        LOG.warn("Unable to parse cron for {} ({})", taskId, schedule, e);
         exceptionNotifier.notify(e, ImmutableMap.of("taskId", taskId.toString()));
         return Optional.absent();
       }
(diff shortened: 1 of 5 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduledJobPoller.java
```diff
@@ -112,10 +112,20 @@ public void runActionOnPoll() {
 
       final CronExpression cronExpression;
 
+      Optional<String> maybeSchedule = request.getRequest().getQuartzScheduleOrSchedule();
+      if (!maybeSchedule.isPresent()) {
```
wsorenson: Is this possible? Isn't bad state like this the point of the validator?

Calvinp: That shouldn't be possible in this spot, but we just encountered (and, to be fair, fixed) a bug where the validator was relying on the method before checking if either schedule was there. So maybe that should raise an exception if the schedule isn't present and log an error (rather than a warning), because the only way it'll be the case is if something went wrong.

ssalinas: It is. This PR was prompted by something that was missing fro the validator causing errors later on. I guess the question is if we want to let those errors happen and fix the validator, or catch them like this.

@Calvinp still not sure this refactor is worth it to just get the 'safe' in the method name correct. Maybe let's come back to this one another time?

Calvinp: Maybe. It just makes me extremely uncomfortable having the word "safe" in the name of a method that is known not to be safe...

tpetr: Yeah I think we need to approach this differently -- the `getQuartzScheduleSafe()` method isn't great because it falls back to the cron schedule if no quartz schedule is set, which doesn't make much sense because the two syntaxes are incompatible with each other. I think the original PR to fix the validator is good enough for now.

wsorenson: Hmm, the entire point of the safe() method is to return a known non-null value. The fact that it's possible to return a null value from there indicates a broken assumption elsewhere that I think we should fix. Otherwise, we can't have nice things anywhere and have to constantly question all non-optional returns.

Calvinp: That's also a fair point. I'll open a new issue for this and we can figure out in that issue what to do about it.

ssalinas: That broken assumption was fixed in #985

Calvinp: True, but calling `getQuartzScheduleSafe()` is still not safe. If you call it on a request that isn't a scheduled request it'll break in the same way.

Calvinp: Deleted the other comment because it was simulposted with one from @wsorenson
I am happy to close this PR and move discussion somewhere else if everyone else is satisfied for now, but it doesn't appear that is the case.

Calvinp:
We're going to look more into this later. For now, closing this PR.

PR #991, opened 2016-04-07
No user isn't an admin user anymore

Calvinp:
The admin check would give you admin rights if you were logged in as no user.
This fixes that.

@tpetr please take a look - it is possible that there is a good reason for what Singularity was doing that I don't know about.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
@@ -69,8 +69,8 @@ public class SingularityAuthorizationHelper {
 
   public void checkAdminAuthorization(Optional<SingularityUser> user) {
     if (authEnabled) {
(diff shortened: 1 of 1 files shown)


ssalinas:
:+1: I can't think of a reason where having no user should allow you any admin actions

tpetr:
LGTM, thanks. Let's be sure to test out the [endpoints](https://github.com/HubSpot/Singularity/search?utf8=%E2%9C%93&q=checkAdminAuthorization) that use this method in the test cluster to make sure there aren't any side effects.

tpetr:
On second thought, Unauthorized is the more correct response to return if the user is not present: http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses

I'd suggest copying what we do elsewhere and use the `checkUnauthorized()` method before `checkForbidden()`

tpetr:
:balloon:

ssalinas:
Thanks for the fix @Calvinp

PR #992, opened 2016-04-08
Restrict Admin Only Pages To Actually Be Admin Only

Calvinp:
This exposes to the UI the ability to check if the user is an admin.
The UI uses this to prevent the user from visiting the pages that are meant to only be available to admins.

### DIFF
--- a/SingularityUI/app/application.coffee
+++ b/SingularityUI/app/application.coffee
@@ -179,7 +179,7 @@ class Application
 
     setupUser: (callback) ->
         @user = new User
-        @user.fetch().done callback # This is not synchronous and does not use local storage
+        @user.fetch().done callback
 
     hasAdminRights: ->
         return false unless @user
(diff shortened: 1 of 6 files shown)


SingularityUI/app/application.coffee
```diff
@@ -23,45 +23,44 @@ class Application
     initialize: ->
         @setupGlobalErrorHandling()
 
-        @setupUser()
+        @setupUser =>
```
Calvinp: The changes here are just changes in indentation - rendering of pages will be delayed until we tried to fetch the user so that we can avoid awkward situations where the admin dropdown in the toolbar disappears or suddenly appears on the second loaded page.

SingularityUI/app/application.coffee
```diff
@@ -178,9 +177,13 @@ class Application
         else
             @page.appendChild view.el
 
-    setupUser: ->
+    setupUser: (callback) ->
         @user = new User
-        @user.fetch() # Syncronous because it uses localStorage
```
Calvinp: Lies.

ssalinas:
@Calvinp going to close this in favor of #1009 , I think that implementation will be better

PR #994, opened 2016-04-11
Upgrade Mesos Version

ssalinas:
Currently built against `0.27.2`, will bump to `0.28.1` if it is ready before we are ready to release.

Major changes:
- `.json` endpoints no longer have the `.json`, however the deprecated versions of the endpoints are still in place, so no worries around the upgrade process
- `TaskStatus.Reason` enum updated for memory limit
- Bumped java version in docker image
- Installing docker in slave image via apt
- credential secret is now a String not a byte array


ssalinas:
Bumped to 0.28.1 now that it has been released. Will not build currently beccause I am still investigating how closely mesos' new `TASK_KILLING` state maps to our `TASK_CLEANING` and how to handle the new one

ssalinas:
So, it seems `TASK_KILLING` is triggered by the executor when it has received a request to kill and started the process of the kill. It doesn't quite map to our `TASK_CLEANING` state. However, reading through the PR comments, it seems that it's up to each executor to decide whether or not to send it, and up to each framework to decide whether or not to do anything with it. So, we can safely support but ignore usage of it for now

PR #996, opened 2016-04-14
deployProgress was undefined yet we tried to get attributes out of it

Calvinp:
This fixes a small bug that was triggered when looking at a request page within a few seconds of adding a new deploy to that request.
The deployProgress attribute of that deploy is not yet defined, yet Singularity tried to get timestamp and deployStepWaitTimeMs out of it, causing an error to be logged to the console.


ssalinas:
:+1: thanks for the fix @Calvinp

ssalinas:
LGTM

PR #997, opened 2016-04-14
Webhook API changes and UI Implementation

Calvinp:
On the API side this adds a summary method to obtain the summaries of all the webhooks with the size of their queue.
On the UI side this implements the UI for the webhooks page in React. 
It also makes many (though not all) of the shared react components easier to reuse:
- propTypes were added to many react components
- Messy code was cleaned up
- Non-custom sorting was implemented in tables (but you still have to provide your own sort function). Note that elements will **not** jump around even if there are more than ten.
- Vex now works with React.

Things that could be improved (but may not be necessary for now):
- There is a bit of messy code. Places where I could have used shared react components but didn't, or put methods into React that maybe should have been part of the model
- Vex + React is a bit of a hack, and implementation could be improved by making the Vex components reusable.


SingularityService/src/main/java/com/hubspot/singularity/resources/WebhookResource.java
```diff
@@ -50,6 +52,24 @@ public WebhookResource(WebhookManager webhookManager, SingularityAuthorizationHe
     return webhookManager.getActiveWebhooks();
   }
 
+  @GET
+  @Path("/summary")
+  @ApiOperation("Retrieve a summary of each active webhook")
+  public List<SingularityWebhookSummary> getActiveWebhookSummaries() {
+    authorizationHelper.checkAdminAuthorization(user);
+    List<SingularityWebhookSummary> webhookSummaries = new ArrayList<SingularityWebhookSummary>();
```
ssalinas: Two things here. First, let's move this into the WebhookManager class, so we can manage building everything there. Second, there are a few methods available in the WebhookManager class that should allow us to do this more efficiently (batch up the zookeeper calls). Using something like the `getAsyncChildren` calls could help here.

Calvinp:
Note that this PR (including its title and top comment) has been updated to include the UI implementation.

Calvinp:
Two bugs:
- First, I managed to create a webhook with a space in the URI and now I can't delete it
- Second, it isn't sending the user creating the webhook.
I'll address these tomorrow.

Calvinp:
Those two bugs are now fixed.

Having the ID in a path param turned out to be problematic because certain browsers and proxies will encode and decode the URL the wrong number of times, so the old API methods that used path params were deprecated. A new set of API methods was added that does exactly the same things as the old ones, except using query params instead of path params - since proxies and browsers leave query params alone.

SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
```diff
@@ -114,10 +114,10 @@
...
 
   private static final String WEBHOOKS_FORMAT = "http://%s/%s/webhooks";
-  private static final String WEBHOOKS_DELETE_FORMAT = WEBHOOKS_FORMAT +"/%s";
-  private static final String WEBHOOKS_GET_QUEUED_DEPLOY_UPDATES_FORMAT = WEBHOOKS_FORMAT + "/deploy/%s";
-  private static final String WEBHOOKS_GET_QUEUED_REQUEST_UPDATES_FORMAT = WEBHOOKS_FORMAT + "/request/%s";
-  private static final String WEBHOOKS_GET_QUEUED_TASK_UPDATES_FORMAT = WEBHOOKS_FORMAT + "/task/%s";
+  private static final String WEBHOOKS_DELETE_FORMAT = WEBHOOKS_FORMAT +"?webhookId=%s";
+  private static final String WEBHOOKS_GET_QUEUED_DEPLOY_UPDATES_FORMAT = WEBHOOKS_FORMAT + "/deploy?webhookId=%s";
+  private static final String WEBHOOKS_GET_QUEUED_REQUEST_UPDATES_FORMAT = WEBHOOKS_FORMAT + "/request?webhookId=%s";
+  private static final String WEBHOOKS_GET_QUEUED_TASK_UPDATES_FORMAT = WEBHOOKS_FORMAT + "/task?webhookId=%s";
```
ssalinas: It would be much cleaner to build the query params as part of the request object rather than tacking them on as part of the url. See `readSandBoxFile` in the client for an example of the query param builder, then you can use the `getCollectionWithParams` method inside `getQueued*` for the webhooks.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -166,6 +170,19 @@ public SingularityRequest checkSingularityRequest(SingularityRequest request, Op
...
   }
 
+  public SingularityWebhook checkSingularityWebhook(SingularityWebhook webhook) {
+    checkNotNull(webhook, "Webhook is null");
+    checkNotNull(webhook.getUri(), "URI is null");
+
+    try {
+      new URI(webhook.getUri());
+    } catch (URISyntaxException e) {
+      checkBadRequest(false, "Invalid URI provided");
```
ssalinas: Might as well do `WebExceptions.badRequest("Invalid URI provided")` here for clarity, since the condition argument is always false

SingularityService/src/main/java/com/hubspot/singularity/resources/WebhookResource.java
```diff
@@ -1,13 +1,16 @@
 package com.hubspot.singularity.resources;
 
+import java.util.ArrayList;
```
ssalinas: unused imports?

SingularityUI/app/components/webhooks/Webhooks.cjsx
```diff
@@ -0,0 +1,215 @@
...
+Link = require '../common/atomicDisplayItems/Link'
+Glyphicon = require '../common/atomicDisplayItems/Glyphicon'
+NewWebhookForm = require './NewWebhookForm'
+vex = require 'vex'
+
+Webhooks = React.createClass
+
+    defaultRowsPerPage: 10
+
+    rowsPerPageChoices: [10, 20, 30, 40]
```
ssalinas: webhooks is generally going to be a very small set of items, I'd be surprised if any users even had enough to fill a page of 10. Maybe just limit this to 10, 20 so we aren't presenting options with way more than we need?

SingularityUI/app/components/webhooks/Webhooks.cjsx
```diff
@@ -0,0 +1,215 @@
...
+                )
+            beforeClose: =>
+                return true unless @data
+                @errors = []
+                uriValidated = @validateInput @uri
+                @errors.push 'Please select a type' unless @type
+                @errors.push 'Invalid URL entered' unless uriValidated
+                @renderedForm.forceUpdate() unless uriValidated and @type
+                return false unless uriValidated
+                return false unless @type
```
ssalinas: In the ui it seems that this defaults to task if you don't click one of the type options, but none of the buttons look selected by default. Let's make sure if we are going to have a default that the button also looks selected when the dialog opens

Calvinp: I looked over the code and went through the UI, but I couldn't find a default. If you don't select a type the request isn't sent, the dialog stays open, and the UI prompts you to select a type.
On a previous iteration I had a default, but I removed that. It's possible that there are artifacts left over from that somewhere, but I can't find them at the moment. Where are you seeing a default? I'll remove it if I can find it.

ssalinas: realized I must have had an older version of code loaded, this should be all set ^

SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
Calvinp: @ssalinas is the current what you meant?

ssalinas:
Looks good @Calvinp , going to merge so we can work more on the PRs that build off these components

PR #998, opened 2016-04-15
Fix to enable buttons on dashboard paused requests table

Calvinp:
The required functionality was not implemented.

### DIFF
--- a/SingularityUI/app/views/dashboard.coffee
+++ b/SingularityUI/app/views/dashboard.coffee
@@ -11,6 +11,9 @@ class DashboardView extends View
             'click [data-action="unstar"]': 'unstar'
             'click [data-action="change-user"]': 'changeUser'
             'click th[data-sort-attribute]': 'sortTable'
+            'click [data-action="viewJSON"]': 'viewJson'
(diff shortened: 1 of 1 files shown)


SingularityUI/app/views/dashboard.coffee
```diff
@@ -119,4 +122,43 @@ class DashboardView extends View
...
+        if maybeRequest
+            return maybeRequest[0]
+        else
+            return
+
+    viewJson: (e) ->
+        id = $(e.target).parents('tr').data 'request-id'
+        request = @getRequest id
+        unless request
+            console.error ("Could not find request #{id}. Perhaps someone removed it?")
```
ssalinas: A console.error would hide this error from anyone without the console open. To most it would look like it just wasn't working and they might keep clicking. Let's keep this message, but use something like `Messenger().error` instead (see application.coffee for an example). This will give a little dispersible message telling the user went wrong.

Calvinp: Fixed

ssalinas:
:+1:

ssalinas:
Thanks for the fix @Calvinp

PR #1000, opened 2016-04-15
Tweaks to s3UploaderAdditionalFiles

ssalinas:
- Make the `s3UploaderAdditionalFiles` paths relative to the task directory, not the logrotate path
- Allow the additional files to specify a glob

This should make things much more flexible and allow us to have something like an upload folder or other globs that match a wider range of things

/cc @tpetr


ssalinas:
Another that has been working well, going to merge. Quick note that we'll need to add a short note in the next release on the change in configuration (ability to add a directory relative to task folder, default directory is the task folder, not logs now)

Issue #1001, opened 2016-04-19
Singularity aborts when sending overdue task emails if smtp not configured


dtabwhite:
I encountered a situation where I has some scheduled requests that got stuck in a PENDING state.  I'm not entirely sure how they got into that state but I think it was related to pausing/unpausing them repeatedly.  Once these pending requests existed, I began seeing singularity abort periodically.  I tracked it down to this in the log:

```
2016-04-19T11:45:52.54317 ERROR [2016-04-19 11:45:52,537] com.hubspot.singularity.scheduler.SingularityLeaderOnlyPoller: Caught an exception while running SingularityScheduledJobPoller
2016-04-19T11:45:52.54320 ! java.lang.IllegalStateException: Optional.get() cannot be called on an absent value
2016-04-19T11:45:52.54321 ! at com.google.common.base.Absent.get(Absent.java:47) ~[SingularityService-shaded.jar:0.5.0]
2016-04-19T11:45:52.54321 ! at com.hubspot.singularity.smtp.SingularityMailer.getDestination(SingularityMailer.java:317) ~[SingularityService-shaded.jar:0.5.0]
2016-04-19T11:45:52.54322 ! at com.hubspot.singularity.smtp.SingularityMailer.prepareTaskMail(SingularityMailer.java:275) ~[SingularityService-shaded.jar:0.5.0]
2016-04-19T11:45:52.54324 ! at com.hubspot.singularity.smtp.SingularityMailer.sendTaskOverdueMail(SingularityMailer.java:249) ~[SingularityService-shaded.jar:0.5.0]
2016-04-19T11:45:52.54325 ! at com.hubspot.singularity.scheduler.SingularityScheduledJobPoller.runActionOnPoll(SingularityScheduledJobPoller.java:94) ~[SingularityService-shaded.jar:0.5.0]
2016-04-19T11:45:52.54325 ! at com.hubspot.singularity.scheduler.SingularityLeaderOnlyPoller.runActionIfLeaderAndMesosIsRunning(SingularityLeaderOnlyPoller.java:108) [SingularityService-shaded.jar:0.5.0]
2016-04-19T11:45:52.54326 ! at com.hubspot.singularity.scheduler.SingularityLeaderOnlyPoller.access$000(SingularityLeaderOnlyPoller.java:24) [SingularityService-shaded.jar:0.5.0]
[...]

PR #1003, opened 2016-04-19
Check for optional.isPresent() before calling any public mailer methods

Calvinp:
Fixes #1001 by checking if smtpConfig is absent before allowing any public smtp methods to be called.

PR #1004, opened 2016-04-19
Convert SingularityMailer into an interface

jhaber:
Alternative to #1003 converts `SingularityMailer` into an interface and binds a no-op impl if the smtp configuration isn't present


ssalinas:
+1 thanks haber

Calvinp:
+1 Much cleaner

ssalinas:
Thanks again for the fix haber, /fixes #1001

PR #1009, opened 2016-04-27
Admin only components pages

Calvinp:
This implements admin restrictions in the UI in a slightly different way.
Instead of implementing the restriction in the router, it's implemented in the view. Instead of popping up a vex dialog, a permission denied page will be shown when trying to access a page you don't have rights to view.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserHolder.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserHolder.java
@@ -8,14 +8,17 @@ public class SingularityUserHolder {
   private final Optional<SingularityUser> user;
   private final boolean authenticated;
   private final boolean authEnabled;
+  private final boolean isAdmin;
 
   @JsonCreator
   public SingularityUserHolder(@JsonProperty("user") Optional<SingularityUser> user,
                                @JsonProperty("authenticated") boolean authenticated,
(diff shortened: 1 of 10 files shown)


ssalinas:
I definitely like this implementation better, @tpetr you agree?

zdhickman:
The `AdminOnly` wrapper component is a great idea.

Do you guys plan to add any other admin-only React pages in the future? If so, what do you think about adding an `AdminOnlyPage` component that makes use of `AdminOnly`? It would be inherited by this page and could be inherited by any other React page implemented in SingularityUI in the future. That would save you from adding some boilerplate on every page.

(maybe you could also make a base `Page` component. `TaskSearch` extends `AdminOnlyPage` extends `Page`, if you have shared structure across the React pages in Singularity)

Calvinp:
I do like the concept of an AdminOnlyPage (and Page) component.
Task Search in particular is not _always_ admin only - only global task search is admin only. That's the only problem I can see that would need to be worked around to implement AdminOnlyPage.

ssalinas:
If we add more, it's a once in a while thing (e.g. we are adding a webhooks page, but no other plans after that). However, we may eventually convert the previous admin pages to react at some point

zdhickman:
:+1: might be extra work without much gain, then.

ssalinas:
We are going to leave this one be for now, since the no-user bug was fixed, closing

PR #1013, opened 2016-04-28
Stop Task Files Jumping Around

Calvinp:
Task files always jumped around and reset the sort order when the page rerendered.
Now that won't happened.
The entire task table is reimplemented in React to facilitate this (which probably wouldn't be worth it if we didn't already have plans to move all of Singularity to React eventually).

This relies on some of the React component changes in #997, so don't merge this before that.


SingularityUI/app/collections/TaskFiles.coffee
```diff
@@ -53,6 +53,24 @@ class TaskFiles extends Collection
 
         taskFiles
 
+    sortBy: (field, sortDirectionAscending) ->
+        if field is 'name'
+            sorted = _.sortBy @models, (file) =>
+                if file.attributes.isDirectory
```
Calvinp: If we want to ignore case (because capital letters are always sorted before lowercase), do a .toLowercase() conversion in this function.

ssalinas:
@Calvinp one more thing I noticed on this. In the old version, when you navigated through the file tree, only the files section of the page would re-render. Now, when you are clicking through, it seems the entire page will. Can we get it so that only the relevant section updates?

Calvinp:
Fixed now. Thanks for that edit, it saved me a bit of time narrowing down the cause.
However, there's another bug - when you navigate on the file tree after sorting, the sort order doesn't actually match the glyphicon. Either navigating should clear the glyphicons, or sort order should be maintained. I'm going to implement the first one, as it seems less likely to confuse people.

ssalinas:
This has been looking good, thanks @Calvinp

PR #1014, opened 2016-04-28
If there are exactly four tasks, show their logs in a square

Calvinp:
The tailer used to show the logs in one row of three tiny log windows, and one huge log window across the bottom.
This changes it to show them in a 2x2 square.


ssalinas:
👍  glad this fix was fairly simple to make

PR #1015, opened 2016-04-28
Show only name of log file in aggregate tail title

Calvinp:
The title of the aggregate log page was showing `$TASK_ID` when the task id was in the path to a log file.
This fixes that, and improves the title to only show the name of the file. The fully qualified path, which was displayed before, is usually not as useful as the name when looking at a tab preview, and it was pushing out the name itself since the space is so tiny.


ssalinas:
Agreed that the file name is probably what we want there 👍

PR #1016, opened 2016-04-28
Don't show the number of bad things on Status Page if they're 0

Calvinp:
Don't show:
- Over provisioned requests
- Under provisioned requests
- Max task lag
unless there actually are any to show.


ssalinas:
This looks good, thanks @Calvinp

PR #1017, opened 2016-04-29
Tasks page improvements

Calvinp:
Implements several improvements to the tasks table page:
- `Name` field is changed to `Task ID` and `Started` is changed to `Started At`
- Host and rack fields link to a page of all tasks on said host or rack. It is now possible to filter by rackId.
- On screens that aren't extra small or large, the racks and memory columns now line up properly.
- The protip is now a tooltip on both this page and the requests page
- Tooltips on task state filter links now explain the meaning of those states. **Please review the wording of these tooltips**
- Fix an error in the decommissioning tasks table that caused it to never render.

### DIFF
--- a/SingularityUI/app/handlebarsHelpers.coffee
+++ b/SingularityUI/app/handlebarsHelpers.coffee
@@ -40,7 +40,6 @@ Handlebars.registerHelper 'ifTaskInList', (list, task, options) ->
     return options.inverse @
 
 Handlebars.registerHelper 'ifInSubFilter', (needle, haystack, options) ->
-    #return options.inverse @ unless haystack
     return options.fn @ if haystack is 'all'
     if haystack.indexOf(needle) isnt -1
         options.fn @
--- a/SingularityUI/app/handlebarsHelpers.coffee
+++ b/SingularityUI/app/handlebarsHelpers.coffee
@@ -40,6 +40,7 @@ Handlebars.registerHelper 'ifTaskInList', (list, task, options) ->
     return options.inverse @
 
 Handlebars.registerHelper 'ifInSubFilter', (needle, haystack, options) ->
+    #return options.inverse @ unless haystack
     return options.fn @ if haystack is 'all'
     if haystack.indexOf(needle) isnt -1
         options.fn @
--- a/SingularityUI/app/templates/requestTypeFilter.hbs
+++ b/SingularityUI/app/templates/requestTypeFilter.hbs
@@ -1,30 +1,30 @@
 <ul class="nav nav-pills nav-pills-multi-select">
     <li {{#unlessInSubFilter 'SERVICE' requestsSubFilter}}class="active"{{/unlessInSubFilter}}>
-        <a data-filter="SERVICE" class='has-tooltip' title="Protip: press cmd to select multiple" data-toggle='tooltip'>
+        <a data-filter="SERVICE">
             <span class="glyphicon glyphicon-ok"></span>
             Service
         </a>
     </li>
     <li {{#unlessInSubFilter 'WORKER' requestsSubFilter}}class="active"{{/unlessInSubFilter}}>
-        <a data-filter="WORKER" class='has-tooltip' title="Protip: press cmd to select multiple" data-toggle='tooltip'>
+        <a data-filter="WORKER">
             <span class="glyphicon glyphicon-ok"></span>
             Worker
         </a>
     </li>
     <li {{#unlessInSubFilter 'SCHEDULED' requestsSubFilter}}class="active"{{/unlessInSubFilter}}>
-        <a data-filter="SCHEDULED" class='has-tooltip' title="Protip: press cmd to select multiple" data-toggle='tooltip'>
+        <a data-filter="SCHEDULED">
             <span class="glyphicon glyphicon-ok"></span>
             Scheduled
         </a>
     </li>
     <li {{#unlessInSubFilter 'ON_DEMAND' requestsSubFilter}}class="active"{{/unlessInSubFilter}}>
-        <a data-filter="ON_DEMAND" class='has-tooltip' title="Protip: press cmd to select multiple" data-toggle='tooltip'>
+        <a data-filter="ON_DEMAND">
             <span class="glyphicon glyphicon-ok"></span>
             On-demand
         </a>
     </li>
     <li {{#unlessInSubFilter 'RUN_ONCE' requestsSubFilter}}class="active"{{/unlessInSubFilter}}>
-        <a data-filter="RUN_ONCE" class='has-tooltip' title="Protip: press cmd to select multiple" data-toggle='tooltip'>
+        <a data-filter="RUN_ONCE">
             <span class="glyphicon glyphicon-ok"></span>
             Run-once
         </a>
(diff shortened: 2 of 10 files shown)


SingularityUI/app/templates/tasksTable/tasksBase.hbs
```diff
@@ -2,19 +2,19 @@
     <div class="col-md-12">
         <ul class="nav nav-pills">
             <li {{#ifEqual tasksFilter "active"}}class="active"{{/ifEqual}}>
-                <a href="{{appRoot}}/tasks">Active</a>
+                <a href="{{appRoot}}/tasks" class='has-tooltip' title="Currently running tasks" data-toggle='tooltip'>Active</a>
             </li>
             <li {{#ifEqual tasksFilter "scheduled"}}class="active"{{/ifEqual}}>
-                <a href="{{appRoot}}/tasks/scheduled">Scheduled</a>
+                <a href="{{appRoot}}/tasks/scheduled" class='has-tooltip' title="Tasks that will run when the time comes" data-toggle='tooltip'>Scheduled</a>
```
ssalinas: Maybe more like 'Tasks that will run at the next scheduled interval'

SingularityUI/app/templates/requestTypeFilter.hbs
```diff
@@ -1,36 +1,32 @@
...
     <li {{#unlessInSubFilter 'ON_DEMAND' requestsSubFilter}}class="active"{{/unlessInSubFilter}}>
-        <a data-filter="ON_DEMAND">
+        <a data-filter="ON_DEMAND" class='has-tooltip' title="Protip: press cmd to select multiple" data-toggle='tooltip'>
             <span class="glyphicon glyphicon-ok"></span>
             On-demand
         </a>
     </li>
     <li {{#unlessInSubFilter 'RUN_ONCE' requestsSubFilter}}class="active"{{/unlessInSubFilter}}>
-        <a data-filter="RUN_ONCE">
+        <a data-filter="RUN_ONCE" class='has-tooltip' title="Protip: press cmd to select multiple" data-toggle='tooltip'>
```
ssalinas: Wondering if this is necessary, is there any reason they can't just be a toggle? Instead of having to hold down cmd, just clicking toggles that filter on/off, and all are selected by default to start with

SingularityUI/app/templates/tasksTable/tasksBase.hbs
```diff
@@ -2,19 +2,19 @@
...
             </li>
             <li {{#ifEqual tasksFilter "cleaning"}}class="active"{{/ifEqual}}>
-                <a href="{{appRoot}}/tasks/cleaning">Cleaning</a>
+                <a href="{{appRoot}}/tasks/cleaning" class='has-tooltip' title="Tasks that are cleaning" data-toggle='tooltip'>Cleaning</a>
             </li>
             <li {{#ifEqual tasksFilter "lbcleanup"}}class="active"{{/ifEqual}}>
-                <a href="{{appRoot}}/tasks/lbcleanup">LB Cleaning</a>
+                <a href="{{appRoot}}/tasks/lbcleanup" class='has-tooltip' title="Tasks that are load balancer cleaning" data-toggle='tooltip'>LB Cleaning</a>
```
ssalinas: 'Tasks that are being removed from the load balancer'

SingularityUI/app/templates/tasksTable/tasksBase.hbs
```diff
@@ -2,19 +2,19 @@
...
-                <a href="{{appRoot}}/tasks">Active</a>
+                <a href="{{appRoot}}/tasks" class='has-tooltip' title="Currently running tasks" data-toggle='tooltip'>Active</a>
             </li>
             <li {{#ifEqual tasksFilter "scheduled"}}class="active"{{/ifEqual}}>
-                <a href="{{appRoot}}/tasks/scheduled">Scheduled</a>
+                <a href="{{appRoot}}/tasks/scheduled" class='has-tooltip' title="Tasks that will run when the time comes" data-toggle='tooltip'>Scheduled</a>
             </li>
             <li {{#ifEqual tasksFilter "cleaning"}}class="active"{{/ifEqual}}>
-                <a href="{{appRoot}}/tasks/cleaning">Cleaning</a>
+                <a href="{{appRoot}}/tasks/cleaning" class='has-tooltip' title="Tasks that are cleaning" data-toggle='tooltip'>Cleaning</a>
```
ssalinas: Technically correct, but maybe for easier to read language something like, 'Tasks that have been asked to shut down'

SingularityUI/app/handlebarsHelpers.coffee
```diff
@@ -40,6 +40,7 @@ Handlebars.registerHelper 'ifTaskInList', (list, task, options) ->
     return options.inverse @
 
 Handlebars.registerHelper 'ifInSubFilter', (needle, haystack, options) ->
+    #return options.inverse @ unless haystack
```
ssalinas: there a reason this comment was added?

Calvinp: Artifact of one way I considered to fix a problem I was having. I'll remove it.

SingularityUI/app/templates/tasksTable/tasksActiveBody.hbs
```diff
@@ -4,15 +4,15 @@
...
                     </th>
                     <th class="hidden-xs" data-sort-attribute="taskId.startedAt">
-                        Started
+                        Started At
                     </th>
                     <th class="hidden-xs" data-sort-attribute="host">
                         Host
                     </th>
-                    <th class="hidden-xs" data-sort-attribute="taskId.rackId">
+                    <th class="visible-lg" data-sort-attribute="taskId.rackId">
```
ssalinas: What was the reasoning for making this visible vs not and lg vs xs? I would argue that just seeing the task id, cpu and mem on a small device is the more important. And probably host over rack if we were to choose one more to add

Calvinp: Honestly, I just chose the one I saw first as the correct one. There probably won't be room for hosts on a smaller screen if we want to show disk resource too though.

ssalinas: If you have the bandwidth it'd be worth a quick pass over the bootstrap classes for the table columns to make sure the collapse nicely (e.g. are xs and lg good breakpoints or is sm better? What are the most important to display and what can we hide?). If I remember correctly, visible-lg is actually deprecated as of 3.2 in favor of visible-(size)-(display style)

Calvinp: I'll take a look at this now.

Calvinp: At the smallest window sizes, the task ID and started at timestamp start running off the screen. But that happens regardless of what we're showing. hidden-xs seems to be ok - everything is able to display pretty much the same at the smaller sizes down to the xs cutoff. It might not be ok for a table with more columns though. Also, the actions icons (kill task and JSON) can fit without causing a change in the way the rest of the table looks even at the smallest sizes.

Calvinp: >Also, the actions icons (kill task and JSON) can fit without causing a change in the way the rest of the table looks even at the smallest sizes.
This does not appear to be the case, however, on the other tasks tables. Only on active. So I'm going to remove hidden-xs for the actions icons for active tasks only.

ssalinas:
One more comment on this. When in a filter that isn't `Active`, it doesn't seem like the service/worker/etc filters are working properly. For example, filters out all pending tasks on the scheduled filter unless all the request type filters are active

Calvinp:
As it turns out, pending tasks don't store their request type - you'd have to fetch the task to acquire that, which is certainly not reasonable to do for every single task.
So instead of fixing that I removed the request type filters entirely for tasks that aren't active.

ssalinas:
Looks good @Calvinp , the new toggles are much nicer/easier

PR #1018, opened 2016-04-29
Optional disk resource column tasks page

Calvinp:
This adds an optional 'disk resource' column to the tasks page. It will be configurable in singularity.yml under `ui`.
This is separate from #1017 because unlike the rest of the improvements to the tasks page, this depends on #993.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
@@ -61,7 +61,7 @@ public class UIConfiguration {
 
   @JsonProperty
   @NotNull
-  private Boolean showTaskDiskSpace = false;
+  private boolean showTaskDiskResource = false;
 
   private boolean hideNewDeployButton = false;
   private boolean hideNewRequestButton = false;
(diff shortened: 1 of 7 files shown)


ssalinas:
Looks good, but lets PR this into the disk resource branch, then we can move it all together

SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
```diff
@@ -59,6 +59,10 @@ public static RootUrlMode parse(String value) {
   @NotNull
   private List<ShellCommandDescriptor> shellCommands = Collections.emptyList();
 
+  @JsonProperty
+  @NotNull
+  private Boolean showTaskDiskSpace = false;
```
tpetr: i know this is a nitpick, but `showTaskDiskResource` better describes what this is

tpetr: also, `Boolean` is nullable while `boolean` is not (because it's a primitive type) -- better to not have to deal with `null` values when you don't need to

SingularityUI/app/templates/tasksTable/tasksActiveBody.hbs
```diff
@@ -21,6 +21,11 @@
                     <th class="visible-lg" data-sort-attribute="memoryMb">
                         Memory
                     </th>
+                    {{#if showDiskSpace}}
+                        <th class="visible-lg" data-sort-attribute="memoryMb">
+                            Disk Space
```
tpetr: I'd just call this `Disk`

SingularityUI/app/templates/tasksTable/tasksDecommissioningBody.hbs
```diff
@@ -22,6 +22,11 @@
                         <th class="visible-lg" data-sort-attribute="memoryMb">
                             Memory
                         </th>
+                        {{#if showDiskSpace}}
+                            <th class="visible-lg" data-sort-attribute="memoryMb">
```
ssalinas: Just noticed, sort attribute here should not be memoryMb

tpetr:
LGTM aside from feedback above. :+1: for PR-ing this against #993 directly.

Calvinp:
PRing against #993 - see #1020

PR #1020, opened 2016-05-02
Optional disk resource column tasks page

Calvinp:
Implements #1018 based on #993


ssalinas:
merging this into `disk` since they are both in stable and go together

PR #1022, opened 2016-05-03
Aggregate tailer tooltip host on instance number and tasks dropdown

Calvinp:
This implements a tooltip showing the host and deploy id on the instance number in the aggregate tailer, and one showing only the host on the dropdown showing all the tasks.

### DIFF
--- a/SingularityUI/app/utils.coffee
+++ b/SingularityUI/app/utils.coffee
@@ -223,17 +223,17 @@ class Utils
         else
             fuzzyObject.score
 
-    @getInstanceNumberFromTaskId: (taskId) ->
+    @getTaskDataFromTaskId: (taskId) ->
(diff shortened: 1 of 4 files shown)


SingularityUI/app/utils.coffee
```diff
@@ -227,6 +227,14 @@ class Utils
         splits = taskId.split('-')
         splits[splits.length - 3]
 
+    @getDeployIdFromTaskId: (taskId) ->
```
ssalinas: I think it might be a bit cleaner and more reusable to have a model for the TaskId. i.e. have something like a `fromString` method for the taskId model to set the attributes. Reusable where we need it, and you are then only calling the split once when you build the object instead of multiple times for the same string

ssalinas:
LGTM

PR #1023, opened 2016-05-03
Dashboard page improvements

Calvinp:
This implements the following changes to the dashboard page:
- Don't show the `change` link next to the username (it tried to call a nonexistent function, so we know it was useless anyway)
- Change the name of the `Requested` column in paused requests and starred requests to `Time of Last Deploy`. This also changes the name of the same column in the paused requests table on the Requests page
- Don't show `Deploy User` on the paused requests table on the dashboard (still will show it on the paused requests table on the Requests page)
- Add a `Request type` column on the starred requests table


ssalinas:
These look good, thanks @Calvinp

PR #1024, opened 2016-05-04
Don't show edit request button if config.hideNewRequestButton is true

Calvinp:
Fix the bug that showed edit request buttons even if `config.hideNewRequestButton` was true.


ssalinas:
👍  didn't realize we never had the config object available for those templates

SingularityUI/app/views/dashboard.coffee
```diff
@@ -41,6 +41,7 @@ class DashboardView extends View
                 requests: pausedRequests
                 haveRequests: pausedRequests.length > 0
                 requestsSubFilter: ''
+                hideNewRequestButton: config.hideNewRequestButton
```
Calvinp: This will cause a merge conflict with [Dashboard Page Improvements](https://github.com/HubSpot/Singularity/pull/1023/files#diff-8ede3f844c775e0db1d8b8c0ce9e6150R43). To resolve, simply keep both of the added lines.

ssalinas:
Thanks for fixing merge conflict, this looks good 👍

PR #1025, opened 2016-05-04
add support for per-task labels + surfacing a label as a LB upstream group

tpetr:
- Adds optional `taskLabels` field to `SingularityDeploy`, which allows the user to set task-specific labels (keyed by instance number)
- Adds optional `taskLabelForLoadBalancerUpstreamGroup` configuration setting, which allows the user to specify a specific Mesos task label whose value will be used in the Baragon upstream's `group` field.
- + Plumbing to make this all work

Depends on HubSpot/Baragon#179.

/cc @ssalinas


pom.xml
```diff
@@ -38,7 +38,7 @@
     <basepom.release.profiles>oss-release,build-swagger-documentation</basepom.release.profiles>
 
     <dropwizard.guicier.version>0.7.1.2</dropwizard.guicier.version>
-    <baragon.version>0.2.0</baragon.version>
+    <baragon.version>0.4.0-STAGE-SNAPSHOT</baragon.version>
```
tpetr: TODO: change this once the Baragon PR is legit rolled out

PR #1026, opened 2016-05-05
Aggregate tailer better handle logs not found

Calvinp:
Change the tailer to not send extra requests when we already know the task file doesn't exist, and to only show the name of the file in the 'does not exist' message.


SingularityUI/app/actions/log.coffee
```diff
@@ -148,8 +148,8 @@ taskGroupFetchNext = (taskGroupId) ->
       return Promise.resolve()
 
     dispatch({taskGroupId, type: 'LOG_REQUEST_START'})
-    promises = tasks.map ({taskId, maxOffset, path, initialDataLoaded}) ->
-      if initialDataLoaded
+    promises = tasks.map ({taskId, exists, maxOffset, path, initialDataLoaded}) ->
+      if initialDataLoaded and exists isnt false
```
Calvinp: Comparing this to false is on purpose - if exists is undefined, we may just not know yet if the file exists or not. If it's false, we know for sure the file doesn't exist.

PR #1029, opened 2016-05-06
Don't count non-active requests towards total number of requests on the dashboard page

Calvinp:
Fixes the bug where the total number of requests are not equal to the sum of the number of each type of request by not counting non-active requests towards the individual number of requests of each type.


Calvinp:
This competes with #1030 - don't merge them both

PR #1030, opened 2016-05-06
Count non-active requests towards request totals on dashboard page

Calvinp:
Fixes the bug where the total number of requests are not equal to the sum of the number of each type of request by counting non-active requests towards the total number of requests.


Calvinp:
This competes with #1029 - don't merge them both.

PR #1032, opened 2016-05-06
Allow glob matching in addition to fuzzy matching on Requests and Tasks page

Calvinp:
This allows the tasks and requests to be filtered by globs instead of by fuzzy matching, deciding which to use by detecting if the search term contains any of the glob characters.


SingularityUI/app/utils.coffee
```diff
@@ -226,7 +228,13 @@ class Utils
     @fileName: (filePath) ->
         filePath.substring(filePath.lastIndexOf('/') + 1)
 
+    @isGlobFilter: (filter) ->
+        for char in @GLOB_CHARS
+            return true if filter.indexOf(char) isnt -1
+        return false
+
     @fuzzyAdjustScore: (filter, fuzzyObject) ->
+        console.log(fuzzyObject)
```
wolfd: Forgotten console.log?

PR #1033, opened 2016-05-10
Support task history search on updatedAt

ssalinas:
Added support to the backend to also query on the `updatedAt` field. Then we will be able to do queries like, 'everything running on slave x at 10am' with  combo of the startedAt and updatedAt values.

TODO:
- [x] UI changes for specifying `updatedAt`


ssalinas:
This has been working well, further tweaks to task search will come in future PRs

PR #1034, opened 2016-05-10
Change package.json 'vex' dependency to 'vex-js'

wolfd:
The package.json for SingularityUI had "vex" as the dependency, but required "vex-js".

Also added .git to the urls for git dependencies. It's not necessary, but makes the file consistent with url schemes.


tpetr:
💃

PR #1037, opened 2016-05-13
Handle scale during deploy

ssalinas:
@stevenschlansker should fix that edge case you ran into the other day when scaling + deploying at the same time

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
@@ -251,6 +251,13 @@ public class SingularityDeployChecker {
       }
     }
 
+    if (request.isDeployable() && deployResult.getDeployState() == DeployState.SUCCEEDED && pendingDeploy.getDeployProgress().isPresent()) {
+      if (pendingDeploy.getDeployProgress().get().getTargetActiveInstances() != request.getInstancesSafe()) {
+        requestManager.addToPendingQueue(new SingularityPendingRequest(request.getId(), pendingDeploy.getDeployMarker().getDeployId(), deployResult.getTimestamp(),
+          pendingDeploy.getDeployMarker().getUser(), PendingType.UPDATED_REQUEST, request.getSkipHealthchecks(), pendingDeploy.getDeployMarker().getMessage()));
+      }
+    }
+
     if (requestWithState.getState() == RequestState.DEPLOYING_TO_UNPAUSE) {
       if (deployResult.getDeployState() == DeployState.SUCCEEDED) {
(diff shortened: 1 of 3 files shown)


tpetr:
Could you add a unit test for this?

ssalinas:
test added

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -603,7 +603,7 @@ private SingularityDeployResult markStepFinished(SingularityPendingDeploy pendin
     int numTasksToShutDown = Math.max(otherActiveTasks.size() - (request.getInstancesSafe() - deployProgress.getTargetActiveInstances()), 0);
     List<SingularityTaskId> sortedOtherTasks = new ArrayList<>(otherActiveTasks);
     Collections.sort(sortedOtherTasks, SingularityTaskId.INSTANCE_NO_COMPARATOR);
-    return sortedOtherTasks.subList(0, numTasksToShutDown);
+    return sortedOtherTasks.subList(0, Math.min(numTasksToShutDown, sortedOtherTasks.size() - 1));
```
tpetr: i might be misunderstanding something obvious, but why subtract one here?

ssalinas: e.g. size of the array is 2, so the max index we can use for the sublist is 1

tpetr: But if you look at that javadoc, `toIndex` is exclusive:

```
     * @param fromIndex low endpoint (inclusive) of the subList
     * @param toIndex high endpoint (exclusive) of the subList
     * @return a view of the specified range within this list
     * @throws IndexOutOfBoundsException for an illegal endpoint index value
     *         (<tt>fromIndex &lt; 0 || toIndex &gt; size ||
     *         fromIndex &gt; toIndex</tt>)
     */
```

ssalinas: So, is it the case that the index out of bounds is coming from the from index then instead? Had just assumed it was on the high end when I made this PR. must have been that the list was empty, will fix, thanks

ssalinas: Fxed to check both for toIndex errors(with the Math.min) and fromIndex (isEmpty check). First condition gets hit by the test I added, second condition will get hit by a number of other tests

ssalinas:
Found one last hitch with scale during deploy. If the timing is right, you will end up with the pre-scale number of requests running after the deploy is finished. The deploy ends with the targetActiveInstances, but we currently don't enqueue another pending request on successful deploy. Going to extend the test to check for this case then figure out the best way to fix it

tpetr:
LGTM

PR #1039, opened 2016-05-13
UI improvements with global search tool

wolfd:
Makes a number of improvements to the global search tool:
- The search tool is brought up immediately instead of waiting for the backbone collection to be loaded
- Pseudo-tab completion is possible (it picks and navigates to the currently selected suggestion, or defaults to the first value in the list if nothing is highlighted)
- Hitting enter does the same thing as Tab, immediately navigating to the result

I want to add one more thing into this PR before I'm satisfied with it, and that is highlighting the portions of each result that are being matched by the fuzzy search. This was present with twitter-typeahead, but requires some additional work with react-typeahead.


SingularityUI/app/components/globalSearch/GlobalSearch.cjsx
```diff
@@ -0,0 +1,52 @@
...
+  render: ->
+    if @props.visible
+      @focus()
+
+    options = _.pluck(@props.requests.toJSON(), 'id')
+
+    globalSearchClasses = [
+      'global-search'
+      'global-search-active' if @props.visible
+    ].join(' ')
```
tpetr: easier to use the `classnames` library for this (example: https://github.com/HubSpot/Singularity/blob/master/SingularityUI/app/components/logs/LoadingSpinner.cjsx#L9-L12)

SingularityUI/webpack.config.js
```diff
@@ -52,7 +52,8 @@ module.exports = {
       'handlebars': 'handlebars/runtime.js',
       'sortable': 'sortable/js/sortable.js',
       'datatables': 'datatables/media/js/jquery.dataTables.js',
-      'bootstrap': 'bootstrap/dist/js/bootstrap.js'
+      'bootstrap': 'bootstrap/dist/js/bootstrap.js',
+      'react': path.join(__dirname, 'node_modules', 'react')
```
tpetr: what needed this tweak?

SingularityUI/app/components/globalSearch/GlobalSearch.cjsx
```diff
@@ -0,0 +1,52 @@
+React = require 'react'
+
+{ Typeahead } = require 'react-typeahead'
+
+
+GlobalSearch = React.createClass
```
tpetr: it'd be preferable to do `class GlobalSearch extends React.Component` instead

SingularityUI/webpack.config.js
wolfd: Good question,

I needed to do some development on react-typeahead, and I was using `npm link` to do so.
This line makes sure that there is only one version of react being used at runtime, because react-typeahead required react locally to build.

I left it in because I anticipated needing the fix to be in place in the future. YMMV.

I can take it out if you want.

SingularityUI/app/components/globalSearch/GlobalSearch.cjsx
wolfd: Good point.

I did notice that in `components/common` all of the React components are in the `X = React.createClass` syntax.

Is there a reason for that, or should we change them to the `extends` syntax?

SingularityUI/webpack.config.js
tpetr: Let's remove it for now, thanks

SingularityUI/app/components/globalSearch/GlobalSearch.cjsx
tpetr: Forgot to circle back here -- extending `React.Component` comes with some benefits, such as `displayName` being properly set by default when you try to inspect the page

wolfd:
Here's a screenshot of the new UI, not much has changed.
The highlighted text is bolder, and all of the options are `a` tags now.

screen shot 2016-05-18 at 1 06 21 pm

In keybinding world, you should be able to just hit enter without hitting the down arrow first if the first option is what you want.

In a later PR, we can make tab move down the list, but that isn't supported by react-typeahead yet, so it will take another PR there.

SingularityUI/app/components/globalSearch/GlobalSearch.cjsx
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    return bolded
+
+  getValueFromOption: (option) ->
+    return option.original
+
+  render: =>
+    if @props.visible
+      @focus()
```
tpetr: the more idiomatic way to do this would be to use [componentDidUpdate()](https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate), and only call `@focus()` if the `visible` prop transitioned from `false` to `true` (as a general rule of thumb, you shouldn't be manipulating the DOM or triggering events like that in `render()`)

PR #1043, opened 2016-05-18
Add support for hourly logrotate

tpetr:
Needed for verbose tasks where daily log rotation isn't often enough.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/executor/SingularityExecutorLogrotateFrequency.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/executor/SingularityExecutorLogrotateFrequency.java
@@ -1,0 +1,27 @@
+package com.hubspot.singularity.executor;
+
+import com.google.common.base.Optional;
+
+public enum SingularityExecutorLogrotateFrequency {
+    HOURLY("hourly", Optional.of("0 * * * *")),
+    DAILY("daily", Optional.<String>absent()),
+    WEEKLY("weekly", Optional.<String>absent()),
+    MONTHLY("monthly", Optional.<String>absent());
+
+    private final String logrotateValue;
+    private final Optional<String> cronSchedule;
+
+    SingularityExecutorLogrotateFrequency(String logrotateValue, Optional<String> cronSchedule) {
+        this.logrotateValue = logrotateValue;
+        this.cronSchedule = cronSchedule;
+    }
+
+    public String getLogrotateValue() {
+        return logrotateValue;
+    }
+
+    public Optional<String> getCronSchedule() {
+        return cronSchedule;
+    }
+}
+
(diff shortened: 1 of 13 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/executor/SingularityExecutorLogrotateFrequency.java
```diff
@@ -0,0 +1,19 @@
+package com.hubspot.singularity.executor;
+
+public enum SingularityExecutorLogrotateFrequency {
+    HOURLY("monthly"),
```
ssalinas: meant hourly here?

tpetr: meant to add a comment here -- `hourly` isn't supported by all versions of logrotate. my intent was to make it run as infrequently as possible (hence `monthly`) and then manually trigger logrotate by the executor. (open to other options if you have other ideas here)

ssalinas: Right, forgot it doesn't actually have an hourly setting. Would it make sense to instead write something to `/etc/cron.hourly/`? Value probably doesn't matter in either case since it's manually triggered. Then we could let cron do it's thing rather than having the executor worry about it. Just a thought, this way would work too.

wsorenson: wouldn't it be far simpler if we made this functionality only work with versons of logrotate which support hourly?

ssalinas: Should have clarified that comment, logrotate straight up doesn't support hourly (as far as I know). Most places suggest setting things up in cron.hourly or running a command/script manually if you need to rotate more than daily. From man pages:
```
       Normally, logrotate is run as a daily cron job.  It will not  modify  a
       log  multiple  times  in  one  day unless the criterium for that log is
       based on the log’s size and logrotate is being run multiple times  each
       day, or unless the -f or -force option is used.
```

tpetr: The more I think about it, the more I think it'd be OK to go down the `/etc/cron.hourly/` route. I'll update the PR.

tpetr:
Updated PR use cron for hourly logrotate, confirmed working in the test cluster.

SingularityBase/src/main/java/com/hubspot/singularity/executor/SingularityExecutorLogrotateFrequency.java
solomonty: The [puppet-logrotate module](https://github.com/rodjek/puppet-logrotate) and newer versions of [logrotate](https://github.com/logrotate/logrotate) support hourly rotation.

We sometimes need to rotate more often than hourly and we create a manual entry in /etc/cron.d with the appropriate schedule and use the `-f` option.

PR #1045, opened 2016-05-18
configurable max metadata message length

ssalinas:
Since this is stored in zk, would be wise to limit the length of the message and title so we aren't storing huge objects. Limits are pretty high by default here, open to suggestion for better ones

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
@@ -340,11 +340,12 @@ public class TaskResource {
 
     authorizationHelper.checkForAuthorizationByTaskId(taskId, user, SingularityAuthorizationScope.WRITE);
 
-    WebExceptions.checkBadRequest(taskMetadataRequest.getTitle().getBytes().length < taskMetadataConfiguration.getMaxMetadataTitleBytes(),
-      "Task metadata title too long, must be less than %s bytes", taskMetadataConfiguration.getMaxMetadataTitleBytes());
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -340,6 +340,12 @@ public void postTaskMetadata(@PathParam("taskId") String taskId, final Singulari
 
     authorizationHelper.checkForAuthorizationByTaskId(taskId, user, SingularityAuthorizationScope.WRITE);
 
+    WebExceptions.checkBadRequest(taskMetadataRequest.getTitle().getBytes().length < taskMetadataConfiguration.getMaxMetadataTitleBytes(),
+      "Task metadata title too long, must be less than %s bytes", taskMetadataConfiguration.getMaxMetadataTitleBytes());
+
+    WebExceptions.checkBadRequest(!taskMetadataRequest.getMessage().isPresent() || taskMetadataRequest.getMessage().get().getBytes().length < taskMetadataConfiguration.getMaxMetadataMessageBytes(),
+      "Task metadata message too long, must be less than %s bytes", taskMetadataConfiguration.getMaxMetadataMessageBytes());
```
tpetr: `getBytes()` causes the string to be encoded into a byte array, which is kind of wasteful -- i think it's reasonable (and easier to reason about) if we impose the limit as a max length of the `title` and `message` Strings

ssalinas: thanks, updated

Issue #1046, opened 2016-05-18
singularity 0.6.0 file tailer broken in browsers that don't do ES6


stevenschlansker:
```
[15:34:57.941] TypeError: Object.assign is not a function @ http://singularity-prod-uswest2.otenv.com/static/js/app.js:64
```

Singularity now requires `Object.assign` and `Promise` support.  My browser doesn't support it 😢 

Can we include polyfills for both?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
https://github.com/taylorhakes/promise-polyfill

stevenschlansker:
fixed by #1050

PR #1050, opened 2016-05-19
add polyfills for Object.assign and Promise

tpetr:
Adds polyfills for `Object.assign()` and `Promise` for older brosers. Longer term fix is to use Babel to add all necessary polyfills.

/cc @stevenschlansker #1046


tpetr:
@ssalinas we should also probably release a `0.6.1` for folks that aren't ready to update their Mesos cluster. We can square that away separately (probably via branching off the `Singularity-0.6.0` tag).

ssalinas:
Yeah, sounds good. I can do the git work for that when we are ready to release

stevenschlansker:
Thanks guys, you're the best!  FWIW I solved my issue locally: https://addons.palemoon.org/extensions/other/pm-promises/ but I'm sure someone else will run into this as well.

PR #1051, opened 2016-05-19
configurable max attempts for docker pull in custom executor

ssalinas:
@jonathanwgoodwin , still need to test this out, but should do what we need

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
@@ -218,6 +218,9 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   private int dockerClientConnectionPoolSize = 5;
 
   @JsonProperty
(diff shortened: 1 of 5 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -218,6 +218,9 @@
   private int dockerClientConnectionPoolSize = 5;
 
   @JsonProperty
+  private int dockerPullRetries = 1;
```
tpetr: does this mean it will attempt once, or retry once? it might make more sense to rename this to something like `maxDockerPullAttempts`

ssalinas: Retry once, you're right that it could be confusing, I'll update the name

ssalinas: @tpetr renamed

ssalinas:
This has been working fine in stable, merging

PR #1052, opened 2016-05-19
Kill tasks below a certain priority level

tpetr:


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityPriorityKillPoller.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            final double minPriorityLevel = maybePriorityFreeze.get().getPriorityFreeze().getMinimumPriorityLevel();
+
+            // map request ID to priority level
+            final Map<String, Double> requestIdToTaskPriority = new HashMap<>();
+            for (SingularityRequestWithState requestWithState : requestManager.getRequests()) {
+                requestIdToTaskPriority.put(requestWithState.getRequest().getId(), requestWithState.getRequest().getTaskPriorityLevel().or(defaultTaskPriorityLevel));
+            }
+
+            // cancel pending tasks below minimum priority level
```
ssalinas: I'd be worried about deleting pending tasks, because there won't necessarily be a pending request in the queue to put them back when we undo the freeze.  We would have to do something like we do on startup where we loop over all tasks. Maybe the scheduler should just be smart enough to ignore those pending tasks?

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -1,5 +1,7 @@
 package com.hubspot.singularity.scheduler;
 
+
```
ssalinas: extra new lines

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityPriorityKillPoller.java
tpetr: That's a fair point -- I certainly don't want to introduce new edge-cases due to deleting pending tasks. I'll update this to ignore pending tasks below the minimum priority level and deal with the fact that it'll cause task lag alerts (which could be good anyways -- a nice reminder that you have a priority freeze enabled...)

tpetr: Addressed your feedback in the latest commit. Not going to tackle sorting scheduled tasks by priority quite yet, but may do it in a future PR.

ssalinas:
Fixed merge conflict with master, going to merge this one.

PR #1056, opened 2016-05-23
UTF-8 support in chunked logs

wolfd:
On hold: Waiting on upstream Mesos patch to their log tailing endpoint. Probably Mesos 1.0 will fix this by using base64 encoded data.
-------------
* Add support for dropping invalid UTF-8 bytes at the beginning and end of log chunks

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
@@ -7,12 +7,15 @@ import java.util.Collections;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Charsets;
 import com.google.common.base.Optional;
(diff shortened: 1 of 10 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
```diff
@@ -66,6 +67,94 @@ public SlaveNotFoundException(Exception e) {
...
+      return 1;
+    } else if (b >= (byte)0b1110_0000 && b <= (byte)0b1110_1111) {
+      return 2;
+    } else if (b >= (byte)0b1111_0000 && b <= (byte)0b1111_0111) {
+      return 3;
+    } else {
+      // this could be an ASCII char or a continuation byte
+      return 0;
+    }
+  }
```
tpetr: - make `isContinuationChar()` and `numberOfFollowingBytes()` static methods, since they don't depend on any local state
- make all those binary numbers `private static final byte` constants

tpetr:
Closing this for now.

PR #1059, opened 2016-06-02
set proper Content-Type and Content-Encoding for s3 uploads

tpetr:
This PR makes the S3Uploader attempt to set the proper `Content-Type` and `Content-Encoding` headers for files uploaded to S3.


jhaber:
This will be so awesome

tpetr:
This would probably be good default settings:
```yaml
s3ContentHeaders:
- filenameEndsWith: .log.gz
  contentType: text/plain
  contentEncoding: gzip
- filenameEndsWith: .log
  contentType: text/plain
```

PR #1060, opened 2016-06-02
Setup babel for ES6 and JSX transformation.

wolfd:
The eventual goal is to have SingularityUI to be entirely a React/Babel ES6 project.

Steps taken:
- Convert GlobalSearch to JSX/ES6

Notes on how to do coffeescript -> es6 transformation:
- the `decaffeinate` npm module does a lot of the heavy lifting, but do it one file at a time
- comment out all of the CJSX parts of the file, convert those manually
- rewrite parts that use coffeescript specific functionality (such as the `in` keyword)
- because of the differences in ES6 imports versus CommonJS, you have to add `.default` to the ES6 imports


SingularityUI/app/components/globalSearch/GlobalSearch.jsx
```diff
@@ -0,0 +1,115 @@
...
+  componentDidUpdate(prevProps, prevState) {
+    if (this.props.visible && (this.props.visible !== prevProps.visible)) {
+      return this.focus();
+    }
+  }
+
+  render() {
+    let options = _.pluck(this.props.requests.toJSON(), 'id');
+
+    globalSearchClasses = classNames({
```
tpetr: add `const globalSearchClasses` here instead of the `let` at the top of the file

SingularityUI/webpack.config.js
```diff
@@ -68,7 +70,7 @@ module.exports = {
       }
     }),
     new webpack.DefinePlugin({
-      'process.env.NODE_ENV': '"production"'
+      'process.env.NODE_ENV': JSON.stringify('development')
```
tpetr: curious, why the change here?

wolfd: Oh, right. I forgot that I put this in there.

We should figure out if there's a way to override the default behavior and not do this.

Minified react (which is used when you use `production`) doesn't have good error messages. I wanted to see what the issue was with something, and they were being suppressed. By the time this patch gets merged I'd like to see this behavior default to production but use development if specified on the command line.

PR #1066, opened 2016-06-07
(logfetch) correct the file extension when auto-decompressing

ssalinas:
Now that we have the `Content-Type` and `Content-Encoding` headers set on the s3 uploads, we need to handle them on downloads too! Keeping the auto-decompress since it will make the search faster in the end, this just adds in some checks to correct the file extension if we decompressed it already


tpetr:
LGTM

PR #1070, opened 2016-06-07
use 0.28.2 to fix MESOS-5449

ssalinas:
Need to wait for the docker images and packages to officially be published, but the release was confirmed today. This will fix the memory leak in `declineOffer`.


ssalinas:
The docker image for 0.28.2 never got published, but I want to get this out to fix the memory leak. Will likely be changing the docker image when upgrading to 1.0.0 anyways and it's fine to develop in for now.

Moving this to staging shortly

PR #1072, opened 2016-06-08
Add linting, add .eslintrc, fix some of the errors

wolfd:
There's a *lot* more stuff to fix up before this stops yelling.

We should also expand on the `.eslintrc` with good rules, there's only a few things there now.


andyhuang91:
Instead of adding your own rules, I recently added https://github.com/HubSpot/javascript/tree/master/packages/eslint-config-hubspot
into the OrionUI project with zach. Running eslint with the `--fix` command will automatically fix some of the errors for you, but you should manually verify that all of those changes are valid

tpetr:
LGTM, thanks

PR #1073, opened 2016-06-08
Build the UI as its own module

kwm4385:
We want Singularity to build independently of SingularityService. This creates a new pom for the UI and moves the build into that. SingularityService then depends on that and copies the files from `dist/` to its own resources.

@tpetr

### DIFF
--- a/SingularityUI/gulpfile.js
+++ b/SingularityUI/gulpfile.js
@@ -41,7 +41,7 @@ var templateData = {
   redirectOnUnauthorizedUrl: process.env.SINGULARITY_REDIRECT_ON_UNAUTHORIZED_URL || ''
 }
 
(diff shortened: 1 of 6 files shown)


SingularityUI/gulpfile.js
```diff
@@ -50,7 +50,8 @@ var WebpackDevServer = require('webpack-dev-server');
 gulp.task("clean", function() {
   return del([
     path.resolve(dest, 'static/**'),
-    path.resolve(dest, 'index.html')], {force: true});
+    path.resolve(dest, 'index.html'),
+    dest], {force: true});
```
tpetr: i had to add `{force: true}` in the original implementation because we were deleting a folder outside the working directory -- this shouldnt be needed anymore

tpetr: in fact, i would further simplify this to just `del([dest])`

PR #1075, opened 2016-06-08
handle new lines better in docker.env using bashEscaped helper

ssalinas:
still sourcing docker.env in the runner script as well so that any vars set can be used in docker arguments

/cc @jonathanwgoodwin

### DIFF
--- a/SingularityExecutor/src/main/resources/deploy.env.hbs
+++ b/SingularityExecutor/src/main/resources/deploy.env.hbs
@@ -1,7 +1,13 @@
 # mesos task env
+{{#if docker}}
+{{#each env}}
+{{{name}}}={{{bashEscaped value}}}
+{{/each}}
+{{else}}
 {{#each env}}
 export {{{name}}}={{{bashEscaped value}}}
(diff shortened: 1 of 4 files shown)


SingularityExecutor/src/main/resources/deploy.env.hbs
```diff
@@ -1,7 +1,13 @@
 # mesos task env
+{{#if docker}}
+{{#each env}}
+{{{name}}}={{{bashEscaped value}}}
```
tpetr: why no export for docker?

ssalinas: it can't read it that way. The docker.env file can only be read as KEY=VAL, although within the value it will act like bash (as far as i know)

tpetr: ohh, docker is reading this file. gotcha.

ssalinas:
docker is silly, closing this and going to try a different strategy (already reverted the change on hs_staging branch)

PR #1081, opened 2016-06-10
handle new lines in docker environment variables correctly

ssalinas:
@jonathanwgoodwin was a bit trickier than I thought, but this should work

### DIFF
--- a/SingularityExecutor/src/main/resources/docker.sh.hbs
+++ b/SingularityExecutor/src/main/resources/docker.sh.hbs
@@ -61,8 +61,12 @@ DOCKER_IMAGE={{{ envContext.dockerInfo.image }}}
 # load env vars
 touch docker.env
 {{#each envContext.env}}
+{{#ifHasNewLines value}}
+{{{name}}}={{{bashEscaped value}}}
+{{else}}
 {{{name}}}={{{bashEscaped value}}}
 echo "{{{name}}}={{{value}}}" >> docker.env
+{{/ifHasNewLines}}
 {{/each}}
 
 # Create log directory for logrotate runs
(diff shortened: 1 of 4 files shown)


SingularityExecutor/src/main/resources/docker.sh.hbs
```diff
@@ -120,8 +124,8 @@ sudo chown -R {{{ runContext.user }}} {{{ runContext.taskAppDirectory }}}
 enable_cgroup_hierarchy || true
 
 # Start up the container
-echo "Creating continer with: sudo -E -H -u {{{ runContext.user }}} docker create $DOCKER_OPTIONS $DOCKER_IMAGE {{{ runContext.cmd }}}"
-cid=`sudo -E -H -u {{{ runContext.user }}} docker create $DOCKER_OPTIONS $DOCKER_IMAGE {{{runContext.cmd }}}`
+echo "Creating continer with: sudo -E -H -u {{{ runContext.user }}} docker create $DOCKER_OPTIONS {{#each envContext.env}}{{#ifHasNewLines value}}-e {{{name}}}={{#escapedNewLines value}}{{/escapedNewLines}}{{/ifHasNewLines}}{{/each}} $DOCKER_IMAGE {{{ runContext.cmd }}}"
+cid=`sudo -E -H -u {{{ runContext.user }}} docker create $DOCKER_OPTIONS {{#each envContext.env}}{{#ifHasNewLines value}}-e {{{name}}}={{#escapedNewLines value}}{{/escapedNewLines}}{{/ifHasNewLines}}{{/each}} $DOCKER_IMAGE {{{runContext.cmd }}}`
```
tpetr: i'm not sure you need closing blocks for `escapedNewLines`

tpetr:
For posterity, can you explain what this PR exactly does? Seems like it adds explicit `-e` args for docker env vars with newlines?

ssalinas:
Explanation for this:

When docker reads a .env file, it is not interpreted like bash as one would expect. Docker has its own parser that reads the variables in. This parser does not handle values that span multiple lines, even if they are quoted. So, any lines following a value with new lines could be messed up when read, and the value with new lines will only have the first line.

When instead using explicit `-e` options versus an env file, as long as they are escaped, new line values are acceptable and interpreted correctly.

What this PR does is separates out any environment variables that may contain new lines and sets them using `-e` instead of putting them in the env file

PR #1088, opened 2016-06-14
Avoid race condition in deploy to unpause

ssalinas:
If the request status is not updated to `DEPLOY_TO_UNPAUSE` _before_ the deploy is saved. It is possible that the deploy checker can run the first check of the deploy while the request is still in the paused state. This code path also left the deploy in a waiting state with no message.

Changes here move the request state update before the deploy save, and will also be sure to complete the deploy with a failure when hitting the code path where request state is not valid to proceed with the deploy


ssalinas:
This one is gtg, going to merge

PR #1091, opened 2016-06-15
support overriding the log level in tests

tpetr:
Haven't 100% tested this yet, but this PR should make it much easier to bump up log verbosity when diagnosing things.

/cc @ssalinas @jhaber


ssalinas:
tested this out as well, worked well for me 👍

PR #1092, opened 2016-06-15
New deploy detail page

kwm4385:
image

Deploy detail react/redux
@tpetr @wolfd @Calvinp


tpetr:
😆 💦

SingularityUI/app/controllers/DeployDetail.es6
```diff
@@ -0,0 +1,45 @@
...
+import { FetchAction as DeployFetchAction} from '../actions/api/deploy';
+import { fetchTask as TaskFetchAction} from '../actions/api/task';
+import { FetchForDeployAction } from '../actions/api/tasks';
+import { FetchForDeploy as TaskHistoryFetchForDeploy } from '../actions/api/taskHistory';
+
+class DeployDetailController extends Controller {
+
+    initialize({store, requestId, deployId}) {
+        app.showPageLoader()
+        this.title('Status');
```
tpetr: need to update this value

SingularityUI/app/controllers/DeployDetail.es6
```diff
@@ -0,0 +1,45 @@
...
+        });
+
+        let tasksPromise = this.store.dispatch(FetchForDeployAction.trigger(requestId, deployId));
+        tasksPromise.then(() => {
+          for (let t of store.getState().api.activeTasksForDeploy.data) {
+            this.store.dispatch(TaskFetchAction(t.taskId.id));
+          }
+        });
+
+        this.store.dispatch(TaskHistoryFetchForDeploy.trigger(requestId, deployId, 5, 1));
```
tpetr: it'd be a better experience for the user if we held of calling `showView()` until all the promises (or atleast, all the ones for visible things) have been resolved. i'd suggest coalescing the individual promises into an uber-promise using `Promise.all()`. and then calling `showView()` once the uber-promise has resolved.

SingularityUI/app/controllers/DeployDetail.es6
```diff
@@ -0,0 +1,45 @@
...
+          this.setView(new DeployView(store));
+          app.showView(this.view);
+        });
+
+        let tasksPromise = this.store.dispatch(FetchForDeployAction.trigger(requestId, deployId));
+        tasksPromise.then(() => {
+          for (let t of store.getState().api.activeTasksForDeploy.data) {
+            this.store.dispatch(TaskFetchAction(t.taskId.id));
+          }
+        });
```
tpetr: there are identical `dispatch()`es in `refresh()` -- can we call `refresh()` instead?

tpetr:
- Active Tasks table shouldn't collapse
- Updated column of Active Tasks and Task History tables always says "a few seconds ago" which isnt always accurate
- Latest Healthchecks table is showing healthchecks for a previous web service (that part of the store needs to be be cleared out)

Calvinp:
>Updated column of Active Tasks and Task History tables always says "a few seconds ago" which isnt always accurate

You should look outside of your code for the fix to this one - timestamps are broken elsewhere too.

kwm4385:
^ Above all should be fixed

tpetr:
@calvinp I disagree -- this PR should keep its focus on reactifying and reduxifying the deploy detail page. It should be up to the owner (ex. Me for racks and slaves page) to fix their own mistakes or omissions. Though, additional PR's are welcome if folks have extra bandwidth too! 😃

kwm4385:
At least in this case the timestamp issue was simply because I had the wrong field. I'm not aware of a general problem with timestamps across the app.

tpetr:
- Don't show Latest Healthchecks table if the Request type isnt `SERVICE`
- Task History shouldn't collapse either

tpetr:
^ should be good to go once those last two issues are squared away, nice work!

kwm4385:
^ also removed collapse from task history

kwm4385:
🔨

PR #1101, opened 2016-06-20
Remove extraneous console.log in utils fuzzy search function

Calvinp:
Removes a console.log that leads to an absurd amount of logging when searching.


kwm4385:
dupe of https://github.com/HubSpot/Singularity/pull/1083

Calvinp:
You are right. Not sure why I didn't see that

PR #1102, opened 2016-06-20
Refactor post action

kwm4385:
Move `url` into a generalized `opts` function for the action creator. Refactored all existing actions to use the new format.

@Calvinp @tpetr @wolfd


SingularityUI/app/actions/api/slaves.es6
```diff
@@ -2,6 +2,6 @@ import buildApiAction from './base';
 
 const POST_JSON = {method: 'POST', headers: {'Accept': 'application/json', 'Content-Type': 'application/json'}};
```
Calvinp: This is unnecessary now that buildJsonSendingApiAction() is implemented.

PR #1103, opened 2016-06-20
New function to do the boilerplate for JSON POST

Calvinp:
Add the ability to do JSON boilerplate work to the action creator.

### DIFF
--- a/SingularityUI/app/actions/api/base.es6
+++ b/SingularityUI/app/actions/api/base.es6
@@ -10,7 +10,7 @@ export default function buildJsonSendingApiAction(actionName, httpMethod, opts={
   let optsFunctionOrObject;
   if (typeof opts === 'function') {
     optsFunctionOrObject = (...args) => {
-      let optsFunctionResult = _.extend({}, jsonBoilerplate, opts(...args));
+      const optsFunctionResult = _.extend({}, jsonBoilerplate, opts(...args));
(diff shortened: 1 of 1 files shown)


SingularityUI/app/actions/api/base.es6
```diff
@@ -1,5 +1,22 @@
 import fetch from 'isomorphic-fetch';
 
+export default function buildPostJsonApiAction(actionName, opts={}) {
+  let jsonBoilerplate = {
+    method: 'POST',
+    headers: {'Content-Type': 'application/json', 'Accept': 'application/json'}
+  }
```
tpetr: this should probably be a `const` defined outside of the method

SingularityUI/app/actions/api/base.es6
```diff
@@ -1,5 +1,22 @@
...
+  let newOpts;
+  if (typeof opts === 'function') {
+    newOpts = (...args) => {
+      return _.extend(jsonBoilerplate, opts(...args));
+    }
```
tpetr: `{ }`'s and `return` can be removed for brevity

SingularityUI/app/actions/api/base.es6
Calvinp: True, but not if we want to be able to use methods other than POST.
`headers` can still be a `const` though

PR #1104, opened 2016-06-21
React-ify New/Edit request form

Calvinp:
Reactify the New/Edit Request Form


Calvinp:
Scheduled request bug is fixed.

SingularityUI/app/actions/api/base.es6
```diff
@@ -46,7 +46,15 @@ export function buildApiAction(actionName, opts={}) {
 
       let options = optsFunc(...args);
       return fetch(config.apiRoot + options.url, _.extend({credentials: 'include'}, _.omit(options, 'url')))
-        .then(response => response.json())
+        .then(response => {
+          if (response.headers.get('Content-Type') === 'application/json') {
+            return response.json();
+          } else {
+            // All API calls respond with JSON, so if we're not getting JSON it's bad
```
tpetr: Not sure if I completely agree with this... I like how you're leveraging the content type to decide how to decode the response, but we should really be keying off the response's status code to decide if there was an error or not

SingularityUI/app/actions/api/request.es6
```diff
@@ -0,0 +1,4 @@
+import { buildApiAction, buildJsonApiAction } from './base';
+
+export const FetchAction = buildApiAction('FETCH_REQUEST', (requestId) => {return {url: `/requests/request/${ requestId }`}});
```
tpetr: you can simplify the lambda here like this:

```javascript
(requestId) => ({url: `/requests/request/${ requestId }`})
```

(the parens around the object make the complier do the right thing)

SingularityUI/app/actions/form.es6
```diff
@@ -0,0 +1,25 @@
...
+;
+
+export const clearForm = (formId) =>
+  ({
+    formId,
+    type: 'CLEAR_FORM'
+  })
+;
+
+export default {modifyField, clearField, clearForm};
```
tpetr: i believe this line is unnecessary

SingularityUI/app/components/common/formItems/Checkbox.jsx
```diff
@@ -0,0 +1,23 @@
...
+import FormField from './FormField';
+
+class CheckBox extends React.Component {
+
+  render() {
+    return (
+      <FormField
+        id = {this.props.id}
+        noFormControlClass = {this.props.noFormControlClass}
+        prop = {{
```
tpetr: why are all these properties nested inside `prop`?

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,560 @@
...
+      }
+    }
+  }
+
+  renderRequestTypeSelectors() {
+    const selectors = [];
+    const tooltip = (
+      <ToolTip id="cannotChangeRequestTypeAfterCreation">Option cannot be altered after creation</ToolTip>
+    );
+    REQUEST_TYPES.map((requestType, key) => {
```
tpetr: `map()` returns an array of all the return values of the inner method -- it'd be much cleaner to do something like:

```javascript
const selectors = REQUEST_TYPES.map((requestType, key) => { return <TheThingToBeAdded/>; })
```

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,560 @@
...
+    clearForm(formId) {
+      dispatch(clearForm(formId));
+    },
+    save(requestBody) {
+      dispatch(SaveAction.trigger(requestBody));
+    }
+  }
+}
+
+export default connect(mapStateToProps, mapDispatchToProps, undefined, {pure: false})(RequestForm);
```
tpetr: why `pure: false` ?

SingularityUI/app/controllers/RequestForm.es6
```diff
@@ -0,0 +1,32 @@
...
+    this.setView(new RequestFormView({store, edit}));
+    app.showView(this.view);
+  }
+
+  initialize({store, requestId = ''}) {
+    app.showPageLoader();
+    this.title(`${requestId ? 'Edit' : 'New'} Request`);
+    this.store = store;
+
+    const racksPromise = this.store.dispatch(RacksFetchAction.trigger());
```
tpetr: I think we should wait on the Racks promise too

SingularityUI/app/components/requestForm/RequestForm.jsx
Calvinp: This is the simplest solution to the fact that react-redux performs a shallow comparison. props.form is an object, and redux doesn't compare its fields to those of the previous props when deciding whether to tell the page about the new props at all.

An alternative is to have all the fields in form surfaced at the top level. Another is to surface the modifications field. Both of those would be more complex, however, than using `pure: false`.

SingularityUI/app/components/common/formItems/Checkbox.jsx
Calvinp: Unfortunately that's how FormField is, and FormField is used in a bunch of other places. So I didn't want to be so fast to modify it.

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,560 @@
...
+                id = 'instances'
+                prop = {{
+                  updateFn: event => this.updateField('instances', event.target.value),
+                  placeholder: "1",
+                  inputType: 'text',
+                  value: this.getValue('instances')
+                }}
+              />
+            </div> :
+            undefined
```
Calvinp: Should these `undefined`s be `null`s?
I know react renders both the same way, but I saw @kwm4385 using `null` for the same purpose.

tpetr: Either work, but I believe it's idiomatic to use `null`

kwm4385: General rule of thumb for `null` vs `undefined` in js is 

- `undefined` = This thing has not been assigned a value yet, but can have one. (generally something shouldn't go from defined -> undefined

- `null` = This has intentionally been defined as having no value. (the case here)

SingularityUI/app/reducers/form.es6
```diff
@@ -0,0 +1,28 @@
...
+    let newState = Object.assign({}, state);
+    if (!newState[formId]) {
+      newState[formId] = {};
+    }
+    newState[formId][fieldId] = newValue;
+    return newState;
+  },
+
+  CLEAR_FORM_FIELD(state, {formId, fieldId}) {
+    return MODIFY_FORM_FIELD(state, {formId, fieldId, undefined});
```
tpetr: did you mean `newValue: undefined`?

Calvinp: Yes

SingularityUI/app/components/requestForm/RequestForm.jsx
tpetr: Gotcha, I'm okay with `pure: false` as long as we implement our own `shouldComponentUpdate` that does the right thing

SingularityUI/app/controllers/RequestForm.es6
```diff
@@ -0,0 +1,40 @@
...
+      this.racksPromiseDone = true;
+      this.showViewIfReady();
+    })
+    if (requestId) {
+      const requestPromise = this.store.dispatch(RequestFetchAction.trigger(this.requestId));
+      requestPromise.then(() => {
+        this.requestPromiseDone = true;
+        this.showViewIfReady();
+      });
+    }
```
tpetr: I'd suggest simplying this by:
- removing the boolean variables
- if `requestId` is not set, set `requestPromise` to `Promise.resolved()`
- Use `Promise.all()` to wait for everything to be ready:
```javascript
Promise.all([racksPromise, requestPromise]).then(() => {
  this.setView(new RequestFormView({store: this.store}));
  app.showView(this.view);
});
```

SingularityUI/app/components/requestForm/RequestForm.jsx
Calvinp: I take it all back.
Turns out `Object.assign()` was doing a shallow clone (despite being the recommended way of obtaining a clone), and so I was actually mutating deeper references inadvertently.
I'm looking into how to do this properly now - `pure: false` may still be necessary but I'm not so sure of this anymore.

SingularityUI/app/components/common/formItems/DropDown.jsx
```diff
@@ -1,39 +1,69 @@
...
+    return (
+      <select
+        id={this.props.id}
+        className={classNames('form-control', this.props.prop.customClass)}
+        type={this.props.prop.inputType}
+        onChange={this.props.prop.updateFn}
+        value={this.props.prop.value}
+        defaultValue={this.props.prop.defaultValue}
+        required={this.props.prop.required ? true : false}>
+        {this.props.prop.forceChooseValue ? undefined : <option key={0} value='' />}{this.dropDownOpts()}
```
kwm4385: I'd use null here too

SingularityUI/app/components/requestForm/RequestForm.jsx
Calvinp: This is now fixed so that `pure: false` is unnecessary.

tpetr:
@Calvinp how does doing a deep clone make `pure: false` not necessary? the auto-generated `shouldComponentUpdate()` is still doing a shallow comparison

Calvinp:
Because I misunderstood what was causing the problem.
The reducer sets `newState[formId][fieldId] = newValue` after the clone.
But with the shallow clone, `newState[formId]` was the same object as `state[formId]`, so the old state would accidentally be mutated as well.

SingularityUI/app/components/common/formItems/FormField.jsx
```diff
@@ -4,9 +4,34 @@ import Utils from '../../../utils';
...
-        return <input className={classNames('form-control', this.props.prop.customClass)} placeholder={this.props.prop.placeholder} type={this.props.prop.inputType} id={this.props.id} onChange={this.props.prop.updateFn} value={this.props.prop.value} disabled={this.props.prop.disabled} min={this.props.prop.min} max={this.props.prop.max} required={this.props.prop.required} />;
+  componentDidMount() {
+    if (this.props.prop.generateSelectBox) {
+      const generateSelectBox = () => {
+        $(`#${this.props.id}`).select2(this.props.prop.selectBoxOptions).on('change', this.props.prop.updateFn);
+      }
+      generateSelectBox();
+      // Delay needed because componentDidMount() is called before the entire document is rendered to HTML
+      // There is no lifecycle method that runs as soon as the document is rendered as HTML...
+      setTimeout(generateSelectBox, 1);
```
wolfd: Are you sure?

Taking a look at this https://github.com/rkit/react-select2-wrapper/blob/master/src/components/Select2.js doesn't appear to need that.

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,564 @@
...
+  RUN_ONCE: [ 'killOldNonLongRunningTasksAfterMillis' ]
+};
+
+class RequestForm extends React.Component {
+
+  componentDidMount() {
+    this.props.clearForm(FORM_ID);
+  }
+
+  componentWillReceiveProps(newProps) {
```
wolfd: Can you add a comment for what this is supposed to do?
This is quite a kicker in terms of a conditional.

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,564 @@
...
+          placement="top"
+          overlay = {<ToolTip id="cannotChangeLoadBalancedAfterCreation">Option cannot be altered after creation</ToolTip>}>
+          {checkbox}
+        </OverlayTrigger>
+      );
+    } else {
+      field = checkbox;
+    }
+    return (
+      <div className = 'form-group'>
```
wolfd: I usually don't put spaces around JSX things like this.

wolfd:
Could you do a `gulp lint` and fix the errors in the changed files there before we merge?

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,564 @@
...
+        {field}
+      </div>
+    );
+  }
+
+  renderRequestTypeSpecificFormFields() {
```
wolfd: Maybe move some of these form definitions out to their own variables just so you can see how this function is rendering on a top-level without seeing the business code for what each field looks like?

SingularityUI/app/components/requestForm/RequestForm.jsx
tpetr: This isn't an idiomatic way to update the URL. Leverage the fact that dispatching the save action returns the original XHR promise:
```javascript
save(requestId) {
    dispatch(SaveAction.trigger(requestBody)).then(() => { Backbone.history.navigate(`blah`); });
}
```

SingularityUI/app/reducers/api/base.es6
```diff
@@ -13,7 +13,9 @@ export default function buildApiActionReducer(ActionGroup, initialData=[]) {
       case ActionGroup.ERROR:
         return _.extend({}, state, {
           isFetching: false,
-          error: action.error
+          error: action.error,
+          data: null,
```
tpetr: this may have unintended consequences for other reducers -- this should be an optional arg in `buildApiActionReducer()` if you truly need data to be set to null in the case of an error. i also wonder if `null` or `initialData` is a better choice here?

Calvinp: Actually, I don't even need to do that - I thought I did before I got deep copying right. I'll remove it entirely.

SingularityUI/app/components/common/formItems/FormField.jsx
kwm4385: Consider using this wrapper if select2 is a must: https://github.com/rkit/react-select2-wrapper

Calvinp: Select2 is only a must in that it's the way the form worked in the past, and we don't want to change the styling.
@tpetr if we're going to keep select2 it could be worth it to use that wrapper, but I'd be perfectly happy to drop it if there's a better alternative.

kwm4385: If replacing it is on the table, https://github.com/JedWatson/react-select is the gold standard for use with React.

tpetr: If we can swap this in easily, yeah, I'd say go with `react-select`

SingularityUI/app/actions/api/base.es6
```diff
@@ -46,9 +46,20 @@ export function buildApiAction(actionName, opts={}) {
...
+            } else {
+              return response.text().then(body => dispatch(success({response: body})));
+            }
+          } else {
+            if (response.headers.get('Content-Type') === 'application/json') {
+              return response.json().then(body => dispatch(error(body)));
+            } else {
+              return response.text().then(body => dispatch(error({message: body})));
+            }
+          }
```
tpetr: this block of code smells weird, but I couldn't put my finger on it initially... You can simplify the code here by chaining these together instead... i.e. `.then(response => response.json())` on line 49 would update to get the `Content-Type` check, and then a separate `.then()` to throw an exception that is caught lower down -- does that make sense?

SingularityUI/app/components/common/formItems/Checkbox.jsx
tpetr: Gotcha, thanks. I understand that you used `prop` to encapsulate all these settings, but it'd be simpler to have them all as individual props on the component. This something that we can address in a future PR.

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,563 @@
...
+        return this.props.form.scheduleType;
+      } else {
+        return CRON_SCHEDULE;
+      }
+    }
+  }
+
+  renderRequestTypeSelectors() {
+    const tooltip = (
+      <ToolTip id="cannotChangeRequestTypeAfterCreation">Option cannot be altered after creation</ToolTip>
```
tpetr: is this id needed? i don't see any other uses of the value

kwm4385: `id` is a required prop by react-bootstrap on Tooltip from accessibility. See https://react-bootstrap.github.io/components.html#tooltips

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -0,0 +1,563 @@
...
+        { this.shouldRenderField('instances') ? instances : null }
+        { this.shouldRenderField('rackSensitive') ? rackSensitive : null }
+        { this.shouldRenderField('hideEvenNumberAcrossRacksHint') ? hideEvenNumberAcrossRacksHint : null }
+        { this.shouldRenderField('loadBalanced') ? this.renderLoadBalanced() : null }
+        { this.shouldRenderField('waitAtLeastMillisAfterTaskFinishesForReschedule') ? waitAtLeastMillisAfterTaskFinishesForReschedule : null }
+        { this.shouldRenderField('rackAffinity') ? rackAffinity : null }
+        { this.shouldRenderField(CRON_SCHEDULE) || this.shouldRenderField(QUARTZ_SCHEDULE) ? schedule : null }
+        { this.shouldRenderField('numRetriesOnFailure') ? numRetriesOnFailure : null }
+        { this.shouldRenderField('killOldNonLongRunningTasksAfterMillis') ? killOldNonLongRunningTasksAfterMillis : null }
+        { this.shouldRenderField('scheduledExpectedRuntimeMillis') ? scheduledExpectedRuntimeMillis : null }
```
tpetr: minor nitpick but `condition && component` is a pattern i've seen that's a little more succinct than using the ternary operator

SingularityUI/app/views/requestForm.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { Provider } from 'react-redux';
+
+import RequestFormPage from '../components/requestForm/RequestForm'
+
+
+class RequestFormView extends View {
+  initialize({store, editing}) {
+    this.store = store;
+    this.editing = editing;
```
tpetr: can you explain why the explicit `editing` prop instead of basing the decision on if a request was loaded?

Calvinp: There is no logic to unload the request. So if you edit a request, then try to create a new one,  the form will think you're still editing the first one and render an edit form.
The other option is to make another reducer to explicitly unload the request (which is fine if you think it would be better, but it is a bunch of extra logic).

SingularityUI/app/actions/api/base.es6
Calvinp: I think that's what we were trying initially - but the error returned ends up being some variant of `invalid json`, and there's no way to get to the original data.

SingularityUI/app/views/requestForm.jsx
tpetr: my advice would be:
- Create a new action that clears out the request
- Update the controller to trigger this action instead of short circuiting to `Promise.resolve()`
- Rip out instances of `editing` once this is confirmed working

SingularityUI/app/actions/api/base.es6
tpetr: you were getting the invalid JSON error because you were always trying to deserialize the JSON no matter what the `Content-Type` was. What i'm saying is update the promise chain to `fetch --> create appropriate object based on Content-Type value --> check HTTP status code, throw exception w/ appropriate data if not successful --> dispatch success action`, keeping the final `catch()` at the end to dispatch the error action

Calvinp: I believe I fully understand what you meant now, and the latest is an implementation of that.

SingularityUI/app/controllers/RequestForm.es6
```diff
@@ -0,0 +1,29 @@
+import Controller from './Controller';
+import RequestFormView from '../views/requestForm';
+import { FetchAction as RequestFetchAction } from '../actions/api/request';
+import { FetchAction as RacksFetchAction } from '../actions/api/racks';
+
+class ReqeustFormController extends Controller {
```
tpetr: typo

SingularityUI/app/actions/api/base.es6
```diff
@@ -70,14 +48,46 @@ export function buildApiAction(actionName, opts = {}) {
     return { type: STARTED };
   }
 
-  function error(error) {
-    return { type: ERROR, error };
+  function error(ex) {
+    return { type: ERROR, ex };
```
tpetr: this changes the action's error key from `error` to `ex` -- was this intentional? is anything depending on it being `error` yet?

Calvinp: This was causing a linting error.
But it's still probably better to have a clear name.

SingularityUI/app/components/common/formItems/DropDown.jsx
```diff
@@ -1,39 +1,57 @@
 import React from 'react';
 import classNames from 'classnames';
 import Utils from '../../../utils';
+import select2 from 'select2';
```
tpetr: is this import still necessary?

PR #1106, opened 2016-06-21
Add UITable and Dashboard as an example of use

wolfd:
Adds a new Dashboard page.
Adds the UITable component.


SingularityUI/app/selectors/requests.es6
```diff
@@ -0,0 +1,101 @@
...
+    // to combine the data
+    // because this is O(awful)
+    return requestsAPI.data.map((r) => {
+      if (starredData.indexOf(r.request.id) > -1) {
+        return {
+          ...r,
+          starred: true
+        };
+      }
+      return r;
```
tpetr: - consider simplifying to `requestsAPI.data.map((r) => ({...r, starred: (starredData.indexOf(r.request.id) > -1)}))`
- you can store starredData as a Set to speed up lookups

SingularityUI/app/selectors/requests.es6
```diff
@@ -0,0 +1,101 @@
...
+          break;
+        case 'WORKER':
+          userRequestTotals.worker += 1;
+          break;
+        case 'RUN_ONCE':
+          userRequestTotals.runOnce += 1;
+          break;
+        case 'SERVICE':
+          userRequestTotals.service += 1;
+          break;
```
tpetr: thoughts on storing the totals using the legit requestType value (`ON_DEMAND`) instead of the camel case version (`onDemand`)? Should simplify this code a bit

SingularityUI/app/utils.es6
```diff
@@ -342,6 +339,20 @@ let Utils = {
...
+
+  maybe(object, path, defaultValue = undefined) {
+    if (path.length === 0) {
+      return object;
+    }
+    if (object.hasOwnProperty(path[0])) {
+      return Utils.maybe(
+        object[path[0]],
+        path.slice(1, path.length)
+      );
```
tpetr: RIP for loops

SingularityUI/app/selectors/requests.es6
wolfd: ES2015 ❤️

SingularityUI/app/selectors/requests.es6
wolfd: Yeah, probably a good idea, this was mostly copied over from the Coffeescript, but I agree with this.

SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -0,0 +1,325 @@
+import React, { Component, PropTypes } from 'react';
+import Waypoint from 'react-waypoint';
+import classNames from 'classnames';
+import _ from 'underscore';
```
kwm4385: `_` is already in the global scope atm

SingularityUI/app/components/deployDetail/DeployDetail.jsx
```diff
@@ -122,7 +124,7 @@ class DeployDetail extends React.Component {
                 <td>{Utils.timeStampFromNow(data.taskId.startedAt)}</td>
                 <td>{Utils.timeStampFromNow(data.updatedAt)}</td>
                 <td className="actions-column"><a href={`${config.appRoot}/request/${data.taskId.requestId}/tail/${config.finishedTaskLogPath}?taskIds=${data.taskId.id}`} title="Log">&middot;&middot;&middot;</a></td>
-                <td className="actions-column"><JSONButton object={data} text="{ }" /></td>
+                <td className="actions-column"><JSONButton object={data}>{'{ }'}</JSONButton></td>
```
kwm4385: `{'{ }'}` == `{ }`

SingularityUI/app/utils.es6
wolfd: 👻

SingularityUI/app/reducers/ui/starred.es6
```diff
@@ -0,0 +1,29 @@
...
+    let newState;
+    if (state.indexOf(action.value) > -1) {
+      // remove
+      newState = state.filter((requestId) => requestId !== action.value);
+    } else {
+      newState = [...state, action.value];
+    }
+    // This part is the bad part
+    // persist the starredRequests
+    window.localStorage.starredRequests = JSON.stringify(newState);
```
kwm4385: The localstorage stuff should definitely happen in the action instead of the reducer. Just pass the object in via `action.data` just like the api calls.

SingularityUI/app/components/common/table/UITable.jsx
wolfd: I like being explicit with my imports, but if you think there's some negative effects to this, I can remove it.

SingularityUI/app/reducers/ui/starred.es6
wolfd: This is true, there's a bit more of things like this that we need to fix up (from the log tailer), but I think it requires more than just putting it in the action creator, because it also has to persist the state.

SingularityUI/app/components/common/table/UITable.jsx
tpetr: Kevin is correct but i think this is OK -- i'd eventually like to remove all of the globally scoped modules from the webpack config

kwm4385: Eventually I'd like to replace underscore with lodash as well. Perhaps at the same time.

SingularityUI/app/reducers/ui/starred.es6
kwm4385: What I mean is do all the same calls to localstorage that you're using here but do it in the action. Then you can dispatch with `data` being the same as what you stored.

SingularityUI/app/components/common/JSONButton.jsx
```diff
@@ -45,15 +54,11 @@ export default class JSONButton extends React.Component {
             </div>
           </Modal.Body>
           <Modal.Footer>
-            <Button bsStyle="default" className="copy-btn" data-clipboard-text={JSON.stringify(this.props.object, null, 2)}>Copy</Button>
-            <Button bsStyle="info" onClick={this.hideJSON.bind(this)}>Close</Button>
+            <Button bsStyle='default' className='copy-btn' data-clipboard-text={JSON.stringify(this.props.object, null, 2)}>Copy</Button>
```
andyhuang91: fyi: https://github.com/HubSpot/javascript/tree/master/react#quotes

SingularityUI/app/components/common/ModalButton.jsx
```diff
@@ -0,0 +1,67 @@
+import React, { Component, PropTypes } from 'react';
+import Modal from 'react-bootstrap/lib/Modal';
+import Button from 'react-bootstrap/lib/Button';
+
+export default class ModalButton extends Component {
```
andyhuang91: Based on what I think this is doing, ConfirmModal might be a better name than ModalButton

SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -0,0 +1,325 @@
...
+  
+  static defaultProps = {
+    paginated: false,
+    rowChunkSize: 30,
+    sortBy: undefined,
+    sortDirection: UITable.SortDirection.DESC,
+    asyncSort: false
+  };
+
+  state;
```
andyhuang91: I'm curious what this does

SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -0,0 +1,325 @@
...
+
+  constructor(props) {
+    super(props);
+    this.displayName = 'UITable';
```
andyhuang91: Is setting displayName necessary? I thought this was done automatically

SingularityUI/app/components/dashboard/DashboardPage.jsx
```diff
@@ -0,0 +1,199 @@
...
+    return myRequests;
+  }
+
+  renderMyPausedRequests() {
+    let pausedRequestsSection = (
+      <div className='empty-table-message'><p>No paused requests</p></div>
+    );
+
+    const pausedRequests = this.props.userRequests.filter((r) => r.state === 'PAUSED');
+    const starredRequests = this.props.requests.filter((r) => 'starred' in r && r.starred);
```
andyhuang91: what is this crazy thing? I think you can just do `(r) => r.started`

SingularityUI/app/components/dashboard/DashboardPage.jsx
```diff
@@ -0,0 +1,199 @@
...
+import * as RequestActions from '../../actions/api/request';
+import * as RequestsSelectors from '../../selectors/requests';
+
+import UITable from '../common/table/UITable';
+import { Starred, RequestId, Type, LastDeploy, DeployUser, State, Actions } from '../requests/Columns';
+
+import RequestCounts from './RequestCounts';
+import RequestCount from './RequestCount';
+
+class DashboardPage extends Component {
```
andyhuang91: I would recommend breaking this up into separate components and many doing the container / dumb component pattern here.

SingularityUI/app/components/dashboard/RequestCount.jsx
```diff
@@ -0,0 +1,31 @@
...
+class RequestCount extends Component {
+  static propTypes = {
+    label: PropTypes.string.isRequired,
+    count: PropTypes.number.isRequired,
+    link: PropTypes.string
+  };
+
+  constructor(props) {
+    super(props);
+    this.displayName = 'RequestCount';
```
andyhuang91: You can do this as a pure function: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md

SingularityUI/app/components/requests/Columns.jsx
```diff
@@ -0,0 +1,158 @@
...
+
+import Utils from '../../utils';
+
+import JSONButton from '../common/JSONButton';
+
+import UnpauseButton from './UnpauseButton';
+import RemoveButton from './RemoveButton';
+
+// use this only with combineStarredWithRequests selector
+export const Starred = (changeStar, sortable = false) => (
```
andyhuang91: Did you mean to use
```
export const Starred = ({changeStar, sortable}) => (
```
and defaultProps? I assume changeStart and sortable are properties for this react component.

SingularityUI/app/reducers/ui/starred.es6
andyhuang91: 👍 for reducers shouldn't have any side effects

SingularityUI/app/utils.es6
```diff
@@ -342,6 +339,20 @@ let Utils = {
     if (response.status === 400) {
       return app.caughtError();
     }
+  },
+
+  maybe(object, path, defaultValue = undefined) {
+    if (path.length === 0) {
```
andyhuang91: Prefer `if (!path.length)`? https://github.com/HubSpot/javascript#comparison--shortcuts

SingularityUI/app/components/common/table/UITable.jsx
wolfd: Nah, good catch

SingularityUI/app/components/dashboard/DashboardPage.jsx
wolfd: yeah good point

SingularityUI/app/components/common/JSONButton.jsx
wolfd: :/ I guess that makes sense.

SingularityUI/app/components/common/table/UITable.jsx
tpetr: me too

tpetr:
LGTM. To address in future PR(s):
- Linting
- User request #'s (they don't match what's in master, but these numbers have always been a little buggy. this could be something we solve or get rid of for good in the redesign)

PR #1109, opened 2016-06-24
allow resource override to be set on a run now request

ssalinas:
For one-offs that have varying usage, it can be useful to specify resources for a specific run of the request. This adds an optional `Resources` field to the `SingularityRunNowRequest` that will override the resources set on the deploy.

/cc @gchomatas @tpetr


ssalinas:
Merging this for Blazar folks to use (/cc @jonathanwgoodwin )

PR #1110, opened 2016-06-27
Add 'extraScript' field to UI configuration for extra analytics/etc

wolfd:
Also update package.json and pom for SingularityUI to use Node 6

### DIFF
--- a/SingularityUI/package.json
+++ b/SingularityUI/package.json
@@ -8,12 +8,12 @@
     "url": "git@github.com:HubSpot/Singularity.git"
   },
   "engines": {
-    "node": "~4.1.2",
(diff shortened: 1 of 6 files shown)


SingularityUI/package.json
```diff
@@ -8,12 +8,12 @@
...
   "engines": {
-    "node": "~4.1.2",
-    "npm": "~2.14.16"
+    "node": "~6.2.1",
+    "npm": "~3.9.3"
   },
   "bin": {},
   "scripts": {
-    "start": "brunch watch --server"
+    "start": "gulp serve"
```
tpetr: can't believe we missed this 👍

PR #1112, opened 2016-06-27
New new deploy page (WIP)

Calvinp:
This is a mostly working version of the react/redux new deploy page.
It's incredibly complicated, and I'm sure there are plenty of bugs I don't know about on top of the ones I do. Please take a close look.
Known bugs that I'm working on:
- [x] The implementation of default values for Select is bad. It was the only way I could get it to work, but it's not a real solution.
- [x] Sometimes an error comes back from the server as `[object object]`
- [x] Can't submit fields that are `key=value` form (the server wants a map, but gets an array of strings)
- [x] Some fields, such as CPUs, need defaults and don't have them
- [x] No client-side validation of fields (eg. making sure numeric fields are actually numbers).
- [x] No explanation to user for why they can't submit the form - this is a huge form, so if one field has a problem it'll take a long time to find that one field.
- [x] It might be better to put the form group components into their own files considering how long that file already is.
- [x] Lots of linting errors

cc @tpetr @wolfd @kwm4385


SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+import RemoveButton from '../common/RemoveButton';
+import Select from 'react-select';
+import { modifyField, clearForm } from '../../actions/form';
+import {SaveAction} from '../../actions/api/deploy';
+import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
+import ToolTip from 'react-bootstrap/lib/Tooltip';
+import Utils from '../../utils';
+import classNames from 'classnames';
+
+class SelectFormGroup extends Component {
```
kwm4385: I would definitely split this file up, starting with giving each component its own file. Also the constants can be grouped in separate files.

Calvinp: I'm planning on making much more use of the constants, so that makes a lot of sense.

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+  getValueOrDefault(fieldId) {
+    return this.getValue(fieldId) || FLATTENED_FIELDS[fieldId].default;
+  }
+
+  isRequestDaemon() {
+    return ['SERVICE', 'WORKER'].indexOf(this.props.request.request.requestType) !== -1;
+  }
+
+  // Returns true unless the object is falsey or an empty array.
+  hasValue(value) {
```
wolfd: An empty array is already a falsey value.

This is essentially a no-op.

SingularityUI/app/components/common/RemoveButton.jsx
```diff
@@ -0,0 +1,22 @@
...
+  };
+
+  render() {
+    return (
+      <Button
+        className='remove-button'
+        id={this.props.id}
+        onClick={(event) => {
+          event.preventDefault();
+          this.props.onClick();
```
kwm4385: This can be shortened to `onClick={this.props.onClick}`

SingularityUI/app/views/newDeployForm.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import View from './view';
+
+import { Provider } from 'react-redux';
+
+import NewDeployFormPage from '../components/newDeployForm/NewDeployForm'
+
+class NewDeployFormView extends View {
```
kwm4385: Should be using `ReactView`

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+    if (this.getValueOrDefault('executorType') === DEFAULT_EXECUTOR_TYPE) {
+      this.addFieldsToDeployObject(deployObject, FIELDS.defaultExecutor);
+    } else {
+      this.addFieldsToDeployObject(deployObject, FIELDS.customExecutor);
+    }
+    deployObject.requestId = this.props.request.request.id;
+    this.props.save({deploy: deployObject});
+  }
+
+  addThingToArrayField(fieldId, thing) {
```
wolfd: What's a `thing`?

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
andyhuang91: can you use _.isEmpty?

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+          id = "artifacts"
+          value = {this.getValue('uris') || []}
+          onChange = {(newValue) => this.updateField('uris', newValue)}
+          placeholder="eg: http://s3.example/my-artifact"
+        />
+      </div>
+    );
+    return (
+      <div>
+        <fieldset id="default-expandable" className='expandable'>
```
wolfd: Run this through `eslint` (not `gulp lint`, the plugin is broken for inherited configuration for some odd reason). The className should be double quotes.

SingularityUI/app/components/common/formItems/MultiInput.jsx
```diff
@@ -0,0 +1,53 @@
...
+      valueClone.push(value);
+    } else {
+      valueClone[key] = value;
+    }
+    this.props.onChange(_.without(valueClone, ""));
+  };
+
+  render() {
+    const valueClone = this.props.value.slice()
+    if (valueClone.length === 0 || _.last(valueClone)) {
```
zdhickman: You can use the shorthand `!valueClone.length`

SingularityUI/app/components/common/formItems/MultiInput.jsx
```diff
@@ -0,0 +1,53 @@
...
+      <div id={this.props.id} className={this.props.className}>
+        {
+          valueClone.map((value, key) => {
+            return (
+              <FormField
+                key = {key}
+                prop = {{
+                  value: value,
+                  updateFn: event => this.change(key, event.target.value),
+                  /* required: this.props.required, Can't mark this as form required because there will always be an extra empty field */
```
zdhickman: Probably should not leave code commented out

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+        <fieldset>
+          {unpauseOnSuccessfulDeploy}
+        </fieldset>
+      </div>
+    );
+
+    const errorMessage = (
+      this.props.saveApiCall.error ?
+        <p className='alert alert-danger'>
+          There was a problem saving your request: {this.props.saveApiCall.error.message}
```
kwm4385: You can use `boolean && <Component/>` here instead of a ternary with null

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+        {hostPath}
+        {mode}
+      </div>
+    );
+  }
+
+  renderDockerVolumes() {
+    const volumes = this.getValue('volumes');
+    if (!volumes) {
+      return (<div id="docker-volumes" />);
```
kwm4385: ids should be avoided (this is likely an artifact from the old template/view)

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+
+  componentDidMount() {
+    this.props.clearForm(FORM_ID);
+  }
+
+  updateField(fieldId, newValue) {
+    this.props.update(FORM_ID, fieldId, newValue);
+  }
+
+  getValue(fieldId) {
```
wolfd: Is it worth having this?

I don't see any prop validation here, so I don't know if form is a required prop, but I think it should be. Then this field can be removed.

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1303 @@
...
+}
+
+const FLATTENED_FIELDS = _.extend(
+  {},
+  makeFlattenedFields(FIELDS.all),
+  makeFlattenedFields(FIELDS.customExecutor),
+  makeFlattenedFields(FIELDS.defaultExecutor)
+);
+
+class NewDeployForm extends Component {
```
wolfd: Add prop validation

Calvinp:
Fixed a bunch of the bugs, now I'll get to responding to PR feedback.

SingularityUI/app/components/common/RemoveButton.jsx
```diff
@@ -0,0 +1,19 @@
+import React, { Component, PropTypes } from 'react';
+import Button from 'react-bootstrap/lib/Button';
+import Glyphicon from 'react-bootstrap/lib/Glyphicon';
+
+export default class RemoveButton extends Component {
```
tpetr: `<Button className='remove-button'` isn't many more characters than `<RemoveButton`, but adds the burden of being yet another component to be aware of -- is it really necessary?

SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -0,0 +1,1423 @@
...
+      this.updateField(fieldId, fieldValue);
+    }
+  }
+
+  addObjectToDeployFieldPreventDefault(fieldId, deployObject, event) {
+    event.preventDefault();
+    this.addDeployObjectToArrayField(fieldId, deployObject);
+  }
+
+  removeThingFromArrayField(fieldId, key) {
```
wolfd: I see that you changed `addThing...` to `addDeployObjectToArrayField`, are these functions related? Are these supposed to be `deployObject`s too?

Calvinp: So the things that can get added are `artifact`s, `portMapping`s, and `volume`s. I don't necessarily think `deployObject` is the best name for that category of things, but I haven't been able to think of a better one. Maybe `SubField`?

PR #1113, opened 2016-06-27
Add sortData to Column for UITable. Fix some UITable issues

wolfd:
Pretty simple stuff.

In the future I want to fix the `setState` issues (table not existing because of async problems) with a cancellable Promise, but that's more work than it needs to be right now.

### DIFF
--- a/SingularityUI/app/components/common/table/UITable.jsx
+++ b/SingularityUI/app/components/common/table/UITable.jsx
@@ -144,7 +144,7 @@ class UITable extends Component {
         rowData
       );
 
-      return <td key={`${tdIndex}${col.props.id}`} className={col.props.className}>{cell}</td>;
+      return <td key={col.props.id} className={col.props.className}>{cell}</td>;
     });
     return <tr key={`row-${this.props.keyGetter(rowData)}`}>{row}</tr>;
   }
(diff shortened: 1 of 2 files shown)


SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -214,9 +215,11 @@ class UITable extends Component {
             onEnter={() => {
               const maxVisibleRows = this.state.chunkNum * this.props.rowChunkSize;
               if (maxVisibleRows < this.state.data.length) {
-                this.setState({
-                  chunkNum: this.state.chunkNum + 1
-                });
+                setTimeout(() => {
```
kwm4385: You can use _.defer() here instead of a 0 timeout.
http://underscorejs.org/#defer

SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -142,7 +144,7 @@ class UITable extends Component {
         rowData
       );
 
-      return <td key={tdIndex} className={col.props.className}>{cell}</td>;
+      return <td key={col.props.id} className={col.props.className}>{cell}</td>;
```
tpetr: are we worried about dupes here? do we have any way to enforce this ID being unique?

wolfd: I could combine them?

tpetr: not sure we necessarily want to combine them -- the whole point of `key` is to help React optimize the rendering (i.e. reordered or destroyed instead of all of them being clobbered). I'm OK with the original approach as long as we have some sort of note somewhere that `props.id` should be unique

wolfd: Good point. Yeah, props.id actually needs to be unique for a lot of reasons. I'll put a note in.

tpetr:
LGTM

PR #1116, opened 2016-06-30
Link to tail_of_finished_service.log

Calvinp:
If a file isn't found, and the file we were looking for was named `service.log`, there will now be a link to `tail_of_finished_service.log`

If the names are too specific, they can be moved to config.


Calvinp:
This doesn't work as well as I thought it did - it will show the link if the log is gone when you load the page, but if it disappears as you're tailing it can't detect that. I'm working on fixing this.

SingularityUI/app/components/logs/FileNotFound.jsx
```diff
@@ -1,8 +1,36 @@
-import React from 'react';
+import React, {PropTypes} from 'react';
+
+const RENAMED_FILES = {
+  'service.log': 'tail_of_finished_service.log'
```
kwm4385: These filenames aren't constant, but configurable. You can get the values from `config.runningTaskLogPath` and `config.finishedTaskLogPath`.

Calvinp: Interesting. That's misconfigured locally, but it is correct in the deployed environments.

SingularityUI/app/actions/log.es6
```diff
@@ -197,7 +197,11 @@ export const taskGroupFetchNext = taskGroupId =>
       }
     });
 
-    return Promise.all(promises).then(() => dispatch({taskGroupId, type: 'LOG_REQUEST_END'}));
+    return Promise.all(promises).then(() => dispatch({taskGroupId, type: 'LOG_REQUEST_END'})).catch((error) => {
+      if (error.status === 404) {
+        dispatch(taskFileDoesNotExist(taskGroupId, error.responseText.split('task ID ')[1]));
```
Calvinp: `error.responseText.split('task ID ')[1]` This feels bad to me. I'm trying to figure out a better way to get the task id.

SingularityUI/app/components/logs/FileNotFound.jsx
kwm4385: Yeah you need to set your env variables correctly when you build. You can look at the jenkins configs to see the correct values.

Calvinp:
Should be good to go now @tpetr @kwm4385 @wolfd

SingularityUI/app/actions/log.es6
```diff
@@ -183,21 +183,26 @@ export const taskGroupFetchNext = taskGroupId =>
...
             nextOffset = offset + data.length;
             return dispatch(taskData(taskGroupId, taskId, data, offset, nextOffset, true, maxLines));
           }
         });
-      } else {
-        return Promise.resolve(); // reject("initialDataLoaded is false for task #{taskId}")
+        promise.taskId = taskId;
+        return promise;
       }
+      return Promise.resolve(); // reject("initialDataLoaded is false for task #{taskId}")
```
wolfd: Can we get rid of this comment?

SingularityUI/app/actions/log.es6
```diff
@@ -174,7 +174,7 @@ export const taskGroupFetchNext = taskGroupId =>
   function(dispatch, getState) {
     let {tasks, taskGroups, logRequestLength, maxLines} = getState();
 
-    let taskGroup = taskGroups[taskGroupId];
+    const taskGroup = taskGroups[taskGroupId];
```
wolfd: :D

SingularityUI/app/actions/log.es6
Calvinp: The comment appears in a few other places in that action. I don't know the original purpose, and I'd rather not delete it in one place without deleting it in the other places (unless it makes sense to).
cc @tpetr

SingularityUI/app/actions/log.es6
Calvinp: Yeah, I tried to go in and fix the linting errors around the code that I was modifying. I didn't get all of them though. For example, the one on line 175 there is strange. `tasks` is modified later, but the other three items in the let are not. So it doesn't like that it's `let`, because three items aren't changed later. But it can't be `const` because one is. And using a mixture would involve repeating the `getState()` call for no other reason.

wolfd: You could store the state in a const, and then two destructures to do that.

Calvinp:
From meeting, todos:

- [ ] Don't display an error when the file disappears
- [ ] Check if the new file actually exists, so it can be stated that it definitely does.

SingularityUI/app/actions/log.es6
```diff
@@ -170,34 +193,54 @@ export const updateTaskStatus = (taskGroupId, taskId) =>
...
           }
-        });
-      } else {
-        return Promise.resolve(); // reject("initialDataLoaded is false for task #{taskId}")
+          return Promise.resolve();
+        }).error(error => Utils.ignore404(error));
+        promise.taskId = taskId;
+        return promise;
       }
+      return Promise.resolve(); // reject("initialDataLoaded is false for task #{taskId}")
```
wolfd: let's get rid of this comment

SingularityUI/app/components/logs/FileNotFound.jsx
```diff
@@ -1,8 +1,33 @@
-import React from 'react';
+import React, {PropTypes} from 'react';
```
wolfd: I've been doing spaces in this one, we don't enforce that with linting, but I think it looks nicer here.

SingularityUI/app/components/logs/TaskGroupContainer.jsx
```diff
@@ -3,23 +3,43 @@ import TaskGroupHeader from './TaskGroupHeader';
 import LogLines from './LogLines';
 import LoadingSpinner from './LoadingSpinner';
 import FileNotFound from './FileNotFound';
-import classNames from 'classnames';
+import {doesFinishedLogExist} from '../../actions/log';
```
wolfd: same for this one, just to align with the pattern we've been using

SingularityUI/app/components/logs/TaskGroupContainer.jsx
```diff
@@ -46,17 +71,21 @@ let mapStateToProps = function (state, ownProps) {
...
+    finishedLogExists: _.any(_.pluck(tasks, 'taskFinishedLogExists')),
     terminated: _.all(_.pluck(tasks, 'terminated')),
-    path: state.path
+    path: state.path,
+    taskIds: taskGroup.taskIds
   };
 };
 
-export default connect(mapStateToProps)(TaskGroupContainer);
+const mapDispatchToProps = { doesFinishedLogExist };
```
wolfd: We haven't been using this pattern for dispatches so far, I didn't realize that you could do this. Should we stick with the pattern we've been using (mapDispatchToProps as a function that takes in `(dispatch, ownProps)`) or is this better for some reason?

wolfd: This is apparently been a pattern in this log code, which makes sense given that it was some of the first code we wrote for redux/react. We can fix this later when this whole thing get's redone.

Calvinp: This is a pattern being used all over the logs page, though not in the rest of the project.

wolfd:
LGTM

PR #1117, opened 2016-06-30
Refactor redux actions to mimic their layout in the actual API

wolfd:
Also bring in a change to Request starring. Not the final change for that, but an update.
Also Util helpers.


tpetr:
LGTM, thanks

PR #1118, opened 2016-07-01
Always launch docker containers in the correct parent cgroup

PaulFurtado:
Currently, the docker executor script hardcodes the parent cgroup of a docker container to `mesos/$CGROUPS_GUID`. This change makes it instead discover the cgroup using `/proc/self/cgroup` per the reccommendation of: https://www.freedesktop.org/wiki/Software/systemd/PaxControlGroups/

This change also removes the code for setting `memory.use_hierarchy=1` because it is ineffective (you can see the warnings in the logs on our mesos slaves). In order to set that property, it must be set on the root cgroup, before any child cgroups are created. My internal changes to cgroups on our mesos slaves takes care of that and all child cgroups inherit the property.

@ssalinas @tpetr Going to try to test this myself tonight, might need your help with the Singularity bits tomorrow. I tested that function pretty thoroughly so I expect this to just work.

### DIFF
--- a/SingularityExecutor/src/main/resources/docker.sh.hbs
+++ b/SingularityExecutor/src/main/resources/docker.sh.hbs
@@ -40,9 +40,9 @@ function handle_signal {
   exit "$exit_code"
 }
 
-# Outputs the path of the deepest cgroup a PID (or self) is in.
+# Outputs the path of the deepest cgroup the current process is in.
 # The output of:
-#   cat /proc/4194/cgroup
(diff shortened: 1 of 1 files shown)


ssalinas:
👍 I can test this further. Nitpick, but can we update the examples to be non-hubspot-specific? Other than that looks good, thanks for the cgroup expertise!

PaulFurtado:
@ssalinas :+1:  Got rid of the reference to HubSpot in the example

PR #1119, opened 2016-07-01
Request form fixes

Calvinp:
Makes the following changes in the New/Edit Request Form:
- Fix the Owners field - now it's a multi-input instead of a multi-select. Much more appropriate for the field type.
- Fix many, many linting errors
- Use the same field metadata style as the New Deploy form
- Validate fields (eg. number fields fail if they can't be parsed as an int)
- Feedback to user on which fields are preventing the form from being submitted
- Use FormGroup components instead of writing out the form groups


Calvinp:
@tpetr @wolfd @kwm4385 This is now ready for you to take a look (and the diff is nice).

tpetr:
LGTM

PR #1120, opened 2016-07-01
Sort task requests by priority order

ssalinas:
@tpetr one addition to the priority PR, sort task requests by priority order first, then by task start so that higher priority tasks get resources first


ssalinas:
@tpetr updated this a bit. Previously we were sorting pending tasks based on the last request start time, which doesn't make sense anymore with priorities in place. This updates the sort to be based on the priority and task overdue time and a configurable parameter to tune the weighting. Formula for a weight is: `(time overdue in ms) * (priority) ^ (configurable parameter)

tl;dr of that formula being that a `0` will sort just on overdue time, and the higher you set the parameter, the more it will weight the priority vs the overdue time. Default is `1`, so your weight is just `(overdue) * (priority)`

ssalinas:
merging this into the priority-level branch since it is only only as far as hs_qa right now

PR #1121, opened 2016-07-01
React/redux Task Search

kwm4385:
:partyparrot:

@tpetr @wolfd @Calvinp


SingularityUI/app/actions/api/taskHistory.es6
```diff
@@ -0,0 +1,19 @@
...
+  if (deployId) params.push(`&deployId=${deployId}`);
+  if (host) params.push(`&host=${host}`);
+  if (lastTaskStatus) params.push(`&lastTaskStatus=${lastTaskStatus}`);
+  if (startedAfter) params.push(`&startedAfter=${startedAfter}`);
+  if (startedBefore) params.push(`&startedBefore=${startedBefore}`);
+  if (orderDirection) params.push(`&orderDirection=${orderDirection}`);
+  return {
+    url: `/history/request/${requestId}/tasks?count=${count}&page=${page}${params.join('')}`
+  }
+});
```
tpetr: can we update this to match the API layout that @wolfd rolled out?

SingularityUI/app/components/taskSearch/TasksTable.jsx
```diff
@@ -0,0 +1,91 @@
...
+    //       </Table>
+    //     </div>
+    //   );
+    // } else {
+    //   return (
+    //     <div className="text-center">
+    //       <div className="page-loader"></div>
+    //     </div>
+    //   );
+    // }
```
tpetr: commented code blocks 🙅🏻

SingularityUI/app/reducers/api/index.es6
```diff
@@ -50,11 +50,13 @@ import {
...
 
+import { FetchAction as TaskHistory } from '../../actions/api/taskHistory';
+
 const user = buildApiActionReducer(FetchUser);
 const webhooks = buildApiActionReducer(FetchWebhooks, []);
 const slaves = buildApiActionReducer(FetchSlaves, []);
 const racks = buildApiActionReducer(FetchRacks, []);
-const request = buildKeyedApiActionReducer(FetchRequest);
+const request = buildApiActionReducer(FetchRequest);
 const saveRequest = buildApiActionReducer(SaveRequest);
```
tpetr: will this affect other pages? (@wolfd?)

kwm4385: I don't think so, looking at the other references. I believe this was a merge error.

SingularityUI/app/actions/api/history.es6
```diff
@@ -37,6 +37,22 @@ export const FetchDeployForRequest = buildApiAction(
...
+    if (requestId) params.push(`&requestId=${requestId}`);
+    if (deployId) params.push(`&deployId=${deployId}`);
+    if (host) params.push(`&host=${host}`);
+    if (lastTaskStatus) params.push(`&lastTaskStatus=${lastTaskStatus}`);
+    if (startedAfter) params.push(`&startedAfter=${startedAfter}`);
+    if (startedBefore) params.push(`&startedBefore=${startedBefore}`);
+    if (orderDirection) params.push(`&orderDirection=${orderDirection}`);
+    return {
+      url: `/history/tasks?count=${count}&page=${page}${params.join('')}`
+    };
```
tpetr: we may want to use a function for building query params (like `$.param()`, but I'd rather not depend on jQuery), since it will handle escaping, etc. for us

Calvinp:
You can delete [DateEntry](https://github.com/HubSpot/Singularity/blob/new_task_search/SingularityUI/app/components/common/formItems/DateEntry.jsx) since you pulled out the only place where it was used.
If #1119 is merged before this, you can also delete [DropDown](https://github.com/HubSpot/Singularity/blob/new_task_search/SingularityUI/app/components/common/formItems/DropDown.jsx) - otherwise that PR can do it.

kwm4385:
@Calvinp Deleted `DateEntry`. I'll hold off on `DropDown` for now.

Calvinp:
```
BOSMM2NGH0JG3QD:app cpomerantz$ eslint controllers/TaskSearch.es6 views/taskSearch.jsx components/taskSearch/* components/common/formItems/ReduxSelect.jsx
...
✖ 82 problems (82 errors, 0 warnings)
```

Calvinp:
Found another file you can delete: [LinkedFormItem](https://github.com/HubSpot/Singularity/blob/new_task_search/SingularityUI/app/components/common/formItems/LinkedFormItem.jsx)

SingularityUI/app/components/taskSearch/TaskSearch.jsx
```diff
@@ -1,166 +1,181 @@
...
+    super(props);
+    this.state = {
+      filter: {
+        requestId: props.requestId,
+        page: 1,
+        count: TaskSearch.TASKS_PER_PAGE
+      },
+      disableNext: false,
+      loading: false
+    };
```
tpetr: would be preferable to have this in redux

kwm4385: I can see the argument for having this in redux, but we should think about where we draw the line between component state and redux state.

tpetr: Long term, task search results should be true permalinks (i.e. have all this state data in the URL), and that would make for a good argument to move this stuff in to Redux (I'm using https://github.com/reactjs/redux/issues/1287#issuecomment-175351978 as a guide). But since permalinks don't exist in the original implementation, I'm okay with circling back on this.

kwm4385: In the context of permalinks, that does make sense. Sounds like a plan 👍

kwm4385:
If anyone here see any reason why these should two branches not be united, speak now or forever hold thy peace.

PR #1122, opened 2016-07-05
Fix machines pages linting errors

Calvinp:
Fixes linting errors in Racks and Slaves views and React pages. The controllers and Actions didn't have linting errors.

Unfortunately the diff isn't pretty because indentation fixes took over.

### DIFF
--- a/SingularityUI/app/components/machines/MachinesPage.jsx
+++ b/SingularityUI/app/components/machines/MachinesPage.jsx
@@ -1,57 +1,57 @@
 import React from 'react';
 import OldTable from '../common/OldTable';
 
-let MachinesPage = React.createClass({
+const MachinesPage = React.createClass({
 
-    propTypes: {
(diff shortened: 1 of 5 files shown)


Calvinp:
The racks and slaves pages have a problem (not caused by the changes in this PR, but this could be a good place to fix it). The buttons to freeze/remove/decommission/reactivate slaves and racks are all broken. Each and every one of them. They throw an error `Uncaught TypeError: slaveModel.promptDecommission is not a function`.

tpetr:
LGTM

Calvinp:
So I'll fix the buttons in a different PR then.

tpetr:
either this PR or a separate one -- whichever makes the most sense to you. thanks!

Calvinp:
Thinking about it, it might be a large change. So go ahead and merge this PR, and I'll open a new one for that once it's implemented.

PR #1123, opened 2016-07-05
Fix webhook linting errors

Calvinp:
Fixes almost all linting errors in view, controller, action and component for webhooks.


SingularityUI/app/components/webhooks/Webhooks.jsx
```diff
@@ -10,206 +10,245 @@ import NewWebhookForm from './NewWebhookForm';
...
+  },
+
+  promptNewWebhook() {
+    const newWebhook = (uri, type) => this.newWebhook(uri, type);
+    return vex.dialog.open({
+      message: "<div class='new-webhook' />",
+      afterOpen: () => {
+        this.validateInput = input => {
+          try {
+            new URL(input);
```
Calvinp: This is a linting error because URL isn't assigned to anything, so it's used for its side effect (of throwing an error if the URL provided is bad).

I didn't want to change the functionality, but there's no other way to fix the linting error here.

zdhickman: Can you achieve the same effect without linting errors by removing the `new`?

Calvinp: No - that causes the following error: `TypeError: Failed to construct 'URL': Please use the 'new' operator, this DOM object constructor cannot be called as a function.`

Calvinp:
There's another problem with this page (it was there already, not caused by the changes in this PR, but this could be a good place to fix it). When a webhook is created or deleted, the page attempts to refresh by calling `this.props.collections.webhooks.fetch()` - but that's not a function anymore.

Calvinp:
That problem is now fixed.

tpetr:
LGTM

PR #1124, opened 2016-07-05
Fix linting errors in tasks table page

Calvinp:
Fixes the linting errors in the components of the Tasks table. (There were no linting errors in the view, action, or controller).


SingularityUI/app/components/tasks/Columns.jsx
```diff
@@ -116,26 +116,26 @@ export const Memory = (
...
-  return <Column
-    label=""
-    id="actions"
-    key="actions"
-    className="actions-column"
-    cellData={
-      (rowData) => rowData
-    }
-    cellRender={(cellData) => {
+  return (
```
tpetr: is the return here part of our style guide? i dont necessarily have an issue with it, but am curious if it's there for an explicit reason

tpetr:
LGTM aside from question above

SingularityUI/app/components/tasks/Columns.jsx
Calvinp: It's not - it would be perfectly fine to get rid of it entirely.

Calvinp:
I'm going to merge this since the above question has been answered.

PR #1129, opened 2016-07-07
Redux API auto refresh

wolfd:
Lifecycle hooks:
```
componentDidMount() {
  this.props.refresh();
}

componentWillUnmount() {
  this.props.cancelRefresh();
}
```

`mapDispatchToProps`:
```
refresh: () => dispatch(RefreshActions.BeginAutoRefresh('RequestDetailPage', [
  FetchRequest.trigger(ownProps.requestId),
  FetchActiveTasksForRequest.trigger(ownProps.requestId),
  FetchTaskCleanups.trigger()
], 10000)),
cancelRefresh: () => dispatch(
  RefreshActions.CancelAutoRefresh('RequestDetailPage')
)
```

### DIFF
--- a/SingularityUI/app/actions/ui/refresh.es6
+++ b/SingularityUI/app/actions/ui/refresh.es6
@@ -2,11 +2,17 @@ export const CANCEL_AUTO_REFRESH = 'CANCEL_AUTO_REFRESH';
 
 export const CancelAutoRefresh = (key, isExpired = false) => {
   return (dispatch, getState) => {
-    const intervalId = getState().ui.refresh[key].intervalId;
+    const state = getState();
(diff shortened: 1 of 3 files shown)


tpetr:
do you mean `mapDispatchToProps` for the 2nd bit?

tpetr:
why not create an `autoRefresh()` decorator a la redux-form to cut down on copypasta?

wolfd:
The decorator is a good idea, but I think we should hold off on it for now. It's not that hard to do it this way, and we'll have a better idea of how to move forward with a decorator once we use this a few times.

tpetr:
LGTM once we remove the action data from the store. 👍 for circling back on the decorator.

PR #1132, opened 2016-07-11
Healthcheck message improvements

Calvinp:
This changes the Healthcheck alert on the task detail page. 
screenshot 2016-07-11 10 09 45
In the top alert, the list of other tasks that lead to the deploy failure is gone. Instead there is just a link to the deploy detail page, where that list can be seen if needed.
In the lower alert:
- Formatting is improved for clarity.
- Milliseconds are replaced by seconds and rounded to the nearest tenth.
- Links to the logs and healthcheck details are placed on their own lines to make the options more obvious.

This also fixes various bugs in Task Detail alerts, and adds proptype validation.


tpetr:
LGTM

PR #1135, opened 2016-07-11
surface taskReconciliationStartedAt in SingularityState object

tpetr:
Explicit reconciliation can sometimes cause delays in Mesos offers. This PR adds a `taskReconciliationStartedAt` field to SingularityState, which makes it easier to understand when a reconciliation run is taking place.


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
```diff
@@ -151,6 +160,8 @@ private void checkReconciliation(final SchedulerDriver driver, final long reconc
     if (taskStatuses.isEmpty()) {
       LOG.info("Task reconciliation ended after {} checks and {}", numTimes, JavaUtils.duration(reconciliationStart));
 
+      stateManager.saveTaskReconciliationStatistics(new SingularityTaskReconciliationStatistics(reconciliationStart, System.currentTimeMillis() - reconciliationStart, numTimes, remainingTaskCounts));
```
ssalinas: this duration (` System.currentTimeMillis() - reconciliationStart`) is not actually an indicator of how long task reconciliation has taken. Since the poller runs 30s later, the time reported for reconciliation will always be a multiple of 30s (or the configured wait time). The time we actually need to mark the finish is the most recent statusUpdate whose reason was `REASON_RECONCILIATON`

tpetr: Yep, realized that after I pushed 😞 . Working on some more updates.

ssalinas:
Reconciliation stats have been reporting fine in stable, merging this

PR #1137, opened 2016-07-12
SingularityService configuration & DC/OS support

martinstuder:
This pull request enables additional configuration of the singularityservice docker container to take place via environment variables. In addition, this pull request adds supporting Singularity images for a potential Singularity DC/OS package.


SingularityService/src/main/docker/start.sh
```diff
@@ -15,7 +15,12 @@ DEFAULT_URI_BASE="http://${SINGULARITY_HOSTNAME:=localhost}:${SINGULARITY_PORT:=
 
 args+=( -Xmx${SINGULARITY_MAX_HEAP:-512m} )
 args+=( -Djava.net.preferIPv4Stack=true )
+args+=( -Ddw.server.applicationContextPath="${SINGULARITY_APP_CTX:=/singularity}" )
```
tpetr: we've got a few other environment variable floating around that serve a similar purpose (see: https://github.com/HubSpot/Singularity/blob/master/SingularityUI/gulpfile.js#L17) -- @ssalinas do you think it'd be good to consolidate here?

tpetr:
LGTM aside from comment above, thanks!

SingularityService/src/main/docker/start.sh
ssalinas: So, we need two different ones. There is the app context path, and the uri base (host + path that the ui uses for config). We can reuse the `SINGULARITY_UI_BASE` var that we use to calculate the default uri base (or set them both to `SINGULARITY_APP_CTX` which might even make more sense)

ssalinas: Given the changes in #1236 , think the config and vars can stay as-is. Tested this out and going to merge though our branches 👍

PR #1141, opened 2016-07-14
Fix user requested destroy when cleaning

ssalinas:
@tpetr fixes the odd out of order signals we saw earlier, as well as making sure to not jump right to -9 for something that's cleaning but not actually sent a kill yet


ssalinas:
merging this into the task_kill branch since they are both in hs_stable now

PR #1142, opened 2016-07-14
Request Detail Fixes

wolfd:
- Added missing table (Scheduled & Pending)
 - Use `Section` instead of `h2` tags
 - Fix `Section` up


kwm4385:
Noticed that the bounce modal states the default expiration is 60 minutes, but there's no default value on the field

wolfd:
Yeah it's actually a backend default.

tpetr:
LGTM

PR #1146, opened 2016-07-14
Fix Running File Redirect

Calvinp:
Using react_router broke the running file redirect; this fixes it.

It also turns FileNotFound into a functional component.


kwm4385:
LGTM

PR #1148, opened 2016-07-15
Fix error handling, improve styling

Calvinp:
- Fix error handling so that errors can actually be caught.
- Change styling so that the error appears in a `pre` instead of a `code` - the path appears in the `code` instead.


kwm4385:
LGTM

PR #1149, opened 2016-07-15
Add the ability to catch certain types of errors

Calvinp:
Sorry @kwm4385 but I have one more error-related PR for you.

Some errors might be anticipated, so this adds the ability to catch those errors and not display them to the user.

### DIFF
--- a/SingularityUI/app/actions/api/base.es6
+++ b/SingularityUI/app/actions/api/base.es6
@@ -1,5 +1,6 @@
 import fetch from 'isomorphic-fetch';
 import Messenger from 'messenger';
+import Utils from '../../utils';
 
 const JSON_HEADERS = {'Content-Type': 'application/json', 'Accept': 'application/json'};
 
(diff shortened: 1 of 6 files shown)


kwm4385:
I'm already doing this here https://github.com/HubSpot/Singularity/blob/3ad0867db1cc6ff83c33e4ba36c5c64cc6931b9d/SingularityUI/app/actions/api/base.es6#L83

with `options.successResponseCodes` but looking back it makes no sense to still return success when that's not the case. Maybe we can integrate the two so that your code uses `options.successResponseCodes` instead, then remove that from the success condition.

Calvinp:
Yeah, that's what I was noticing. It actually looks like `options.successResponseCodes` is not even used anywhere, so we might be able to get rid of it. Alternatively, it could make sense to have both - in some cases certain response codes may be desired, wheras in others they're an expected error.

kwm4385:
It actually is being used, despite the naming being a bit misleading. See `FetchRequestRunHistory` for example

Calvinp:
Yeah, it probably makes more sense for the uses of those that I saw to catch those responses but not count it as success - since 404 isn't successful.

It might be more effort than it's worth to refactor all of those right now, but the request and deploy forms depend on getting the 400 back as an error. So maybe leave them both in for now.

kwm4385:
It won't break anything I don't think for the existing stuff to get an error back instead of success when using `successResponseCodes`. I'd be happy just doing it your way, and I can fix any side-affects that may come up.

kwm4385:
Yeah feel free to rename it.

Calvinp:
Done

kwm4385:
Away we go!

PR #1150, opened 2016-07-15
Singularity Request - Timezone field (API)

### LINKED ISSUE #1002: Should be able to specify TZ for SCHEDULED requests
CronExpression in the Quartz library has the ability to set the TimeZone that the cron expression should be resolved in (see: http://www.quartz-scheduler.org/api/2.1.7/org/quartz/CronExpression.html#setTimeZone(java.util.TimeZone)).  It would be nice if the SingularityRequest provided an optional string field called `scheduledTimeZone` that could be set to a valid TimeZone (i.e. passed to `java.util.TimeZone.getTimeZone`) and then have this field be used when building the CronExpression to test within `SingularityScheduler`.

ghost:
This PR extends the Singularity Request code to accept an optional (string) timezone through the API.  Local testing passed.

Still needs frontend changes to make the timezones accessible to the end user.

/cc @ssalinas @Calvinp 
/fixes #1002


SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
```diff
@@ -24,6 +25,7 @@
   private final Optional<String> schedule;
   private final Optional<String> quartzSchedule;
   private final Optional<ScheduleType> scheduleType;
+  private final Optional<TimeZone> scheduledTimeZone;
```
tpetr: i'd keep the naming consistent and go with `scheduleTimeZone`

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -695,6 +696,10 @@ private int numInstancesExpected(SingularityRequest request, SingularityPendingR
           } else {
             scheduleFrom = new Date(now);
             final CronExpression cronExpression = new CronExpression(request.getQuartzScheduleSafe());
+            Optional<TimeZone> scheduledTimeZone = request.getScheduledTimeZone();
+            if (scheduledTimeZone.isPresent()) {
+              cronExpression.setTimeZone(scheduledTimeZone.get());
+            }
```
tpetr: nitpick, but the variable here is probably unnecessary. also, if the variable is needed, try to make all your variables `final` unless they definitely need to be mutable (see above for an example)

tpetr:
Thanks for the PR! Could you add a unittest that exercises this feature? It would also be good to verify that setting a bogus timezone fails in an appropriate way (i.e. you get a meaningful error message back from the API)

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -9,6 +9,9 @@
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import javax.ws.rs.WebApplicationException;
+
+import org.apache.commons.jexl2.UnifiedJEXL.Exception;
```
ssalinas: don't think this import is used anywhere

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -3059,6 +3063,77 @@ public void testCronScheduleChanges() throws Exception {
...
 
   @Test
+  public void testValidQuartzTimeZone() {
+    SingularityRequest req = new SingularityRequestBuilder(requestId, RequestType.SCHEDULED)
+        .setQuartzSchedule(Optional.of("*/1 * * * * ? 2020"))
+        .setScheduleType(Optional.of(ScheduleType.QUARTZ))
+        .setScheduleTimeZone(Optional.of("GMT"))
+        .build();
+
+    requestResource.postRequest(req);
```
ssalinas: this test isn't validating anything separate and isn't asserting anything, I don't think we need it. Testing the post/validation of a valid timezone is already covered by the more detailed test you wrote further down

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
MattCCS: The WebApplicationException is used, but yes I don't believe the JEXL one is.

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -3058,6 +3061,66 @@ public void testCronScheduleChanges() throws Exception {
...
+    if (pendingTaskIds.get(0).getRequestId().equals(requestEST.getId())) {
+      nextRunEST = pendingTaskIds.get(0).getNextRunAt();
+      nextRunGMT = pendingTaskIds.get(1).getNextRunAt();
+    } else {
+      nextRunEST = pendingTaskIds.get(1).getNextRunAt();
+      nextRunGMT = pendingTaskIds.get(0).getNextRunAt();
+    }
+
+    // GMT happens first, so EST is a larger timestamp
+    Assert.assertEquals(nextRunEST - nextRunGMT, fiveHoursInMilliseconds);
```
tpetr: Consider using TimeUnit.HOURS.toMillis() instead

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -165,6 +166,12 @@ public SingularityRequest checkSingularityRequest(SingularityRequest request, Op
       checkBadRequest(!request.getScheduleType().isPresent(), "ScheduleType can only be set for scheduled requests");
     }
 
+    if (request.getScheduleTimeZone().isPresent()) {
+      if (!ArrayUtils.contains(TimeZone.getAvailableIDs(), request.getScheduleTimeZone().get())) {
+        badRequest("scheduleTimeZone %s was not a valid timezone code (e.g. 'US/Eastern' or 'GMT')", request.getScheduleTimeZone().get());
```
tpetr: Is there an ISO or RFC standard we can reference so that users know what they have to choose from?

ssalinas: As far as I can tell, the getAvailableIds() list is maintained by Oracle, not necessarily conforming to any particular standard (that I can find).

https://garygregory.wordpress.com/2013/06/18/what-are-the-java-timezone-ids/

tpetr: We can just mention that, then -- something like "... does not map to a valid Java TimeZone object." We should also make sure all the verb tenses match, i believe the rest are present tense

MattCCS: One other is not present tense, but I can make this change and align the tenses as well

ssalinas:
@Calvinp has this field been added to the UI in the rewrite? If not can we add it?

Calvinp:
It has been, yes.

ssalinas:
Going to merge this since UI field has been added in the rewrite

PR #1152, opened 2016-07-18
Trim whitespace around filename links

Calvinp:
Trim filenames in a task's file browser to avoid showing whitespace and including it in links.


kwm4385:
LGTM

PR #1156, opened 2016-07-19
Truncate long tags; Hover trigger to show the whole thing

Calvinp:
In a modal tagging input, truncate long tags.
On hover, an Popover is opened to show the whole thing.

cc @markhazlewood @tpetr @kwm4385 @wolfd

### DIFF
--- a/SingularityUI/app/components/common/modal/FormModal.jsx
+++ b/SingularityUI/app/components/common/modal/FormModal.jsx
@@ -1,10 +1,12 @@
 import React from 'react';
 import classNames from 'classnames';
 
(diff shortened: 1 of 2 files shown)


Calvinp:
Potential improvements:
- Make the character limit dynamically adjust based on size of the modal (I'm not exactly sure how to find that information).
- Use a click trigger rather than hover for mobile-friendliness (Currently using hover because it's entirely unclear just from looking that clicking would do anything).

I'd be fine merging this without those improvements but other opinions are welcome.

kwm4385:
LGTM

PR #1158, opened 2016-07-19
Fix the request detail page bounce messages

Calvinp:
Fixes:
- [x] The notification stating that the request is bouncing - now it catches all bouncing tasks, even incremental ones, and when there are too many running tasks it uses wording that clearly states that this is ok for now.
- [x] The notification stating that the request is being deployed - this has the same problem and that will be fixed as well.

Please sanity check the wording of the message.
CC @markhazlewood @wolfd @tpetr @kwm4385


SingularityUI/app/components/requestDetail/header/RequestAlerts.jsx
```diff
@@ -74,14 +73,14 @@ const RequestAlerts = ({requestId, requestAPI, bounces, activeTasksForRequest})
...
             <span>
-              Finished deploying {targetActiveInstances} of {instances} total instances, {nextDeployStepRemark}
+              Finished deploying {targetActiveInstances} total instances, {nextDeployStepRemark}
             </span>
           );
         } else {
           maybeDeployProgress = (
             <span>
               {
-                `Trying to deploy ${targetActiveInstances} of ${instances}
```
Calvinp: The problem here is that `of` didn't make sense. `targetActiveInstances` is not a fraction of `instances` in all cases. Instead, if the deploy is successful, it will override `instances`. If we want to continue mentioning `instances`, better wording may be `Trying to deploy ${targetActiveInstances} (instead of ${instances}) instances.`

tpetr: Gotcha, thanks

tpetr:
LGTM

PR #1160, opened 2016-07-20
Request -- Null Owner fix

MattCCS:
This PR prevents new requests from being created which have one or more null owners, which have been causing errors elsewhere in Singularity.

/cc @ssalinas @Calvinp @tpetr

PR #1161, opened 2016-07-20
Request -- Null Owner fix

MattCCS:
This PR prevents new requests from being created which have one or more null owners, which have been causing errors elsewhere in Singularity.

(Dupe of #1160 for master at ssalinas' request)

/cc @ssalinas @Calvinp @tpetr

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
@@ -115,7 +115,7 @@ public class SingularityValidator {
     checkBadRequest(request.getRequestType() != null, "RequestType cannot be null or missing");
 
     if (request.getOwners().isPresent()) {
-      checkBadRequest(!request.getOwners().get().contains(null), "Request owner(s) must not be null");
+      checkBadRequest(!request.getOwners().get().contains(null), "Request owners cannot contain null values");
     }
 
     if (!allowRequestsWithoutOwners) {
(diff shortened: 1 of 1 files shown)


ssalinas:
👍

Calvinp:
The problem was that there were `null`s in the list of owners? I had thought the problem was that the owners field itself was null. @wolfd

ssalinas:
The owners field itself is an optional. I can double check this to be sure, but as far as I know that should come through as an Optional.absent (meaning we are already handling it) if the owners field as a whole is null. If that is not the case though please update us here @Calvinp

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -114,6 +114,10 @@ public SingularityRequest checkSingularityRequest(SingularityRequest request, Op
     checkBadRequest(request.getId() != null && !StringUtils.containsAny(request.getId(), JOINER.join(REQUEST_ID_ILLEGAL_CHARACTERS)), "Id can not be null or contain any of the following characters: %s", REQUEST_ID_ILLEGAL_CHARACTERS);
     checkBadRequest(request.getRequestType() != null, "RequestType cannot be null or missing");
 
+    if (request.getOwners().isPresent()) {
+      checkBadRequest(!request.getOwners().get().contains(null), "Request owner(s) must not be null");
```
tpetr: wording here is a little vague (i.e. `owners: null` is actually valid, but the message makes it sound like it wouldn't be) -- perhaps something like "Request owners cannot contain null values" ?

MattCCS: Good point, I'll change that.

PR #1162, opened 2016-07-20
Reorganize task label colors

Calvinp:
Reorganizes task label colors such that:
- Task States that don't require action, but aren't serving requests are `info` or `primary` (different shades of blue that distinguish similar but different states)
- Task States that are currently serving requests are `success` (green)
- Task States that indicate a failure and may require attention are `danger` (red)
- Task States that indicate either a small problem or no problem, but aren't a task-initiated completion, are `warning` (orange)
- Task States that are unknown to the UI are `default` (black)

cc @ssalinas @tpetr @markhazlewood @wolfd @kwm4385 
This PR should not be merged until all of those people have a chance to look at it and share their opinions because it is a fairly large change in the look of tasks (even if the code is small).
The original motivation was to distinguish tasks that are currently serving requests from ones that aren't, but then grew to group similar states as a whole.

Screenshots:
screenshot 2016-07-20 10 42 06
screenshot 2016-07-20 11 08 45
screenshot 2016-07-20 14 20 20
screenshot 2016-07-20 14 20 58

### DIFF
--- a/SingularityUI/app/components/tasks/Columns.jsx
+++ b/SingularityUI/app/components/tasks/Columns.jsx
@@ -205,11 +205,11 @@ export const NextRun = (
       (rowData) => rowData.pendingTask.pendingTaskId.nextRunAt
     }
     cellRender={(cellData) => {
-      let label = <span className="label label-default">SCHEDULED</span>;
+      let label = <span className={`label label-${Utils.getLabelClassFromTaskState('TASK_SCHEDULED')}`}>SCHEDULED</span>;
       if (Utils.timestampWithinSeconds(cellData, config.pendingWithinSeconds)) {
-        label = <span className="label label-info">PENDING</span>;
+        label = <span className={`label label-${Utils.getLabelClassFromTaskState('TASK_PENDING')}`}>PENDING</span>;
       } else if (cellData < Date.now() - config.pendingWithinSeconds * 1000) {
-        label = <span className="label label-danger">OVERDUE</span>;
+        label = <span className={`label label-${Utils.getLabelClassFromTaskState('TASK_OVERDUE')}`}>OVERDUE</span>;
       }
       return (
         <div>
--- a/SingularityUI/app/utils.es6
+++ b/SingularityUI/app/utils.es6
@@ -66,24 +66,27 @@ const Utils = {
 
   getLabelClassFromTaskState(state) {
     switch (state) {
-      case 'TASK_STARTING':
-      case 'TASK_CLEANING':
-        return 'warning';
       case 'TASK_STAGING':
+      case 'TASK_STARTING':
+      case 'TASK_SCHEDULED':
+        return 'info'; // Not serving requests but no action necessary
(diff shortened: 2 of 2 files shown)


ssalinas:
A few things:
- `TASK_KILLED` should not be an orange. Tasks will end up in this state if they are asked to shut down and do so gracefully. The orange signifies to me that something is still wrong/might be wrong. I'd argue for keeping the current grey for killed
- I'm unsure about `TASK_CLEANING`, it'd make more sense to keep it green if we had another state to transition to when a task was in the process of being killed, and no longer serving requests (i.e. making better use of `TASK_KILLING`, but currently a cleaning task (for however short a time) may not always be serving requests (in the case of a web service). Could be convinced otherwise on this, but wanted to state the concern

kwm4385:
This might be easier to reason about if we make a tree of the task lifecycle chronologically and see if it makes sense to assign colors by region.

markhazlewood:
Approve of these changes and agree w/ @ssalinas and @kwm4385 on all counts.

Calvinp:
Here's the tree as best I understand it. Sorry it's not pretty, I'm not Mark :p There are probably a bunch more arrows going to `Lost`, `Failed`, and `Lost While Down`, but this should be the general idea.

screenshot 2016-07-20 14 56 45

The reason `TASK_CLEANING` was green is because I thought it was always serving requests. Remember the original problem was that `TASK_CLEANING` being orange was worrying developers, who thought there was nothing serving their requests for a short period. What color would you suggest?
Another possibility would be to redesign the request detail page entirely to show tasks that are serving requests in their own table. Then there could be at most one state of each color in each table (except the completed tasks table, since there are a bunch of bad states).

kwm4385:
taskstates

My reasoning:
Green: Task is running and guaranteed to be serving requests at this time
Light/dark blue: Task may not be serving requests, but the state that it's in is expected and no action is needed
Red: Something went wrong and you may want to investigate
Orange: Something is wrong but action might not be required on your part
Grey: Task is waiting/not relevent now

Calvinp:
While I agree with most of that chart, it still doesn't solve the anxiety that occurs when all tasks are in `cleaning` or `starting` during certain kinds of bounce.

kwm4385:
I suppose the only answer to that would be to make `cleaning` and `running` the same color.

Calvinp:
Then we're back to the problem that @ssalinas had which is that `cleaning` isn't guaranteed to be serving requests...
We're kind of stuck between a rock and a hard place here.

kwm4385:
Yeah I think its either we solve that anxiety by telling people something that isn't _quite_ true, or be absolutely accurate in our labeling.

Calvinp:
Maybe we could use dark blue for `cleaning` and light blue for `finished`. Then add a note of some sort (maybe an alert) that cleaning tasks may still be accepting requests.

Calvinp:
Closing this to avoid making huge UX changes in the `decaf` branch.
Once `deacf` is merged we will circle back to this.

PR #1163, opened 2016-07-20
RETRY for scheduled task should keep cmdLineArgs

### LINKED ISSUE #1133: Task retry does not take extra command-line args into account?
We've just seen the following pattern for a request:

1. Task with extra command line args fails
1. New task is created with message `Task launched because of RETRY`
1. New task does not have any extra command line args

Is that expected or a bug?

/cc @ssalinas

ssalinas:
/cc @demobox 
/fixes #1133


demobox:
Thanks for jumping on this so quickly, @ssalinas!

PR #1164, opened 2016-07-21
Request and Deploy Title Copy Link

Calvinp:
Adds a link next to the request ID to copy it to the clipboard.

TODO:
- [x] Make the link disappear on mouse off
- [x] Properly handle short request IDs


SingularityUI/app/components/requestDetail/header/RequestTitle.jsx
```diff
@@ -16,40 +17,67 @@ const errorDescription = (requestAPI) => {
...
+          </span>
         </span>
-      </span>
+      );
+    }
+
+    const requestIdToDisplay = Utils.maybe(requestAPI, ['data', 'request', 'id']) || requestId;
+
+    return (
+      <div onMouseOver={() => this.onMouseOver()}>
```
zdhickman: If `this.onMouseOver` is bound, you can just do `onMouseOver={this.onMouseOver}`

Calvinp:
Trying to have a link pop up on hover that could be clicked to copy things wasn't working. Instead, now the requestID itself can be clicked to copy it.

An overlay trigger explains this fact to the user when they hover over the request id.

Calvinp:
@tpetr @kwm4385 @wolfd this PR is good to be reviewed/merged.

SingularityUI/app/components/requestDetail/header/RequestTitle.jsx
```diff
@@ -39,13 +40,22 @@ const RequestTitle = ({requestId, requestAPI}) => {
     );
   }
 
+  const requestIdToDisplay = Utils.maybe(requestAPI, ['data', 'request', 'id']) || requestId;
```
tpetr: why bother with the `maybe()` if we have the `requestId` variable?

Calvinp: Honestly, I don't know. I copied it from what was there before (see line 48 on the other side of the diff), but I can't justify it.
I'll delete it - `requestId` is marked as required in propTypes after all.

Calvinp:
I also added the same ability to the Deploy Detail page. It doesn't make sense to add it to the Task Detail page because the task ID isn't displayed as the title.

tpetr:
LGTM

PR #1165, opened 2016-07-21
Improvements to Task Search

Calvinp:
Makes the following improvements to Task Search:
- Changes the URL from `/taskSearch` to `/task-search` to be more in line with URLs in the rest of the UI.
- Changes the header from `(Global) Historical Tasks` to `Task Search`.
- Adds labels for the current search parameters so that it's possible to see current search terms (even if the fields have changed).
- Removes the JSON button.
- Changes the log button to be a `file` glyphicon rather than `...`. This is also changed in the Task History table in the request detail page.
- Fixes a bug that caused the table to not update when a new search is performed that returns 0 results.
- Preemptively fixes the bug that occurred in the deployDetail page - contributed to by making api calls in `componentWillMount()`.

### DIFF
--- a/SingularityUI/app/components/taskSearch/TaskSearch.jsx
+++ b/SingularityUI/app/components/taskSearch/TaskSearch.jsx
@@ -160,13 +160,7 @@ function mapDispatchToProps(dispatch) {
   };
 }
 
-let firstLoad = true;
-
 function refresh(props) {
-  if (!firstLoad) {
-    return null;
-  }
-  firstLoad = false;
   const promises = [];
   const filter = _.extend({}, { requestId: props.params.requestId }, props.filter);
   promises.push(props.fetchTaskHistory(INITIAL_TASKS_PER_PAGE, 1, filter));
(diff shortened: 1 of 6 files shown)


Calvinp:
Did it again... Forgot to cc @tpetr @kwm4385 @wolfd

SingularityUI/app/components/taskSearch/TaskSearch.jsx
```diff
@@ -141,4 +160,18 @@ function mapDispatchToProps(dispatch) {
   };
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(rootComponent(TaskSearch, 'Task Search'));
+let firstLoad = true;
+
+function refresh(props) {
+  if (!firstLoad) {
```
kwm4385: If you only want refresh called when the page is first loaded and not again in an interval, just use the optional `refreshInterval` bool in `rootComponent` to false. `RequestDetail` does this https://github.com/HubSpot/Singularity/blob/decaf/SingularityUI/app/components/requestDetail/RequestDetailPage.jsx#L103

kwm4385:
LGTM

wolfd:
LGTM

PR #1169, opened 2016-07-25
support multiple docker parameters and task labels

### LINKED ISSUE #1074: SingularityDockerInfo.parameters only supports a single value per key
`docker run` has many options that take an array for the same key.  You can see the list with `docker run --help | grep "\[\]"`.

The `parameters` field is a Map[string, string] so specifying multiple values for the same option only uses the last value for the key.  For example, to get this run command `docker run --dns 8.8.8.8 --dns 8.8.4.4 ...` you would need to add multiple dns entries with the same key like so:
```
"parameters": {
  "dns": "8.8.8.8",
  "dns": "8.8.4.4"
}
```

Since `docker run` has so many options that take an array, it would be good to have some general support for specifying them in `SingularityDockerInfo.parameters` and passing them through.

We're currently running Singularity 0.5.0, Mesos 0.26.1, and using the Mesos Docker Containerizer.

tpetr:
Updates Docker parameters and task labels to support multiple values for a specific key. Still testing this out.

Fixes #1074


ssalinas:
@tpetr it seems the `@JsonDeserialize` parameter wasn't playing nice with the `List<T>`. It was trying to deserialize each individual SingularityDockerParameter with the custom class, not the entire list. Serialization was a bit odd also.

In cef1df4 I implemented some 'wrapper' classes for handling the custom (de)serialization. Tested it and the old Map<String, String> as well as the list of parameters was working locally. Let me know what you think of this method of getting it working.

ssalinas:
Note here, there was not an easy way to do this that wasn't a breaking change (that I could find). So we will need to make it clear in the release notes that while the api will accept `Map<String, String>` or `List<SingularityDockerParameter>`, it will always be returning a `List<SigularityDockerParameter>`, similar for task labels.

SingularityBase/src/main/java/com/hubspot/mesos/SingularityDockerInfo.java
```diff
@@ -16,26 +17,41 @@
...
                                @JsonProperty("privileged") boolean privileged,
                                @JsonProperty("network") SingularityDockerNetworkType network,
                                @JsonProperty("portMappings") Optional<List<SingularityDockerPortMapping>> portMappings,
                                @JsonProperty("forcePullImage") Optional<Boolean> forcePullImage,
-                               @JsonProperty("parameters") Optional<Map<String, String>> parameters) {
+                               @JsonProperty("parameters") Optional<Map<String, String>> parameters,
+                               @JsonProperty("dockerParameters") Optional<List<SingularityDockerParameter>> dockerParameters) {
+    if (dockerParameters.isPresent() && !dockerParameters.get().isEmpty() && parameters.isPresent() && !parameters.get().isEmpty()) {
+      throw new IllegalArgumentException("Can only specify one of 'parameters' or 'dockerParameters'");
+    }
```
tpetr: I'm a little hesitant of putting these kinds of exceptions in constructors -- if we somehow had a bogus one saved we'd be throwing every time it was deserialized. What do you think about moving this into `SingularityValidator` instead?

ssalinas:
@tpetr moved those checks to the validator. Since I'm setting both objects when the deprecated one is set, validating by making sure that they are the same

PR #1170, opened 2016-07-25
Add TASK_ID environment variable with the singularity task_id

grepsr:
This could be helpful to get the actual TASK_ID of a Singularity task, rather than just the TASK_REQUEST_ID


grepsr:
Our containers rely exclusively on the TASK_ID variable to do some cleanup when webhooks are triggered. I think it is a good addition, because I did not find a very straight forward way to get the exact TASK_ID for a  request (there might be multiple tasks in a request)

grepsr:
Can we please get this small patch into 0.10?

Having the actual TASK_ID accessible on every container is certainly helpful for referencing purposes.

Thanks

PR #1171, opened 2016-07-25
Do not remove from LB if not yet added

ssalinas:
In this order of events, we were submitting a LB `ADD` request _after_ the `REMOVE` request, with the result that the presence of the REMOVE caused us to never properly remove the task from the LB:

- deploy `1` running
- deploy `2` starts
- replacement task for deploy `1` enters `TASK_RUNNING`
- deploy `2` tasks are healthy and LB `ADD`/`REMOVE` is triggered (but contains the replacement task above)
- replacement task from deploy `1` becomes healthy and LB `ADD` is sent

This fixes the deploy logic to only issue the `REMOVE` as part of the deploy if the `ADD` is already present and successful. If the `ADD` completes after/during the deploy lb actions, the task will still be properly removed from the LB during cleanup since the LB `REMOVE` will not be present.


ssalinas:
This tested well in each of our environments, merging the fix

PR #1172, opened 2016-07-25
Deploy detail page fixes and improvements

Calvinp:
- Fixes the bug that caused the deployDetailPage to get stuck in an infinite render/API call loop.
- Fixes the bug that prevented pending deploys from being rendered.
- Implements new deploy form redirecting to the created deploy if creating that deploy was successful.
- Implements 404 scaffolding - an api call can have an options parameter `isMainApiCall`, which if set automatically ignores 404s (so the user doesn't get the popup dialog). The page can then see the 404 and render a not found page instead of trying to render itself (annoying boilerplate that it would be nice to get rid of). Note - once this is merged, other pages can make use of this functionality to render the not found page.
- Now deploys that don't exist will show up as a not found page to the user.
- General code cleanliness improvements.

cc @tpetr @kwm4385 @wolfd

### DIFF
--- a/SingularityUI/app/rootComponent.jsx
+++ b/SingularityUI/app/rootComponent.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
 
-// eslint-disable-next-line no-unused-vars react/no-multi-comp
 const rootComponent = (Wrapped, title, refresh = _.noop, refreshInterval = true, pageMargin = true) => class extends React.Component {
 
   constructor(props) {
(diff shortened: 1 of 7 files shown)


SingularityUI/app/rootComponent.jsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
 
-// eslint-disable-next-line no-unused-vars react/no-multi-comp
```
Calvinp: Forgot about this line. That `eslint-disable` was not necessary so I removed it when I thought the bugfix was in `rootComponent`.
Its deletion in this PR isn't harmful so I'm going to leave it deleted.

kwm4385:
LGTM

PR #1176, opened 2016-07-28
remove the now > start check in RFC5545

ssalinas:
Need to loop from the real start each time to get the scheduling correct.

@grepsr can test and confirm that this works the way you would expect?

/cc @tpetr


grepsr:
@ssalinas Thanks I will test this!

Regarding MAX_ITERATIONS = 10000, I think 10k is a small number.

If someone wants to do an hourly job, it will stop working after 1.1 years

Since the loop will return midway, once a matching interval is found - it will hardly ever loop a lot?
Perhaps we should just increase the number to MAX_ITERATIONS = 1000000?
This way it will work properly for 114 years lol.

Hate to have so many loops, but if you think of it, the cron library probably loops internally like this too - so it seems ok?

ssalinas:
@grepsr bumped the max iterations as well.

grepsr:
Awesome, thanks!

PR #1177, opened 2016-07-28
Task detail bug fixes

Calvinp:
Fixes the following bugs in the task detail page:
- Stray `0`s popped up in the `Resource Usage` section from time to time.
- `UsageInfo` component had a failed propType
- Healthcheck failure message said `Last healthcheck responded with [object Object]` instead of the proper HTTP status code.
- On 404 of task files, a notification was shown to the user that made it seem like API requssts are failing. Wheras in reality it was just that the slave was decommissioned.
- There were some variable names such as `hcs` or `h`. These are now expanded to `healthcheckResults` and `healthcheck`
- Extra Command Line Args could get passed into InfoBox as an array. This is now properly handled.

Note that the first three of these are bugs that I myslef introduced while fixing bigger bugs or code uncleanliness.


Calvinp:
cc @wolfd @kwm4385 @tpetr

tpetr:
LGTM

PR #1178, opened 2016-07-28
Replace dashes with underscores when searching for matching hosts

Calvinp:
Sometimes hostnames are presented with dashes, sometimes with underscores.  On the tasks page they are presented with underscores. Because of this, when linked from a page where they're presented with dashes, no match would be found. This fixes that.

### DIFF
--- a/SingularityUI/app/selectors/tasks.es6
+++ b/SingularityUI/app/selectors/tasks.es6
@@ -68,7 +68,7 @@ export const getFilteredTasks = createSelector(
         _.each(tasks, (t) => {
           t.id = id.extract(t);
         });
-        const hostMatch = fuzzy.filter(filter.filterText, tasks, host);
+        const hostMatch = fuzzy.filter(filter.filterText.split('-').join('_'), tasks, host);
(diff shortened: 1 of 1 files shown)


Calvinp:
cc @wolfd @kwm4385 @tpetr

SingularityUI/app/selectors/tasks.es6
```diff
@@ -68,7 +68,7 @@ export const getFilteredTasks = createSelector(
         _.each(tasks, (t) => {
           t.id = id.extract(t);
         });
-        const hostMatch = fuzzy.filter(filter.filterText, tasks, host);
+        const hostMatch = fuzzy.filter(filter.filterText.split('-').join('_'), tasks, host);
```
tpetr: why `split()` and `join()` vs. `replace()`

Calvinp: Because I didn't realize I could use a `/g` flag to make the replacement global.

tpetr:
LGTM

PR #1179, opened 2016-07-29
Handle 404 properly

Calvinp:
Handles 404s on pages for specific requests, deploys, and tasks by showing the 404 page rather than rendering something meaningless and showing an error.

The following pages are affected:

- [x] DeployDetail
- [x] RequestDetail
- [x] TaskDetail
- [x] New Deploy Form
- [x] Edit Request Form

Please let me know if I've missed any @tpetr @kwm4385 @wolfd


Calvinp:
Looks like it's not going to be possible to detect 404 on Task Search right now.
The server responds with an empty array rather than a 404, even when using the `/request/:requestId/task-search` URL.
@ssalinas a possible future improvement to the Task Search API would be to 404 when the request isn't found in this case.

@tpetr @kwm4385 @wolfd This PR is good to be reviewed/merged from my perspective now.

SingularityUI/app/actions/api/history.es6
```diff
@@ -3,8 +3,9 @@ import Utils from '../../utils';
 
 export const FetchTaskHistory = buildApiAction(
   'FETCH_TASK_HISTORY',
-  (taskId) => ({
-    url: `/history/task/${taskId}`
+  (taskId, isMainApiCall) => ({
+    url: `/history/task/${taskId}`,
+    isMainApiCall
```
wolfd: What does this mean?

SingularityUI/app/components/deployDetail/DeployDetail.jsx
```diff
@@ -346,6 +342,7 @@ function mapStateToProps(state) {
 
   return {
     notFound: state.api.deploy.statusCode === 404,
+    pathname: ownProps.location.pathname,
```
wolfd: hm, not sure how I feel about this. Even though the web api doesn't capitalize pathname, it also doesn't capitalize `onClick`, etc. Maybe we should use `pathName`? @tpetr

SingularityUI/app/actions/api/history.es6
Calvinp: Unfortunately the meaning snuck into another PR - that's on me, I implemented it with the Deploy Detail page fixes.
It's basically an idiomatic way of saying it ignores 404s. It signifies that if the API call fails with a 404 then the page should 404 (but it can also be extended to mean other things later).

SingularityUI/app/components/deployDetail/DeployDetail.jsx
Calvinp: I was following the wording/capitalization of `react-router` here.

SingularityUI/app/actions/api/history.es6
tpetr: `isMainApiCall` doesn't convey that information at all -- if that's all that this variable controls, can we think of a more descriptive name for it?

Calvinp: Naming things isn't my strong suit. Maybe `pageCantRenderIfThisFails`? I don't really like that name either though.

Calvinp: Or `page404sIfThisCall404s`.

tpetr: It's cool, naming is hard 😄  -- I'd suggest something like `renderNotFoundIf404`

wolfd:
LGTM

tpetr:
LGTM

PR #1182, opened 2016-07-29
Add timezone field and kill dropdowns if there are less than 5 options

Calvinp:
Adds the timezone field to the New Request Form, and turns dropdowns into buttons if there are less than 5 options.
This looks like this:
screenshot 2016-07-29 14 39 09

cc @kwm4385 @tpetr @wolfd


kwm4385:
Is there an enumeration somewhere of valid timezones we can autocomplete/have the user choose from instead of guessing?

SingularityUI/app/components/common/formItems/formGroups/SelectFormGroup.jsx
```diff
@@ -1,20 +1,46 @@
...
-      className={props.id}
-      options={props.options}
-      onChange={props.onChange}
-      value={props.value}
-      clearable={false}
-    />
-  </FormGroup>
-);
+const SelectFormGroup = (props) => {
+  let selectors;
```
zdhickman: A more accurate name would relate to the options you're rendering

Calvinp:
Yes, there are lots of enumerations out there. I'm not sure if any of them correspond exactly to the one in Java.utils.TimeZone though.

kwm4385:
Looks like it's listable though `TimeZone.getAvailableIDs();`
While it may be overkill to write an endpoint to get that, perhaps we could run it once and just capture the output.

Calvinp:
This is now implemented.

tpetr:
LGTM

PR #1183, opened 2016-08-01
Fix hopefully the last bug in the deploy detail page on decaf

Calvinp:
Fixes a bug in the deploy detail page that caused the TaskHistory table to show tasks from the last deploy viewed.

Implementation justification:
The problem was that `firstLoad` wasn't an instance variable, and so was only set when the app was first loaded.
One simple solution would be to put this functionality into `componentWillMount()`, but that was what caused the infinite API requests bug.
Another simple solution would be to set `refreshInterval` to be false, but the component does need a refresh interval for other pieces.
So instead I added an `initialize` parameter to the `rootComponent`. If it's set, it will be called instead of `refresh` on the first load. In this case, `initialize` also calls `refresh` to avoid code duplication.

cc @tpetr @kwm4385 @wolfd

### DIFF
--- a/SingularityUI/app/components/deployDetail/DeployDetail.jsx
+++ b/SingularityUI/app/components/deployDetail/DeployDetail.jsx
@@ -353,17 +353,9 @@ function mapStateToProps(state) {
   };
 }
 
-let firstLoad = true;
-
-function refresh(props) {
-  const promises = [];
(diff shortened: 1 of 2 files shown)


wolfd:
LGTM

tpetr:
LGTM

PR #1184, opened 2016-08-02
fail differently if request data not present during deploy check

ssalinas:
@tpetr


tpetr:
LGTM, could we add a unit test for this situation?

ssalinas:
FYI, in testing I found this was only triggered when the deployChecker ran before the cleaner after a request delete during a deploy. If the cleaner has already run, we do not hit the optional.get() error because the pending deploy is removed by the cleaner.

ssalinas:
Forgot to comment earlier that unit test was added, going to merge

PR #1185, opened 2016-08-02
Add JSON Button to the Racks and Slaves pages

Calvinp:
Exactly what it says on the tin - adds a JSON button to the Racks and Slaves pages.


tpetr:
LGTM

PR #1186, opened 2016-08-02
Add ability to show task disk resource if configured to do so

Calvinp:
Exactly what it says on the tin.

If `config.showTaskDiskResource` is true, the disk resource will be shown on the tasks table and as an option on the New Deploy form.

cc @tpetr @kwm4385 @wolfd


tpetr:
LGTM

Issue #1187, opened 2016-08-02
Taks that exits in less than a millisecond can cause incorrect UI display


stevenschlansker:
We have a task that ran for less than 1ms:
https://gist.github.com/stevenschlansker/3a8ecd30098fdd9f901736cad66a8d4d

In the task state it shows as finished, which is correct:
![screen shot 2016-08-01 at 5 31 41 pm](https://cloud.githubusercontent.com/assets/129097/17336766/7fe07a3c-5894-11e6-978b-eec74ae65ffd.png)

but in the history it shows as still running!
![screen shot 2016-08-01 at 5 31 38 pm](https://cloud.githubusercontent.com/assets/129097/17336775/878e4494-5894-11e6-80be-77cf325ea0c5.png)

here's the singularity logs grepping for task id:
```
singularity.log.1:INFO  [2016-08-01 23:15:18,696] com.hubspot.singularity.mesos.SingularityMesosScheduler: Launching task cpennello_pop_incrementality_reso_time_hist-cpennello.2016.08.01T23.15.17-1470093318696-1-mesos_slave4_prod_uswest2.otsql.opentable.com-us_west_2c slot on slave 0b9fd3a2-7e1a-4ee7-8d95-59425710b8c7-S34 (mesos-slave4-prod-uswest2.otsql.opentable.com)
singularity.log.1:INFO  [2016-08-01 23:15:18,713] com.hubspot.singularity.mesos.SingularityMesosScheduler: 1 tasks ([cpennello_pop_incrementality_reso_time_hist-cpennello.2016.08.01T23.15.17-1470093318696-1-mesos_slave4_prod_uswest2.otsql.opentable.com-us_west_2c]) launched with status DRIVER_RUNNING
singularity.log.1:DEBUG [2016-08-01 23:15:29,425] com.hubspot.singularity.mesos.SingularityMesosScheduler: Task cpennello_pop_incrementality_reso_time_hist-cpennello.2016.08.01T23.15.17-1470093318696-1-mesos_slave4_prod_uswest2.otsql.opentable.com-us_west_2c is now TASK_RUNNING () at 1470093329403 
[...]

PR #1188, opened 2016-08-02
sort task history updates by ExtendedTaskState enum ordinal, not timestamp

ssalinas:
I found a few cases where it could happen that what we would consider the 'current state' of the task, is not the update that has the latest timestamp:
- Rapid updates in succession, causing the history updates to have identical or out-of-order timestamps
- cleaning triggered early in task lifecycle, followed quickly by a starting/running update will indefinitely leave the task in that state (not allowing it to clean/shut down properly)

This PR updates the sort called when fetching task histories to sort by the `ExtendedTaskSTate` ordinal instead of relying only on timestamp. This should give us the correct 'current' state in all cases since the enum is defined in the order we expect statuses to progress during the task lifecycle.

/cc @tpetr @stevenschlansker @subratbasnet #1187

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdHistory.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdHistory.java
@@ -17,19 +17,12 @@ public class SingularityTaskIdHistory implements Comparable<SingularityTaskIdHis
   private final Optional<ExtendedTaskState> lastTaskState;
   private final Optional<String> runId;
 
-  private static final Comparator<SingularityTaskHistoryUpdate> comparator = new Comparator<SingularityTaskHistoryUpdate>() {
-    @Override
-    public int compare(SingularityTaskHistoryUpdate o1, SingularityTaskHistoryUpdate o2) {
-      return o1.getTaskState().compareTo(o2.getTaskState());
-    }
-  };
-
   public static SingularityTaskIdHistory fromTaskIdAndTaskAndUpdates(SingularityTaskId taskId, SingularityTask task, List<SingularityTaskHistoryUpdate> updates) {
     ExtendedTaskState lastTaskState = null;
     long updatedAt = taskId.getStartedAt();
 
     if (updates != null && !updates.isEmpty()) {
-      SingularityTaskHistoryUpdate lastUpdate = Collections.max(updates, comparator);
+      SingularityTaskHistoryUpdate lastUpdate = Collections.max(updates);
       lastTaskState = lastUpdate.getTaskState();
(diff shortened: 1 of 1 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdHistory.java
```diff
@@ -15,17 +18,20 @@
...
 
     if (updates != null && !updates.isEmpty()) {
-      SingularityTaskHistoryUpdate lastUpdate = updates.get(0);
-      for (SingularityTaskHistoryUpdate update : updates.subList(1, updates.size())) {
-        if (update.getTimestamp() > lastUpdate.getTimestamp()) {
-          lastUpdate = update;
-        }
-      }
+      Collections.sort(updates, comparator);
+      SingularityTaskHistoryUpdate lastUpdate = Iterables.getLast(updates);
```
stevenschlansker: Could consider using `Collections.max` here for a slight efficiency gain, also avoids mutating an argument (which is usually dicey code style IMO)

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdHistory.java
```diff
@@ -15,17 +18,20 @@
   private final Optional<ExtendedTaskState> lastTaskState;
   private final Optional<String> runId;
 
+  private static final Comparator<SingularityTaskHistoryUpdate> comparator = new Comparator<SingularityTaskHistoryUpdate>() {
```
stevenschlansker: Should this be composed with a comparator over timestamp, in case a task enters the same state more than once, as a tiebreaker?  or is that not possible?

ssalinas: We can't have two separate state updates for the same state, so shouldn't need that

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdHistory.java
ssalinas: 👍 on `Collections.max`, updated

tpetr:
(FYI @wsorenson)

ssalinas:
Another fyi, the original change was introduced by me back in #932 when trying to improve task search speed. The list of updates being returned by a new way of fetching them was an unmodifiable list, so I had resorted to checking timestamps. Should have been checking ordinals instead there

grepsr:
I've tested this extensively yesterday with over a 100 tasks. The earlier problem did not recur. So looks like this working!

Thanks!

ssalinas:

PR #1189, opened 2016-08-03
Better cleanup of incremental tasks

ssalinas:
This PR addresses two bugs:
- When an incremental deploys fails, it will wait for the active deploy to return to normal instance count so there is never reduced capacity in a load balancer. If there are not enough resources to return the active deploy to the original instance count, these can essentially be 'stuck' in  cleaning state waiting for more tasks to start.
- If bounce is initiated during a deploy that will fail and the replacement tasks for the bounce are continually failing healthchecks, the failed deploy tasks will not shut down and will remain in cleaning until the bounce completes (not harmful, but annoying since they should be killed more quickly)

These are both due to the change introduced in d6ba7e3ea431f387b0d06de7795e6d5ab0d60d1b which considered deploy failures and cancellations as non-immediate kills for the sake of incremental deploy rollbacks as mentioned in the first point above.

This PR updates the deploy checker and cleaner to be more aware of incremental vs non-incremental deploy failures so that the cleanup can be done appropriately, as well as cleaning up tasks from failed incremental deploys in a more rolling fashion (similar to incremental bounce)

/cc @tpetr


ssalinas:

PR #1190, opened 2016-08-04
Run now fixes

Calvinp:
Implements the following changes:
- A rerun button now appears on the task history table for requests whose tasks can be rerun.
- The run now dialog imports default command line arguments from the previously run task - unless it's a rerun in which case it imports from the task to be rerun.
- The run now dialog guesses that you want to tail the running log path.
- Code cleanliness, including not mutating state.

### DIFF
--- a/SingularityUI/app/components/common/modal/FormModal.jsx
+++ b/SingularityUI/app/components/common/modal/FormModal.jsx
@@ -8,17 +8,28 @@ import Select from 'react-select';
 
 const TAGS_CHARACTER_LIMIT = 75;
 
+function getDefaultFormState(props) {
+  const formState = {};
+  props.formElements.forEach((formElement) => {
+    const { defaultValue } = formElement;
+    if (defaultValue) {
+      if (Array.isArray(defaultValue)) {
+        formState[formElement.name] = defaultValue;
+      } else {
+        formState[formElement.name] = formElement.defaultValue.toString();
+      }
+    }
+  });
+  return formState;
+}
+
 export default class FormModal extends React.Component {
   constructor(props) {
     super(props);
-    const formState = {};
-    props.formElements.forEach((formElement) => {
-      formState[formElement.name] = formElement.defaultValue && formElement.defaultValue.toString();
-    });
 
     this.state = {
       visible: false,
-      formState,
+      formState: getDefaultFormState(props),
       errors: {}
     };
 
@@ -54,13 +65,14 @@ export default class FormModal extends React.Component {
 
   show() {
     this.setState({
-      visible: true
+      visible: true,
+      formState: getDefaultFormState(this.props)
     });
   }
 
   handleFormChange(name, value) {
     const formState = this.state.formState;
-    formState[name] = value;
+    formState[name] = value; // Mutates pre-existing state, probably bad
     this.setState({ formState });
   }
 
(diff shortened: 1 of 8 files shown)


SingularityUI/app/components/common/modal/FormModal.jsx
```diff
@@ -59,7 +77,7 @@ export default class FormModal extends React.Component {
   }
 
   handleFormChange(name, value) {
-    const formState = this.state.formState;
+    const formState = Utils.deepClone(this.state.formState);
     formState[name] = value;
```
Calvinp: This was mutating the existing state, which is bad and was causing problems.

Calvinp:
Forgot to cc here too...  @tpetr @kwm4385 @wolfd

kwm4385:
LGTM

SingularityUI/app/components/common/modal/FormModal.jsx
```diff
@@ -5,26 +5,44 @@ import { Modal, Button, Popover, OverlayTrigger } from 'react-bootstrap';
...
     };
 
     _.bindAll(this, 'hide', 'show', 'confirm');
   }
 
+  componentWillReceiveProps(newProps) {
+    if (_.isEqual(this.state.formState, getDefaultFormState(this.props))) {
+      this.setState({formState: getDefaultFormState(newProps)});
+    }
+  }
```
tpetr: can you explain what this method does?

SingularityUI/app/components/common/modal/ModalWrapper.es6
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 
-export const getClickComponent = (component) => (
+export const getClickComponent = (component, doFirst = (() => Promise.resolve())) => (
```
tpetr: i think you can replace `() => Promise.resolve()` with a method reference (`Promise.resolve`)

tpetr:
LGTM aside from my comments

SingularityUI/app/components/common/modal/FormModal.jsx
Calvinp: If the new props set a new default form state, and the form hasn't been changed off of the default, this will update the current value of the form to the new default value.

This is needed because the default value of the command line arguments isn't always present when the modal is initially rendered on page load. When the run now button is pressed, the last task run (or the task to be rerun) is loaded, then the default command line args are taken from that task. The value of the field then needs to be updated with these command line arguments.

SingularityUI/app/components/common/modal/ModalWrapper.es6
Calvinp: I initially thought so too, but when I try to do that I get `Uncaught TypeError: function resolve() { [native code] } called on non-object`.

SingularityUI/app/components/common/modal/FormModal.jsx
wolfd: I'm curious why the command line args aren't _always_ present. How is that data coming in?

Calvinp: The command line args are taken from:
- The task being rerun, if you are rerunning a task.
- Else the last task that was completed on that request.
Right now, to find the command line arguments, it's necessary to load the task's information. That's fine when on the request detail page, as the task is already loaded (the command line args are part of the data retrieved while loading the task history table). But on the requests page we don't load the task detail. At some point we need to load the detail of the last task run, but it doesn't make sense to do that before rendering (there could be thousands of requests). So instead that call is only sent when a user clicks the run now button.

PR #1193, opened 2016-08-05
better indiciation of request deletion progress

ssalinas:
When checking if a request has been deleted (i.e. to recreate it as another type) an api can often hit the api too soon causing a 409 because the request is still in cleaning. This saves a `DELETE_REQUESTED` history event, followed by a `DELTED` when the cleaner has finished so that a polling client has some indication of when it can re-post it's new request data.

Another option would be to build a way to change types into Singularity (have it handle the delete followed by creation/etc), but that seemed rather involved for something that is likely rarely done.

/cc @gchomatas @tpetr


ssalinas:
Closing this and will look into other solutions for this

PR #1194, opened 2016-08-05
deploy failure message for non-task-specific failures

ssalinas:
If we hit a case where there is a non-task-specific failure (such as the not enough resources to launch all tasks one), we should use the correct failure message in the deploy result and omit the list of deploy failures. In other words the calculated one takes precedence

/cc @gchomatas


ssalinas:

PR #1197, opened 2016-08-09
Guarantee durationMillis is present in getExpiringBounce response

MattCCS:
This PR guarantees that a bounce request's result always includes the durationMillis, whether a value was given or not.  If no value was given, the durationMillis value is the appropriate default.

/cc @ssalinas @wolfd @tpetr

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityBounceRequestBuilder.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityBounceRequestBuilder.java
@@ -74,9 +74,9 @@ public class SingularityBounceRequestBuilder {
     return "SingularityBounceRequestBuilder{" +
         "durationMillis='" + durationMillis + '\'' +
         ", actionId='" + actionId + '\'' +
-        ", message=" + message + '\'' +
-        ", incremental=" + incremental + '\'' +
-        ", skipHealthchecks=" + skipHealthchecks + '\'' +
+        ", message='" + message + '\'' +
+        ", incremental='" + incremental + '\'' +
+        ", skipHealthchecks='" + skipHealthchecks + '\'' +
         "}";
   }
 }
--- a/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityBounceRequestBuilder.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityBounceRequestBuilder.java
@@ -1,0 +1,82 @@
+package com.hubspot.singularity.api;
+
+import com.google.common.base.Optional;
+
+public class SingularityBounceRequestBuilder {
+
+  private Optional<Boolean> incremental;
+  private Optional<Boolean> skipHealthchecks;
+
+  private Optional<Long> durationMillis;
+  private Optional<String> actionId;
+  private Optional<String> message;
+
+  public SingularityBounceRequestBuilder() {
+    this.incremental = Optional.absent();
+    this.skipHealthchecks = Optional.absent();
+
+    this.durationMillis = Optional.absent();
+    this.actionId = Optional.absent();
+    this.message = Optional.absent();
+  }
+
+  public SingularityBounceRequest build() {
+    return new SingularityBounceRequest(incremental, skipHealthchecks, durationMillis, actionId, message);
+  }
+
+  public Optional<Long> getDurationMillis() {
+    return durationMillis;
+  }
+
+  public SingularityBounceRequestBuilder setDurationMillis(Optional<Long> durationMillis) {
+    this.durationMillis = durationMillis;
+    return this;
+  }
+
+  public Optional<String> getActionId() {
+    return actionId;
+  }
+
+  public SingularityBounceRequestBuilder setActionId(Optional<String> actionId) {
+    this.actionId = actionId;
+    return this;
+  }
+
+  public Optional<String> getMessage() {
+    return message;
+  }
+
+  public SingularityBounceRequestBuilder setMessage(Optional<String> message) {
+    this.message = message;
+    return this;
+  }
+
+  public Optional<Boolean> getIncremental() {
+    return incremental;
+  }
+
+  public SingularityBounceRequestBuilder setIncremental(Optional<Boolean> incremental) {
+    this.incremental = incremental;
+    return this;
+  }
+
+  public Optional<Boolean> getSkipHealthchecks() {
+    return skipHealthchecks;
+  }
+
+  public SingularityBounceRequestBuilder setSkipHealthchecks(Optional<Boolean> skipHealthchecks) {
+    this.skipHealthchecks = skipHealthchecks;
+    return this;
+  }
+
+  @Override
+  public String toString() {
+    return "SingularityBounceRequestBuilder{" +
+        "durationMillis='" + durationMillis + '\'' +
+        ", actionId='" + actionId + '\'' +
+        ", message=" + message + '\'' +
+        ", incremental=" + incremental + '\'' +
+        ", skipHealthchecks=" + skipHealthchecks + '\'' +
+        "}";
+  }
+}
--- a/SingularityBase/src/main/java/com/hubspot/singularity/expiring/SingularityExpiringBounce.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/expiring/SingularityExpiringBounce.java
@@ -20,6 +20,34 @@ public class SingularityExpiringBounce extends SingularityExpiringParent<Singula
     return deployId;
   }
 
+  public static Optional<SingularityExpiringBounce> withDefaultExpiringMillis(
+      Optional<SingularityExpiringBounce> maybeExpiringBounce,
+      Long durationMillis) {
+    if (!maybeExpiringBounce.isPresent()) {
+      return maybeExpiringBounce;
+    }
+    final SingularityExpiringBounce bounce = maybeExpiringBounce.get();
+    if (bounce.getExpiringAPIRequestObject().getDurationMillis().isPresent()) {
+      return maybeExpiringBounce;
+    }
+    return Optional.of(
+        new SingularityExpiringBounce(
+            bounce.getRequestId(),
+            bounce.getDeployId(),
+            bounce.getUser(),
+            bounce.getStartMillis(),
+            new SingularityBounceRequest(
+                bounce.getExpiringAPIRequestObject().getIncremental(),
+                bounce.getExpiringAPIRequestObject().getSkipHealthchecks(),
+                Optional.of(durationMillis),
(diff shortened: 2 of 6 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -385,7 +390,31 @@ public SingularityDeleteResult deleteLbCleanupRequest(String requestId) {
   }
 
   public Optional<SingularityExpiringBounce> getExpiringBounce(String requestId) {
-    return getExpiringObject(SingularityExpiringBounce.class, requestId);
+    final Optional<SingularityExpiringBounce> optionalBounce = getExpiringObject(SingularityExpiringBounce.class, requestId);
```
ssalinas: more of a convention thing, we tend to name vars for optionals as maybe..., so in this case `maybeExpiringBounce` or something similar

MattCCS: Will fix

tpetr:
- Are more than just expring bounces affected by the missing field issue? I feel like *all* expiring objects that have an absent `durationMillis` field could benefit.
- What happens when you create an expiring bounce with no `durationMillis` set, and then you redeploy Singularity with a different default value? Will the right thing happen in every case?
- Expanding on above, why not populate `durationMillis` right before it's written in ZK, instead of populating it after every read?

SingularityBase/src/main/java/com/hubspot/singularity/expiring/SingularityExpiringBounce.java
```diff
@@ -20,6 +20,34 @@ public String getDeployId() {
     return deployId;
   }
 
+  public static Optional<SingularityExpiringBounce> withDefaultExpiringMillis(
```
tpetr: we should make a `SingularityExpiringBounceBuilder` object instead... that way you can do something like:

```java
SingularityExpiringBounceBuilder.buildUpon(myExpiringBounceObject).setDurationMillis(Optional.of(7)).build()
```

tpetr: nevermind, i didn't realize this was so nested

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -201,8 +202,17 @@ public SingularityRequestParent bounce(@ApiParam("The request ID to bounce") @Pa
 
     requestManager.bounce(requestWithState.getRequest(), System.currentTimeMillis(), JavaUtils.getUserEmail(user), message);
 
+    SingularityBounceRequest defaultBounceRequest = bounceRequest.or(SingularityBounceRequest.defaultRequest());
+    if (!defaultBounceRequest.getDurationMillis().isPresent()) {
+      final Long durationMillis = TimeUnit.MINUTES.toMillis(configuration.getDefaultBounceExpirationMinutes());
```
tpetr: this is a nitpick, but you should utilize primitive types wherever possible (i.e. `long` instead of `Long`)

MattCCS: In this case I was following the convention established in the constructor of the `SingularityBounceRequest` class (`@JsonProperty("durationMillis") Optional<Long> durationMillis`).  If I change it here, should I change it in the constructor as well?

ssalinas: When using it as a type in the Optional<> you have to use the generic object (`Long`) in that case. A primitive does not extend any `Object` so it cannot be the type argument there. But, when it isn't required to have the generic object (like in your case, just setting a variable and using the value) it is preferred to use the primitive. You can pass that primitive (`long`) as an arg for something that takes `Long` just fine. (i.e. `Optional.of(yourVar)`)

tpetr: ^ 👍. The JVM's ability to hop between primitives (`long`) and "boxed" types (`Long`) is called auto-boxing. This is just one of those style things everyone learns the hard way 😃

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -201,8 +202,17 @@ public SingularityRequestParent bounce(@ApiParam("The request ID to bounce") @Pa
 
     requestManager.bounce(requestWithState.getRequest(), System.currentTimeMillis(), JavaUtils.getUserEmail(user), message);
 
+    SingularityBounceRequest defaultBounceRequest = bounceRequest.or(SingularityBounceRequest.defaultRequest());
```
ssalinas: last nitpick on this, since this can be the user-provided data or the default, `default` doesn't make sense as a variable name. Maybe something like `effectiveBounceRequest` (or something similar??)

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -80,14 +81,16 @@
...
+      Transcoder<SingularityExpiringScale> expiringScaleTranscoder,  Transcoder<SingularityExpiringPause> expiringPauseTranscoder, Transcoder<SingularityExpiringSkipHealthchecks> expiringSkipHealthchecksTranscoder,
+      SingularityConfiguration singularityConfiguration) {
     super(curator, configuration, metricRegistry);
     this.requestTranscoder = requestTranscoder;
     this.requestCleanupTranscoder = requestCleanupTranscoder;
     this.pendingRequestTranscoder = pendingRequestTranscoder;
     this.requestHistoryTranscoder = requestHistoryTranscoder;
     this.singularityEventListener = singularityEventListener;
     this.requestLbCleanupTranscoder = requestLbCleanupTranscoder;
+    this.singularityConfiguration = singularityConfiguration;
```
tpetr: it doesn't look like `singularityConfiguration` is in use in this class anymore, can you remove it?

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -436,4 +440,15 @@ private boolean isValidInteger(String strValue) {
       return false;
     }
   }
+
+  public SingularityBounceRequest validateBounceRequest(SingularityBounceRequest defaultBounceRequest) {
```
tpetr: I know this is a nitpick, but all the other validation methods use `check` instead of `validate` in this class.

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
MattCCS: Oops, sure thing.  The perils of moving methods around.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
MattCCS: Will fix!

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -80,7 +80,8 @@
   public RequestManager(CuratorFramework curator, SingularityConfiguration configuration, MetricRegistry metricRegistry, SingularityEventListener singularityEventListener,
       Transcoder<SingularityRequestCleanup> requestCleanupTranscoder, Transcoder<SingularityRequestWithState> requestTranscoder, Transcoder<SingularityRequestLbCleanup> requestLbCleanupTranscoder,
       Transcoder<SingularityPendingRequest> pendingRequestTranscoder, Transcoder<SingularityRequestHistory> requestHistoryTranscoder, Transcoder<SingularityExpiringBounce> expiringBounceTranscoder,
-      Transcoder<SingularityExpiringScale> expiringScaleTranscoder,  Transcoder<SingularityExpiringPause> expiringPauseTranscoder, Transcoder<SingularityExpiringSkipHealthchecks> expiringSkipHealthchecksTranscoder) {
+      Transcoder<SingularityExpiringScale> expiringScaleTranscoder,  Transcoder<SingularityExpiringPause> expiringPauseTranscoder, Transcoder<SingularityExpiringSkipHealthchecks> expiringSkipHealthchecksTranscoder,
+      SingularityConfiguration singularityConfiguration) {
```
tpetr: `singularityConfiguration` is not in use

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -66,6 +68,8 @@
 
   @Inject
   public SingularityValidator(SingularityConfiguration configuration, DeployHistoryHelper deployHistoryHelper, RequestManager requestManager) {
+    this.configuration = configuration;
```
tpetr: I'd mimic what we're doing everywhere else in this class and just have a variable for the specific configuration value you need (in this case, `defaultBounceExpirationMinutes`).

tpetr:
LGTM after last 2 comments are addressed

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
MattCCS: And define this variable in SingularityConfiguration?

tpetr: the variable is already in SingularityConfiguration, it's accessed via `configuration.getDefaultBounceExpirationMinutes()`

MattCCS:
Both comments have been addressed.  Does this look good to go?

ssalinas:
Looks good @MattCCS

tpetr:
LGTM, good to merge once this has been re-merged / redeployed through the 3 environments

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityBounceRequestBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return "SingularityBounceRequestBuilder{" +
+        "durationMillis='" + durationMillis + '\'' +
+        ", actionId='" + actionId + '\'' +
+        ", message=" + message + '\'' +
+        ", incremental=" + incremental + '\'' +
+        ", skipHealthchecks=" + skipHealthchecks + '\'' +
```
darcatron: nit: you might have changed this already, but the last three parts are missing the first apostrophe `", incremental="` => `", incremental='"`

MattCCS: Will fix

PR #1198, opened 2016-08-09
Add ability to redeploy from the UI

### LINKED ISSUE #1079: Feature Request - Ability to re-deploy previous deployments

Calvinp:
Basing this on `run_now_fixes` because it uses the `doBefore()` functionality introduced in that branch to load the data for the deploy being repeated.

The modal looks like this:
Two things to notice:

Fixes #1079.
cc @tpetr @kwm4385 @wolfd


kwm4385:
Nice! 
Nit: The text might be a bit cleaner if it just said `(New deploy ID must be unique)`. I think that still conveys the same message.

SingularityUI/app/components/common/modalButtons/RedeployModal.jsx
```diff
@@ -0,0 +1,106 @@
...
+  static propTypes = {
+    requestId: PropTypes.string.isRequired,
+    saveDeploy: PropTypes.func,
+    doAfterRedeploy: PropTypes.func,
+    deploy: PropTypes.object
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {};
```
kwm4385: no-op here

Calvinp: It's not a no-op - it sets initial state to be an empty object rather than null, so that a bunch of things in `render()` don't need to check if state exists.

PR #1199, opened 2016-08-09
sentry support

kwm4385:
This PR allows sentry to support the UI by surfacing errors out of the main rendering promise in `rootComponent`. Sentry can't detect errors inside promises because they aren't bubbled up to window.onerror. (Actual sentry tracking code is injected normally through `SINGULARITY_EXTRA_SCRIPTS`.

@tpetr @Calvinp @wolfd

### DIFF
--- a/SingularityUI/app/assets/index.mustache
+++ b/SingularityUI/app/assets/index.mustache
@@ -47,10 +47,10 @@
         };
     </script>
     <script src="{{{staticRoot}}}/js/vendor.bundle.js"></script>
-    <script src="{{{staticRoot}}}/js/app.js"></script>
     {{#extraScript}}
     {{{extraScript}}}
     {{/extraScript}}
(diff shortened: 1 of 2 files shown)


SingularityUI/app/assets/index.mustache
```diff
@@ -47,10 +47,10 @@
         };
     </script>
     <script src="{{{staticRoot}}}/js/vendor.bundle.js"></script>
-    <script src="{{{staticRoot}}}/js/app.js"></script>
     {{#extraScript}}
     {{{extraScript}}}
     {{/extraScript}}
+    <script src="{{{staticRoot}}}/js/app.js"></script>
```
tpetr: will this ordering affect things like our Heap integration? /cc @wolfd

kwm4385: I don't think so (wolf can verify). Sentry does need to be loaded before our code though.

PR #1200, opened 2016-08-10
Aggregate Tailer fix

kwm4385:
Fixes the case where the tailer wouldn't respect the taskIds query param when viewing a log via the `/request/*/tail` path. (Example: https://private.hubteamqa.com/singularity-test/request/ORION_TestService_TrafficSplitting-userweb/tail/ORION_TestService_TrafficSplitting-userweb-152_15_19-1470241646336-3-flaky_lion.iad03.hubspot_networks.net-us_east_1e/tail_of_finished_service.log?taskIds=ORION_TestService_TrafficSplitting-userweb-152_15_19-1470241646336-3-flaky_lion.iad03.hubspot_networks.net-us_east_1e)

@tpetr @Calvinp @wolfd

### DIFF
--- a/SingularityUI/app/actions/log.es6
+++ b/SingularityUI/app/actions/log.es6
@@ -242,11 +242,11 @@ export const taskGroupFetchPrevious = taskGroupId =>
   function(dispatch, getState) {
     let {tasks, taskGroups, logRequestLength, maxLines} = getState();
 
-    let taskGroup = taskGroups[taskGroupId];
+    const taskGroup = taskGroups[taskGroupId];
     tasks = getTasks(taskGroup, tasks);
 
     // bail early if all tasks are at the top
-    if (_.all(tasks.map(({minOffset}) => minOffset === 0))) {
+    if (_.all(tasks.map((task) => task.minOffset === 0))) {
       return Promise.resolve();
     }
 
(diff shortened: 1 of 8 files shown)


SingularityUI/app/components/logs/LogLines.jsx
```diff
@@ -58,13 +58,14 @@ class LogLines extends React.Component {
         taskId={taskId}
         timestamp={timestamp}
         isHighlighted={offset === initialOffset}
-        color={colorMap[taskId]} />;
+        color={colorMap[taskId]}
```
tpetr: might be misreading, but doesn't this need the `/>`?

kwm4385:
WIP: Fixing other bugs with the aggregate view (auto decaffeinate did a number on it)

kwm4385:
Everything seems to be working except for the unified instance switching, but as discussed in Slack, this is already affecting QA and PROD.

SingularityUI/app/actions/log.es6
```diff
@@ -256,6 +256,7 @@ export const taskGroupFetchPrevious = taskGroupId =>
     }
 
     dispatch({taskGroupId, type: 'LOG_REQUEST_START'});
+    tasks = _.without(tasks, undefined);
```
tpetr: just curious, do we know why `undefined` was here in the first place?

kwm4385: I'm not sure, there seems to be an issue with how the actions are removing instances from the list but I wasn't able to pinpoint where. This might actually be a no op since this is related to the same problem on QA and PROD.

tpetr:
LGTM

PR #1192, opened 2016-08-05
Code cleanliness improvements

Calvinp:
- Renames most one and two letter variable names to things that are more descriptive.
- Adds an eslint rule to disallow most one and two letter variable names. `id`, `ui`, and `s3` are ok because they are useful and informative when they are used, unlike `hc` (for `healthCheck`)
- Adds a block exception to this rule for the moment locale rules.
- Fixes a spot where props were mutated.
- Uses `moment.updateLocale()` instead of `moment.locale()` to get rid of the deprecated warning.
- Makes several components functional.

PR #1201, opened 2016-08-10
Cleanup Webhooks Page

Calvinp:
This makes the following changes to the webhooks page to bring it more in line with the rest of the codebase:

- Brings the new and delete webhooks modals out into their own components.
- Makes the top-level Webhooks page a functional component.

cc @tpetr @kwm4385 @wolfd


Calvinp:
Folding this into #1192

PR #1202, opened 2016-08-10
Add case-sensitive paths plugin

Calvinp:
This adds the [case-sensitive paths plugin](https://www.npmjs.com/package/case-sensitive-paths-webpack-plugin) Singularity UI - thus preventing the issue where builds succeed locally but not in Jenkins because the OSX filesystem is case-insensitive.
Now it'll fail locally with this error:
```
ERROR in ./app/components/tasks/TasksPage.jsx
Module not found: Error: [CaseSensitivePathsPlugin] `/Users/cpomerantz/repos/Singularity/SingularityUI/app/rootcomponent.jsx` does not match the corresponding path on disk `rootComponent.jsx`.
 @ ./app/components/tasks/TasksPage.jsx 17:21-51
```

cc @tpetr @wolfd @kwm4385

### DIFF
--- a/SingularityUI/package.json
+++ b/SingularityUI/package.json
@@ -68,6 +68,7 @@
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-plugin-transform-react-jsx": "^6.8.0",
     "babel-preset-es2015": "^6.9.0",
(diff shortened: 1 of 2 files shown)


tpetr:
LGTM

PR #1205, opened 2016-08-11
Refresh request detail after actions

Calvinp:
After performing an action from the request detail page, refresh the page so that the effects of the action can be seen.

The exception is removing the request - on performing that action, the user will be navigated to the requests page.

Omission from initial PR: It may be nice to add similar capabilities to the requests table, so that, for example, when you remove a request successfully it disappears from the table.

cc @tpetr @kwm4385 @wolfd


Calvinp:
Oops, accidentally branched this off of staging instead of decaf.

PR #1207, opened 2016-08-11
UI support for absent deploy field

kwm4385:
In cases where the `deploy` field was absent from the deploy object, the UI would fail. This PR short circuits the render to display an empty message in this case.

image

@tpetr @wolfd @Calvinp @ssalinas


ssalinas:
@kwm4385 we should still be able to show the task history. just wouldn't have the `Info` section

kwm4385:
Readded those sections
image

tpetr:
LGTM

PR #1208, opened 2016-08-11
Fix duration field overflowing modal in firefox

Calvinp:
This fixes the bug that caused duration fields to overflow the modal in firefox.
Before:
screenshot 2016-08-11 15 57 40

After:
screenshot 2016-08-11 16 36 37

cc @tpetr @wolfd @kwm4385


tpetr:
LGTM

PR #1209, opened 2016-08-11
Name all the modals that don't already have a name

Calvinp:
Exactly what it says on the tin - gives a name to all the modals that didn't already have a name.

eg.
screenshot 2016-08-11 16 50 18

cc @tpetr @wolfd @kwm4385


tpetr:
LGTM

PR #1211, opened 2016-08-12
Render every row on dashboard tables

Calvinp:
Renders every row on the `My Paused Requests` and `My Starred Requests` tables on the dashboard page.

Note that this will render every row even if there are thousands - but there won't be thousands unless a user stars every request in the system for some reason.

cc @tpetr @wolfd @kwm4385

### DIFF
--- a/SingularityUI/app/components/common/table/UITable.jsx
+++ b/SingularityUI/app/components/common/table/UITable.jsx
@@ -452,7 +452,7 @@ UITable.propTypes = {
   keyGetter: PropTypes.func.isRequired,
   children: PropTypes.arrayOf(PropTypes.node).isRequired,
   paginated: PropTypes.bool,
-  renderAllRows: PropTypes.bool, // Will render all rows even if there are 1,000,000
+  renderAllRows: PropTypes.bool,
   rowChunkSize: PropTypes.number,
   rowChunkSizeChoices: PropTypes.arrayOf(PropTypes.number),
   maxPaginationButtons: PropTypes.number,
--- a/SingularityUI/app/components/common/table/UITable.jsx
+++ b/SingularityUI/app/components/common/table/UITable.jsx
@@ -288,7 +288,7 @@ class UITable extends Component {
     // infinite scrolling
     // Only render a number of rows at a time
     // check to see if we can render of everything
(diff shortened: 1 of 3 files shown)


SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -452,6 +452,7 @@ UITable.propTypes = {
   keyGetter: PropTypes.func.isRequired,
   children: PropTypes.arrayOf(PropTypes.node).isRequired,
   paginated: PropTypes.bool,
+  isTruelyInfinite: PropTypes.bool, // Will render all rows even if there are 1,000,000
```
tpetr: can we think of a better name for this prop type? we shouldn't be depending on comments to explain what a prop does. furthermore, why cant `paginated` set to `false` support our use case here?

Calvinp: I'll respond to the second point first. The problem is there are now three pagination options, so a boolean can't encompass them all:
- Paginated (this can be split into api paginated and local paginated, but that's not relevant here)
- Not paginated, but uses infinite scrolling (which is what the requests and tasks tables do)
- Not paginated, renders all rows no matter what (which is introduced in this PR and is what the `My Starred Requests` table will do)

The third pagination type is implemented because we want to render all starred requests for ctrl-f access - which won't happen if the user has more than a certain number if we're using infinite scrolling.

So to be more descriptive with the name, maybe `renderAllRowsImmediately` or something like that would be a better name.

tpetr: Thanks for the clarification -- let's go with `renderAllRows`

SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -452,6 +452,7 @@ UITable.propTypes = {
   keyGetter: PropTypes.func.isRequired,
   children: PropTypes.arrayOf(PropTypes.node).isRequired,
   paginated: PropTypes.bool,
+  renderAllRows: PropTypes.bool, // Will render all rows even if there are 1,000,000
```
tpetr: comment is unnecessary now

tpetr:
LGTM after last piece of feedback

PR #1212, opened 2016-08-12
allow searching of deleted request logs

ssalinas:
The prefixes and files for deleted requests will still be in s3, so we should be able to search them. Updates to just return an absent group if we can't find the request and continue our searching.

@tpetr


ssalinas:

PR #1215, opened 2016-08-12
Check for presence of promise before attempting to catch errors

kwm4385:
Refresh doesn't require a promise to always be returned, so check for one before attempting to catch errors for sentry. Fixes "catch of undefined sentry".

@tpetr @wolfd @Calvinp


tpetr:
LGTM

PR #1216, opened 2016-08-12
Redeploy

### LINKED ISSUE #1079: Feature Request - Ability to re-deploy previous deployments

Calvinp:
Reopening based on decaf.

Basing this on `run_now_fixes` because it uses the `doBefore()` functionality introduced in that branch to load the data for the deploy being repeated.

The modal looks like this:
Two things to notice:

Fixes #1079.
cc @tpetr @kwm4385 @wolfd


SingularityUI/app/components/common/modalButtons/RedeployButton.jsx
```diff
@@ -0,0 +1,64 @@
...
+import RedeployModal from './RedeployModal';
+import Utils from '../../../utils';
+
+const redeployTooltip = (
+  <ToolTip id="redeploy">
+    Redeploy
+  </ToolTip>
+);
+
+class RedeployButton extends Component {
```
kwm4385: Components like this without lifecycle methods or state can be rewritten as stateless components.

Calvinp: I have a stretch goal of rewriting all of these buttons. Having refs and using `getClickComponent()` makes it complicated, but I know I can do it anyway.

kwm4385: Yeah that's actually related to a problem I've run into, where I want to use the buttons in a dropdown but clicking anything in the modal closes it because the modal is a child of the dropdown.

kwm4385: Stateless wouldn't necessarily solve that problem though, I'll have to find a way to stop click events from bubbling up.

SingularityUI/app/components/common/modalButtons/RedeployModal.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+  show() {
+    this.refs.redeployModal.show();
+  }
+
+  confirmRedeploy(data) {
+    this.props.saveDeploy(data.deployId).then((response) => {
+      if (response.statusCode === 200) {
+        this.setState({newDeployId: data.deployId});
+        this.showSuccessModal();
```
kwm4385: Instead of having a full-fledged modal to show success that the user then has to dismiss, would it make more UX sense to display a `Messenger` notification instead?

kwm4385:
LGTM besides comments

SingularityUI/app/components/common/modalButtons/RedeployButton.jsx
Calvinp: This is how I did it: https://github.com/HubSpot/Singularity/blob/0e65a7407b9c1a9eee0417d20aabd8d68ceb5b2d/SingularityUI/app/components/common/modalButtons/NewWebhookButton.jsx

That doesn't help with your dropdown problem though. (Anyway, remember @markhazlewood doesn't like dropdowns - I agree with his reasoning as to why)

Calvinp: As it turns out, having a `doFirst()` makes my cheat not work for some reason. I'm trying to figure out why.

SingularityUI/app/components/common/modalButtons/RedeployModal.jsx
Calvinp: Either that, or simply redirect to the page of the new deploy (but that might not always be what the user wants).

Calvinp:
Closing in favor of #1227

PR #1218, opened 2016-08-15
Use direct string matching with glob patterns instead of fuzzy matching

Calvinp:
Gets rid of fuzzy matching on the requests page, using exact string matching instead (or glob matching if the user specifies a glob pattern).

TODO:
- [x] Implement glob matching on the tasks page too.
- [x] Make it clear to the user that glob matching is a possibility.
- [ ] Implement glob matching with highlighting on global search (Maybe).

Note that matches at the start of a requestId are prioritized over matches anywhere else (including at the start of the user field).

cc @tpetr @kwm4385 @wolfd


Calvinp:
I added a tooltip with a click overlay trigger to explain to the user that they can use `*` if they need wildcard matching:
screenshot 2016-08-15 16 55 24

kwm4385:
I wouldn't be opposed to having the tip in the placeholder, but I'm not sold either way.

tpetr:
I'm not hugely wild about that approach because it seems hard for people to discover, and then once they know about it, it'll be an annoyance whenever it shows up. What do you guys think about a ? icon somewhere nearby that either shows a tooltip or pops open a modal that explains how filtering works? That way people can find it easily, and only engage it when they need it.

Calvinp:
I like that idea better than what I have now.

Calvinp:
Once user settings are fully implemented, we could even show a modal the first time a user goes to the requests or tasks page - then when they close the modal we send an API call to never show it again.

That might have some pitfalls, but at least we can start thinking about things like that now.

Calvinp:
This is what I have now:
screenshot 2016-08-16 09 12 29
screenshot 2016-08-16 09 12 39

I would not be opposed to moving the `?` button if someone has a better idea of where to put it.

kwm4385:
Maybe have it inline like the clear button? I'd put it right before it.

tpetr:
- Put it on the same row as the request type toggles. Not sure if left or right aligned would look better.
- "Did you know?" doesn't really match the tone of the UI. I'd suggest titling it "Filtering requests" or not having a title at all.
- We should include more information about how the search works. If we still have fuzzy matching enabled, explain that, etc...

Calvinp:
Button on the left:
screenshot 2016-08-16 09 25 47

Button centered:
screenshot 2016-08-16 09 25 27

Button on the right:
screenshot 2016-08-16 09 32 25


I personally like the centered version the best.

New modal:
screenshot 2016-08-16 09 36 28

Calvinp:
Added it to the tasks page too:
screenshot 2016-08-16 11 11 40


screenshot 2016-08-16 11 10 23

Calvinp:
Closed in favor of #1228

PR #1219, opened 2016-08-15
Starred requests persistence

MattCCS:
This PR enables "starred" Singularity requests to persist in Zookeeper, rather than be stored client-side, so clearing the cache or changing devices won't require re-starring.

/cc @ssalinas @tpetr


Calvinp:
Are we still planning on having endpoints to add or remove one star at a time?

ssalinas:
@MattCCS as part of the implementation, I believe we also discussed adding the `SingularityUserSettings` into the `SingularityUserHolder` object so we can grab things on the /auth/user call when possible. Can we add that in?

SingularityService/src/main/java/com/hubspot/singularity/resources/UserResource.java
```diff
@@ -0,0 +1,44 @@
...
+      @ApiParam("The user id to use") @QueryParam("userId") String userId) {
+    return userManager.getUserSettings(SingularityValidator.encodeZkName(userId));
+  }
+
+  @POST
+  @Path("/settings")
+  public void setUserSettings(
+      @ApiParam("The user id to use") @QueryParam("userId") String userId,
+      @ApiParam("The new settings") SingularityUserSettings settings) {
+    userManager.updateUserSettings(SingularityValidator.encodeZkName(userId), settings);
```
tpetr: this ZK encoding stuff should be inside `UserManager` -- the point of it is to encapsulate away all ZK-related stuff

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -436,4 +437,12 @@ private boolean isValidInteger(String strValue) {
       return false;
     }
   }
+
+  public static String encodeZkName(String name) {
+    checkBadRequest(name != null, "Name may not be a null value");
```
ssalinas: Can use `Strings.isNullOrEmpty(name)` to hit both these lines in one go

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -436,4 +437,12 @@ private boolean isValidInteger(String strValue) {
       return false;
     }
   }
+
+  public static String encodeZkName(String name) {
```
ssalinas: All our other method names are named in the format `check...`, maybe `checkUserId` or `checkAndEncodeUserId` to show what's being returned + be consistent with other methods in the class

SingularityService/src/main/java/com/hubspot/singularity/data/UserManager.java
```diff
@@ -0,0 +1,44 @@
...
+  private static final String SETTINGS_ROOT = USER_ROOT + "/settings";
+
+  @Inject
+  public UserManager(CuratorFramework curator, SingularityConfiguration configuration, MetricRegistry metricRegistry,
+                     Transcoder<SingularityUserSettings> settingsTranscoder) {
+    super(curator, configuration, metricRegistry);
+    this.settingsTranscoder = settingsTranscoder;
+  }
+
+  private String getUserSettingsPath(String id) {
```
ssalinas: nit, but would be good to use `userId` (or something else more descriptive than id) as the variable name. There are many things with ids that get passed around Singularity

MattCCS:
@Calvinp I think we're instead going to do the unioning client-side, and post the entire union back to ZK.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
MattCCS: Oh, handy, thanks.

SingularityService/src/main/java/com/hubspot/singularity/resources/UserResource.java
MattCCS: Will fix.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
MattCCS: Right, my bad, will fix.

SingularityService/src/main/java/com/hubspot/singularity/data/UserManager.java
MattCCS: Right, will fix.

SingularityService/src/main/java/com/hubspot/singularity/resources/UserResource.java
```diff
@@ -0,0 +1,43 @@
...
+  @Path("/settings")
+  public Optional<SingularityUserSettings> getUserSettings(
+      @ApiParam("The user id to use") @QueryParam("userId") String userId) {
+    return userManager.getUserSettings(userId);
+  }
+
+  @POST
+  @Path("/settings")
+  public void setUserSettings(
+      @ApiParam("The user id to use") @QueryParam("userId") String userId,
```
Calvinp: I don't believe we use query params in `POST` requests anywhere else.

kwm4385: Yep those should be part of the request body.

Calvinp: This is going to be fixed by adding userId to the settings object itself.

SingularityService/src/main/java/com/hubspot/singularity/resources/UserResource.java
```diff
@@ -0,0 +1,54 @@
...
+  @POST
+  @Path("/settings")
+  public void setUserSettings(@ApiParam("The new settings") SingularityUserSettings settings) {
+    userManager.updateUserSettings(settings.getUserId(), settings);
+  }
+
+  @POST
+  @Path("/settings/starred-requests")
+  public void addStarredRequests(@ApiParam("The new starred requests") SingularityUserSettings settings) {
+    userManager.addStarredRequestIds(settings.getUserId(), settings.getStarredRequestIds());
```
Calvinp: I wonder if it makes sense to explicitly accept a new object that has an array of request ids and a user id rather than the entire settings object (which will eventually have more parameters, which might end up being required).

MattCCS: @ssalinas We had discussed earlier but I believe the reasoning against it was that a "S..StarredRequestsRequest" would basically be an exact subset of "S..UserSettings", right?

Calvinp: It would be a subset, but now if some required field gets added to settings the UI will have to include that field to add or delete starred requests.

MattCCS:
Update: merged master branch back into PR after decaf was merged into master (i.e. tested and working on decaf).

SingularityService/src/main/java/com/hubspot/singularity/data/UserManager.java
```diff
@@ -0,0 +1,76 @@
...
+      save(path, new SingularityUserSettings(starredRequestIds), settingsTranscoder);
+      return;
+    }
+    save(path, settings.get().addStarredRequestIds(starredRequestIds), settingsTranscoder);
+  }
+
+  public void deleteStarredRequestIds(String userId, Set<String> starredRequestIds) {
+    final String path = getUserSettingsPath(userId);
+    final Optional<SingularityUserSettings> settings = getData(path, settingsTranscoder);
+    if (!settings.isPresent()) {
```
tpetr: do we even need to save if user settings aren't present?

MattCCS: I don't believe so, no!  Will fix

SingularityService/src/main/java/com/hubspot/singularity/resources/UserResource.java
```diff
@@ -0,0 +1,63 @@
...
+  public UserResource(UserManager userManager,
+                      Optional<SingularityUser> user) {
+    this.userManager = userManager;
+    this.user = user;
+  }
+
+  private String getAuthUserId() {
+    checkBadRequest(user.isPresent() && user.get().getName().isPresent(), "Singularity userId must be provided by auth");
+    return user.get().getName().get();
+  }
```
tpetr: This method needs some work:
- The error message isn't accurate because we're not using the `userId` query param anymore.
- Being unauthenticated should return an Unauthorized response, not a Bad Request response. Use SingularityAuthorizationHelper as a guide here; it uses `checkUnauthorized(user.isPresent(), "Please log in to perform this action.");` for stuff like this.
- It's using the `name` field of SingularityUser (`Thomas Petr`) when it should be using the `id` field (`tpetr`).

MattCCS: Ah, my bad.  So, should I be checking `authEnabled` before I do the `checkAuthorized()` call (as happens in SingularityAuthorizationHelper)?  Or should I just be checking auth every time?

ssalinas:
@tpetr any problem with merging this one now?

PR #1220, opened 2016-08-16
[logtail] Flush stdout after printing logs

marcob:
This will make `logtail -r REQUEST | grep something` work as expected


ssalinas:
👍 thanks

PR #1221, opened 2016-08-16
Get count for blended history calls

kwm4385:
Not knowing how many total entries there are makes pagination ugly for many applications, including the frontend. The plan is to first have separate endpoints for each paginated one that uses blended history. Eventually, we can convert the old endpoints to return objects with metadata including total count instead of a simple array.

**WIP:**
As of right now, I've implemented a count endpoint for deploy history on a request as a proof of concept. Feedback on style and implementation would be appreciated!

Update: Added count for task history

@tpetr @ssalinas


SingularityService/src/main/java/com/hubspot/singularity/data/history/BlendedHistoryHelper.java
```diff
@@ -25,6 +27,10 @@
   protected abstract List<T> getFromZk(Q id);
   protected abstract List<T> getFromHistory(Q id, int historyStart, int numFromHistory);
 
+  protected Optional<Integer> getTotalCount(Q id) {
+    throw new NotImplementedException();
+  }
```
kwm4385: Once this method is implemented in all the subclasses, this could be changed to `abstract` like the others.

ssalinas:
Overall this seems pretty good to me. My only concern might be how long it's going to take to get that result back on the mysql side (not for deploys, but if we do anything similar for task history), will have to speed test it a bit and/or make sure this call isn't blocking anything while we wait for it.

kwm4385:
Testing with small datasets, the task history count is taking <1ms to process the entire request. We'll still need to test it with lots more data than I can provide though.

kwm4385:
Created `SingularityPaginatedResponse` to encapsulate the array and metadata for the new endpoints. Responses look like:
`GET http://192.168.59.103:7099/singularity/api/history/request/test/tasks/count?count=3&page=1`

```
{
   "dataCount":9,
   "pageCount":3,
   "page":1,
   "data":[
      {
         "taskId":{
            "requestId":"test",
            "deployId":"9",
            "startedAt":1471463710973,
            "instanceNo":1,
            "host":"localhost",
            "sanitizedHost":"localhost",
            "sanitizedRackId":"DEFAULT",
            "rackId":"DEFAULT",
            "id":"test-9-1471463710973-1-localhost-DEFAULT"
         },
         "updatedAt":1471463734056,
         "lastTaskState":"TASK_KILLED"
      },
      {
         "taskId":{
            "requestId":"test",
            "deployId":"8",
            "startedAt":1471463692948,
            "instanceNo":1,
            "host":"localhost",
            "sanitizedHost":"localhost",
            "sanitizedRackId":"DEFAULT",
            "rackId":"DEFAULT",
            "id":"test-8-1471463692948-1-localhost-DEFAULT"
         },
         "updatedAt":1471463724040,
         "lastTaskState":"TASK_KILLED"
      },
      {
         "taskId":{
            "requestId":"test",
            "deployId":"7",
            "startedAt":1471463680924,
            "instanceNo":1,
            "host":"localhost",
            "sanitizedHost":"localhost",
            "sanitizedRackId":"DEFAULT",
            "rackId":"DEFAULT",
            "id":"test-7-1471463680924-1-localhost-DEFAULT"
         },
         "updatedAt":1471463703986,
         "lastTaskState":"TASK_KILLED"
      }
   ]
}
```

TODO: Unspecified `count` throws NPE

SingularityService/src/main/java/com/hubspot/singularity/data/history/HistoryJDBI.java
```diff
@@ -153,4 +158,58 @@ private void addWhereOrAnd(StringBuilder sqlBuilder, boolean shouldUseWhere) {
     return query.list();
   }
 
+  public int getTaskIdHistoryCount(Optional<String> requestId, Optional<String> deployId, Optional<String> host,
+      Optional<ExtendedTaskState> lastTaskStatus, Optional<Long> startedBefore, Optional<Long> startedAfter) {
+
+    final Map<String, Object> binds = new HashMap<>();
+    final StringBuilder sqlBuilder = new StringBuilder(GET_TASK_ID_HISTORY_COUNT_QUERY);
+
+    if (requestId.isPresent()) {
```
ssalinas: rather than rewriting all of this from the method above, maybe make it one private method that get... and get..Count can call, just with different SELECT statements at the front?

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -215,6 +238,22 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
   }
 
   @GET
+  @Path("/request/{requestId}/deploys/count")
```
ssalinas: now that we are returning a paginated response with full data, not sure /count still makes sense as an endpoint name

kwm4385: For sure, we'll be deprecating the old one but still need both concurrently so we'll still need a way to differentiate.

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -215,6 +238,22 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
...
+  @ApiOperation("Get deploy history with metadata for a single request")
+  public SingularityPaginatedResponse<SingularityDeployHistory> getDeploysCount(
+          @ApiParam("Request ID to look up") @PathParam("requestId") String requestId,
+          @ApiParam("Maximum number of items to return") @QueryParam("count") Integer count,
+          @ApiParam("Which page of items to view") @QueryParam("page") Integer page) {
+    authorizationHelper.checkForAuthorizationByRequestId(requestId, user, SingularityAuthorizationScope.READ);
+
+    final int dataCount = deployHistoryHelper.getBlendedHistoryCount(requestId).get();
```
ssalinas: Maybe the count should just be optional on the response object rather than risking the NPE. Would be better for the ui to just handle rendering without a count than to have a bogus count or get an error probably

ssalinas:
Finished up the `withmetadata` endpoints and the `getTotalCount` implementations to make that helper method an abstract like Kevin mentioned. Still needs some testing and any frontend changes to use the new metadata

SingularityBase/src/main/java/com/hubspot/singularity/SingularityPaginatedResponse.java
```diff
@@ -0,0 +1,36 @@
...
+  private final Optional<Integer> dataCount;
+  private final Optional<Integer> pageCount;
+  private final Optional<Integer> page;
+  private final List<Q> data;
+
+  public SingularityPaginatedResponse(final Optional<Integer> dataCount, final Optional<Integer> pageCount, final Optional<Integer> page, final List<Q> data) {
+    this.dataCount = dataCount;
+    this.pageCount = pageCount;
+    this.page = page;
+    this.data = data;
```
tpetr: can we standardize on `objects` instead of `data`? (we're using objects in other places already)

SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -215,6 +267,23 @@ public SingularityDeployHistory getDeploy(@ApiParam("Request ID for deploy") @Pa
...
+          @ApiParam("Maximum number of items to return") @QueryParam("count") Integer count,
+          @ApiParam("Which page of items to view") @QueryParam("page") Integer page) {
+    authorizationHelper.checkForAuthorizationByRequestId(requestId, user, SingularityAuthorizationScope.READ);
+
+    final Optional<Integer> dataCount = deployHistoryHelper.getBlendedHistoryCount(requestId);
+    final int limitCount = getLimitCount(count);
+    final List<SingularityDeployHistory> data = this.getDeploys(requestId, count, page);
+    final Optional<Integer> pageCount = getPageCount(dataCount, limitCount);
+
+    return new SingularityPaginatedResponse<>(dataCount, pageCount, Optional.fromNullable(page), data);
```
tpetr: shouldn't `page` default to 1?

ssalinas:
@tpetr addressed both comments (data -> objects and default page)

ssalinas:
Will tackle the frontend of this in a separate PR, going to merge this one

PR #1222, opened 2016-08-16
Starred requests persistence frontend

Calvinp:
Implements the frontend for #1219 
TODO:
- [x] Add the ability to import starred requests.
- [x] Implement UI changes to go along with various backend upgrades that are coming down the pipeline for this.
- [x] Right now the stars all blink when changing a star. Fix that.
- [x] Delete stars stored locally after importing them.
- [x] Fix slowness.
- [x] Fix a nasty bug that causes all old starred requests to be deleted from API in some cases in which a user switches browsers.

cc @tpetr @ssalinas @MattCCS


Calvinp:
To fix the fact that there API calls take a fairly significant amount of time, which can lead to frustration waiting for the clicked star to show up as being starred, the Ui uses temporary settings. These get set right before the API call is sent, and cleared after it returns or errors. Then whenever settings are used, temporary settings are checked first.

Calvinp:
This is ready to be reviewed and merged.

Calvinp:
Closed in favor of #1229

PR #1225, opened 2016-08-18
Fix the back button when navigating through files on the task detail page

Calvinp:
Fixes the bug that caused the back button not to work while navigating through file system on the TaskDetail page.

The problem was it wasn't actually using the route to determine the file path, but to set state that determined the file path. The state wouldn't be updated when the back button was hit. The fix was to not use state at all.

cc @tpetr @kwm4385 @wolfd

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskDetail.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskDetail.jsx
@@ -108,7 +108,7 @@ class TaskDetail extends Component {
       files: PropTypes.array,
       currentDirectory: PropTypes.string
     }),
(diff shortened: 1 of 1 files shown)


kwm4385:
Beat me to it lol
LGTM

Calvinp:
This actually turned out to be a great example for why [using props to set initial state is bad](https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html).

jonathanwgoodwin:
Thanks @Calvinp !

tpetr:
LGTM

PR #1226, opened 2016-08-18
Multiple deploy groups

darcatron:
Provides devs with ability to give deploy access to multiple groups.

@ssalinas @tpetr


SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
```diff
@@ -142,7 +142,7 @@ public void checkForAuthorization(SingularityRequest request, Optional<Singulari
 
     final boolean userIsAdmin = adminGroups.isEmpty() ? false : groupsIntersect(userGroups, adminGroups);
     final boolean userIsJITA = jitaGroups.isEmpty() ? false : groupsIntersect(userGroups, jitaGroups);
-    final boolean userIsRequestOwner = request.getGroup().isPresent() ? userGroups.contains(request.getGroup().get()) : true;
+    final boolean userIsRequestOwner = request.getGroup().isPresent() ? groupsIntersect(userGroups, request.getReadWriteGroups().get()) : true;
```
tpetr: we still need to be checking `request.getGroup()` (so essentially, the union of `group` and `readWriteGroups`)

darcatron:
@tpetr @ssalinas

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
```diff
@@ -138,13 +143,17 @@ public void checkForAuthorization(SingularityRequest request, Optional<Singulari
     checkUnauthorized(user.isPresent(), "user must be present");
 
     final Set<String> userGroups = user.get().getGroups();
+    final Set<String> readWriteGroups = request.getReadWriteGroups().or(Collections.<String>emptySet());
+    if (request.getGroup().isPresent()) {
+      readWriteGroups.add(request.getGroup().get());
```
tpetr: `Collections.emptySet()` is immutable, unfortunately, and we generally shouldn't be mutating stuff in `SingularityRequest`. I'd suggest just using `Sets.union()` to generate a the union here.

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
```diff
@@ -112,12 +113,16 @@ public boolean isAuthorizedForRequest(SingularityRequest request, Optional<Singu
...
+    }
 
-    final boolean userIsAdmin = adminGroups.isEmpty() ? false : groupsIntersect(userGroups, adminGroups);
-    final boolean userIsJITA = jitaGroups.isEmpty() ? false : groupsIntersect(userGroups, jitaGroups);
-    final boolean userIsRequestOwner = request.getGroup().isPresent() ? userGroups.contains(request.getGroup().get()) : true;
+    final boolean userIsAdmin = !adminGroups.isEmpty() && groupsIntersect(userGroups, adminGroups);
+    final boolean userIsJITA = !jitaGroups.isEmpty() && groupsIntersect(userGroups, jitaGroups);
```
tpetr: Nice optimization 🌃

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
darcatron: My bad. I didn't see that 

I can't really use `Sets.union()` because `request.getGroup().get()` is just a String. I can just create a `new HashSet<>()` and add whatever `isPresent()` to it

tpetr: No worries -- you can still use `Sets.union()` by coercing the `Optional<>` to a `Set<>` via `request.getGroup().asSet()`

darcatron: 🍠 that's great!

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
```diff
@@ -138,13 +140,14 @@ public void checkForAuthorization(SingularityRequest request, Optional<Singulari
     checkUnauthorized(user.isPresent(), "user must be present");
 
     final Set<String> userGroups = user.get().getGroups();
+    final SetView readWriteGroups = Sets.union(request.getGroup().asSet(), request.getReadWriteGroups().asSet());
```
tpetr: nitpick but `SetView<>` implements `Set<>`, so for simplicity just use a `Set<String>` here

tpetr: (also surprised that the compiler didn't scream at you for not doing `SetView<String>`)

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
```diff
@@ -239,12 +239,12 @@ public boolean isDeployable() {
...
   @JsonIgnore
   public boolean isRackSensitive() {
-    return rackSensitive.or(Boolean.FALSE).booleanValue();
+    return rackSensitive.or(Boolean.FALSE);
   }
 
   @JsonIgnore
   public boolean isLoadBalanced() {
-    return loadBalanced.or(Boolean.FALSE).booleanValue();
+    return loadBalanced.or(Boolean.FALSE);
```
tpetr: i think it'd be even smoother to refactor this down to `loadBalanced.or(false)`

tpetr:
One last tweak that I recently realized -- we should only allow admins and owners (i.e. members of `group`) add or remove groups from `readWriteGroups`. Once that's squared away this should be good to keep merging through.

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -125,6 +123,18 @@ private void submitRequest(SingularityRequest request, Optional<SingularityReque
     requestHelper.updateRequest(request, oldRequest, requestState, historyType, JavaUtils.getUserEmail(user), skipHealthchecks, message);
   }
 
+  private void checkForAuthorization(SingularityRequest request, Optional<SingularityRequest> oldRequest, SingularityAuthorizationScope scope) {
```
darcatron: @tpetr this check is in `RequestResource` because passing along the `SingularityUser` to `SingularityValidator` would be confusing

tpetr: could we keep the code as it was and add something like a `SingularityValidator.checkForAuthorizedChanges(Optional<SingularityRequest> old, SingularityRequest new, SingularityUser user)` method that we call after the two `checkForAuthorization()` calls?

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -113,6 +113,7 @@ private void submitRequest(SingularityRequest request, Optional<SingularityReque
 
     if (oldRequest.isPresent()) {
       authorizationHelper.checkForAuthorization(oldRequest.get(), user, SingularityAuthorizationScope.WRITE);
+      authorizationHelper.checkForAuthorizedChanges(request, oldRequest.get(), user.get()); // previous check guarantees that user is present
```
darcatron: @tpetr I put it here so we weren't checking the same conditional again

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java
```diff
@@ -153,14 +156,21 @@ public void checkForAuthorization(SingularityRequest request, Optional<Singulari
...
+      checkForbidden(userIsRequestOwner || userIsJITA, "%s must be a member of one or more groups to %s %s: %s", user.get().getId(), scope.name(), request.getId(), JavaUtils.COMMA_JOINER.join(Iterables.concat(readWriteGroups, jitaGroups)));
     } else if (scope == SingularityAuthorizationScope.ADMIN) {
       checkForbidden(userIsAdmin, "%s must be a member of one or more groups to %s %s: %s", user.get().getId(), scope.name(), request.getId(), JavaUtils.COMMA_JOINER.join(adminGroups));
     }
   }
 
+  public void checkForAuthorizedChanges(SingularityRequest request, SingularityRequest oldRequest, SingularityUser user) {
+    if (oldRequest.getGroup().isPresent() && !oldRequest.getReadWriteGroups().equals(request.getReadWriteGroups())) {
+      checkUnauthorized(user.getGroups().contains(oldRequest.getGroup().get()),
+          "Only admins and owners (members of accessGroup as specified in the deploy configs) can add or remove groups from readWriteGroups");
```
tpetr: `accessGroup` and "deploy configs" are HubSpot-specific terms. I'd suggest re-wording to something like "Only admins and members of the Request's owner group can add or remove groups from readWriteGroups"

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
tpetr: comment isn't exactly right -- if auth is disabled it's not going to check if user is present or not (https://github.com/HubSpot/Singularity/blob/4b3e6f90860bf65734ed1ffef0dee7241093b425/SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizationHelper.java#L136)

darcatron: So in this situation, if no user is present, do we fail since they _technically_ are not in the request group?

darcatron:
@tpetr @ssalinas tested and working. 🚢 ?

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -54,30 +54,29 @@
...
   private final int defaultBounceExpirationMinutes;
+  private final int maxMemoryMbPerRequest;
   private final int maxMemoryMbPerInstance;
   private final boolean allowRequestsWithoutOwners;
   private final boolean createDeployIds;
   private final int deployIdLength;
   private final DeployHistoryHelper deployHistoryHelper;
   private final Resources defaultResources;
+  private final Pattern DAY_RANGE_REGEXP = Pattern.compile("[0-7]-[0-7]");
+  private final Pattern COMMA_DAYS_REGEXP = Pattern.compile("([0-7],)+([0-7])?");
```
tpetr: we should make these `private static final`, since they're constants that wont change between different instances of `SingularityValidator` (I know you could make the argument that since this is annotated with `@Singleton` it shouldn't matter, but it's still good code style to do this)

tpetr:
LGTM to merge into `hs_stable` once that last piece of feedback is addressed

SingularityService/src/test/java/com/hubspot/singularity/SingularityAuthorizationHelperTest.java
```diff
@@ -2,10 +2,10 @@
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
ssalinas: Last thing on this then we can merge @darcatron . We try to avoid any import * when possible. Can you split this out?

darcatron: donzo

PR #1227, opened 2016-08-18
Redeploy

### LINKED ISSUE #1079: Feature Request - Ability to re-deploy previous deployments

Calvinp:
Reopening this against master.

The modal looks like this:
Two things to notice:

Fixes #1079.
cc @tpetr @kwm4385 @wolfd


ssalinas:
Merge conflicts fix ed here and a few tweaks made, another good for testing soon

PR #1228, opened 2016-08-18
Use direct string matching with glob patterns instead of fuzzy matching

Calvinp:
Reopening this against master.

Gets rid of fuzzy matching on the requests page, using exact string matching instead (or glob matching if the user specifies a glob pattern).

TODO:
- [x] Implement glob matching on the tasks page too.
- [x] Make it clear to the user that glob matching is a possibility.
- [ ] Implement glob matching with highlighting on global search (Maybe).

Note that matches at the start of a requestId are prioritized over matches anywhere else (including at the start of the user field).

cc @tpetr @kwm4385 @wolfd


ssalinas:
Closing in favor of #1324

PR #1229, opened 2016-08-18
Starred requests persistence frontend

Calvinp:
Reopen against `starred_requests_persistence` (#1219)

Implements the frontend for #1219 
TODO:
- [x] Add the ability to import starred requests.
- [x] Implement UI changes to go along with API upgrades that separate out adding and removing starred requests from changing settings.
- [x] Right now the stars all blink when changing a star. Fix that.
- [x] Delete stars stored locally after importing them.
- [x] Fix slowness.
- [x] Fix a nasty bug that causes all old starred requests to be deleted from API in some cases in which a user switches browsers.
- [x] Implement fall-back to local stars when there's no auth or stars are broken.
- [ ] UI changes to go along with API upgrades that remove need to send user id param.

cc @tpetr @ssalinas @MattCCS


wolfd:
The best way of doing this would be to immediately update the store when the action goes through the reducer, and having the request error/finish action either revoke or leave the new star the same.

Because this doesn't really align with the previous api action stuff, we should put that into a separate reducer, we can still use the same action creators and even leave the `api` store the same, just look for the same actions in the other reducer.

Does that make sense?

Calvinp:
Ok, I've implemented this. How does it look now?

SingularityUI/app/actions/api/user.es6
```diff
@@ -0,0 +1,41 @@
+import { buildApiAction, buildJsonApiAction } from './base';
+
+export const FetchUser = buildApiAction(
+  'FETCH_USER',
+  {url: '/auth/user'}
+);
```
tpetr: Can we move this back to `auth.es6`? We're trying to match the actions folder structure with the API paths.

Calvinp: Done.

SingularityUI/app/actions/api/base.es6
```diff
@@ -23,12 +23,12 @@ export function buildApiAction(actionName, opts = {}, keyFunc = undefined) {
     return { type: CLEAR };
   }
 
-  function started(key = undefined) {
-    return { type: STARTED, key };
+  function started(key = undefined, options = undefined) {
```
tpetr: not exactly asking you to change anything, but won't these args default to `undefined` by default? is there a benefit to being explicit here?

SingularityUI/app/actions/api/users.es6
```diff
@@ -0,0 +1,36 @@
...
+    url: `/users/settings?userId=${userId}`,
+    catchStatusCodes: [404] // Remove this once the API doesn't 404 when user has nothing set
+  })
+);
+
+export const UpdateUserSettings = buildJsonApiAction(
+  'UPDATE_USER_SETTINGS',
+  'POST',
+  (userId, settings) => ({
+    url: `/users/settings?userId=${userId}`,
```
tpetr: this might be more of a question for @MattCCS, but why do these endpoints require the userId, but the other two dont? We should standardize on not requiring an explicit userId query param anywhere and let Singularity's authentication mechanism figure out who you are

SingularityUI/app/actions/api/base.es6
Calvinp: That is a good point. I think I was copying the style of the key argument, but when I think about it it really doesn't make much sense. I'll take that out.

SingularityUI/app/actions/api/users.es6
MattCCS: This was the decision @ssalinas advised me on, I believe it had to do with convention for GET requests, but I'm not sure.

ssalinas: @MattCCS and @Calvinp , talked with @tpetr and we will remove the use of the query param from here. Matt, we will rely just on the injected `Optional<SingularityUser>` to determine the userId. (i.e. this will work only when using a SingularityAuthenticator setup besides the disabled one). Calvin, in the case where we aren't able to get the user settings from the api, we can fall back to using localStorage.

Calvinp: I like that, it'll make the whole settings API simpler.
What status code will be returned if users aren't supported? I'll have to ignore that code.

ssalinas:
Moving all of these changes into #1219, closing this one

PR #1230, opened 2016-08-18
Disaster detection and disabled actions

ssalinas:
In a critical situation it is helpful to limit the amount of task churn in Singularity. This PR adds the ability for an admin to globally disable certain actions. So far it is implemented for `BOUNCE`, `DEPLOY`, `SCALE`, `REMOVE`, and `DECOMMISSION` but it's easy to add more if needed.

a `POST` to `/disasters/disabled-actions/{action}` adds an action to the list of ones that are disabled with an optional message
a `DELETE` to `/disasters/disabled-actions/{action}` removes it from that list

Singularity will respond with a 423 (locked) and the message given when disabling (or a default message)

In this PR I am also adding an automated way of disabling actions based on things such as task lag and the frequency of lost slaves or lost tasks.

TODO for this PR:
- [x] Add a test endpoint to explicitly trigger task reconciliation
- [x] Add the ability to globally disable actions of certain types
- [x] Admin ui page to add/remove disabled actions
- [x] Add disaster detection around things like number of overdue tasks, slaves lost, etc, that will run on the leader only
- [x] Allow the disaster detection to disable certain actions when one of its thresholds is crossed
- [x] Email admins when a disaster is detected
- [x] Tune metrics appropriately to decide when something bad is happening
- [x] Add an api endpoint to manually enter/exit disaster mode
- [x] Admin ui page to manually enter/exit disaster mode and show current stats
- [x] Docs

/cc @tpetr


ssalinas:
@tpetr @wolfd @Calvinp feel free to tear apart my react if I'm doing anything silly here. Modeled this page off of the webhooks one since it's fairly similar and pretty simple.

SingularityUI/app/components/common/modalButtons/NewDisabledActionModal.jsx
```diff
@@ -0,0 +1,62 @@
...
+        action="Create Disabled Action"
+        buttonStyle="success"
+        onConfirm={(data) => this.props.newDisabledAction(data.type, data.message)}
+        formElements={[
+          {
+            type: FormModal.INPUT_TYPES.SELECT,
+            name: 'type',
+            label: 'Type',
+            isRequired: true,
+            options: DISABLED_ACTION_TYPES.map((type) => ({
```
Calvinp: It would be nice to exclude already-disabled actions from this list.
(It would also be nice to be able to select multiple. But I don't think we have a multiselect component in FormModal yet, and I don't know if the backend would even allow it.)

SingularityUI/app/components/disabledActions/DisabledActions.jsx
```diff
@@ -0,0 +1,93 @@
...
+  return {
+    fetchDisabledActions: () => dispatch(FetchDisabledActions.trigger())
+  };
+}
+
+function refresh(props) {
+  return props.fetchDisabledActions();
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(rootComponent(DisabledActions, 'DisabledActions', refresh));
```
Calvinp: We probably want a space in the page title. (`'Disabled Actions'` instead of `'DisabledActions'`)

ssalinas: fixed

SingularityUI/app/components/common/modalButtons/NewDisabledActionModal.jsx
ssalinas: Yeah, I didn't write the backend to support multiple yet, might be a good idea at some point. However, it does support overriding the message for one that is already present (i.e if you issue a new request for disabling bounces when they are alreayd disabled, it will take the message form the newer one). So I might leave it be for the moment

ssalinas:
@tpetr Updated the PR a bit to take the time range over which events are occurring into account better. - task lag takes into account how long the calculated value has been over the specified threshold. A disaster will trigger if it has been over for a certain amount of time (45s default)
- For lost tasks/slaves, the poller will take all lost items from the last X seconds into account (30s by default) instead of just checking an instantaneous value. Now the interval on which the poller runs does not matter as much (though things get more accurate the more often it does)
- Poller now keeps more data points (10 by default)
- Eliminated the call to get all pending tasks for efficiency (just getting the IDs) so now the poller can run more often as well.

ssalinas:
@tpetr added a more comprehensive `SingularityAction` enum, each action in the list has a property to determine if it can be disabled or not (i.e. if I actually put the validator code in place to make it disableable)

SingularityBase/src/main/java/com/hubspot/singularity/SingularityAction.java
```diff
@@ -0,0 +1,22 @@
...
+  BOUNCE_REQUEST(true), SCALE_REQUEST(true), REMOVE_REQUEST(true), CREATE_REQUEST(true), UPDATE_REQUEST(true), VIEW_REQUEST(false), PAUSE_REQUEST(false),
+  KILL_TASK(true), BOUNCE_TASK(true), RUN_SHELL_COMMAND(true), ADD_METADATA(true),
+  DEPLOY(true), CANCEL_DEPLOY(true),
+  ADD_WEBHOOK(true), REMOVE_WEBHOOK(true), VIEW_WEBHOOKS(false),
+  TASK_RECONCILIATION(true),
+  ADD_DISASTER(false), REMOVE_DISASTER(false), DISABLE_ACTION(false), ENABLE_ACTION(false), VIEW_DISASTERS(false),
+  FREEZE_SLAVE(true), ACTIVATE_SLAVE(true), DECOMMISSION_SLAVE(true), VIEW_SLAVES(false),
+  FREEZE_RACK(true), ACTIVATE_RACK(true), DECOMMISSION_RACK(true), VIEW_RACKS(false);
+
+  private final boolean disableable;
```
tpetr: can we word this better? `canDisable` maybe?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDisasterDetectionPoller.java
```diff
@@ -0,0 +1,262 @@
...
+    long totalTaskLagMillis = 0;
+    int numPastDueTasks = 0;
+
+    for (SingularityPendingTaskId pendingTask : pendingTasks) {
+      long taskLagMillis = now - pendingTask.getNextRunAt();
+      if (taskLagMillis > 0) {
+        numPastDueTasks++;
+        totalTaskLagMillis += taskLagMillis;
+        if (taskLagMillis > configuration.getDeltaAfterWhichTasksAreLateMillis()) {
+          numLateTasks ++;
```
tpetr: nit: spacing

SingularityService/src/main/java/com/hubspot/singularity/data/transcoders/SingularityTranscoderModule.java
```diff
@@ -2,6 +2,8 @@
 
 import static com.hubspot.singularity.data.transcoders.SingularityJsonTranscoderBinder.bindTranscoder;
 
+import javax.ws.rs.HEAD;
```
tpetr: this snuck in from merge conflicts

PR #1235, opened 2016-08-19
Properly handle deleted requests

Calvinp:
Instead of showing the not found page when a user navigates to the detail page of a deleted request, show the requestId, an alert saying that the request has been deleted, and the request history.

screenshot 2016-08-19 15 47 34

cc @tpetr @ssalinas @wolfd


tpetr:
LGTM

PR #1236, opened 2016-08-19
UI Updates for default config

ssalinas:
- Updates the start script so local dev will work with newer and older docker setups by default. Without this, was constantly getting the 404 page when trying to navigate directly to a page
- Fix some other 404s when features like s3 logs are not in use

@tpetr @Calvinp


tpetr:
LGTM

PR #1239, opened 2016-08-22
Add href to navitems

Calvinp:
Changes global nav:
- No longer use react-bootstrap nav items, since they were causing weird tab issues.
- Add an href (or Link `to` prop) to all the links to allow opening the correct link in a new tab.

cc @tpetr @wolfd


tpetr:
LGTM

PR #1242, opened 2016-08-22
Sort environment variables alphabetically

Calvinp:
Exactly what it says on the tin - sorts environment variables alphabetically on the task detail page.

cc @wolfd @tpetr


tpetr:
LGTM

PR #1243, opened 2016-08-22
Automatically refresh some tables

Calvinp:
Automatically refreshes the historical tasks table on the request detail page and deploy detail page (shortly).

This allows the table to detect new finishing tasks within a few seconds of when they finish.

cc @tpetr @wolfd


SingularityUI/app/components/deployDetail/DeployDetail.jsx
```diff
@@ -162,6 +162,8 @@ class DeployDetail extends React.Component {
           keyGetter={(task) => task.taskId.id}
           rowChunkSize={5}
           paginated={true}
+          shouldRefresh={true}
+          refreshInterval={10000}
```
Calvinp: I'd like to call specific attention to this (and same on the TaskHistoryTable below) as something to discuss.
Since the table can't detect page blur and focus events, sometimes the page will refresh but the table won't. At 60 seconds (the global config), this change barely had any effect - I would get frustrated/scared waiting that long and refresh the page manually.
But at the same time we make an API call every time this interval comes up. So I didn't want to make the number too small either.
I settled on 10 seconds as a good in between, but that was fairly arbitrary and I'm very open to changing it.

ssalinas:
Closing in favor of #1325

PR #1245, opened 2016-08-23
Update scripts README - request id supports pattern

marcob:
Add to the README that `-r | --request-id` supports https://docs.python.org/2/library/fnmatch.html matching.

@ssalinas


ssalinas:
👍 thanks @marcob

PR #1246, opened 2016-08-23
Make Duration Fields String Inputs Again

Calvinp:
Duration fields used to be string inputs that were validated and parsed by the `juration` library.
Then we changed them, but that change introduced a 24 hour maximum duration restriction.
This reverts duration fields to strings parsed by `juration` and removes that maximum.

cc @tpetr @wolfd


tpetr:
LGTM

PR #1247, opened 2016-08-23
(WIP) Disaster detection

ssalinas:
- [x] Add a test endpoint to explicitly trigger task reconciliation
- [x] Add disaster detection around things like number of overdue tasks, slaves lost, etc, that will run on the leader only
- [x] Allow the disaster detection to disable certain actions when one of its thresholds is crossed
- [x] Email admins when a disaster is detected
- [ ] Tune metrics appropriately to decide when something bad is happening
- [x] Add an api endpoint to manually enter/exit disaster mode
- [ ] Admin ui page to manually enter/exit disaster mode (Maybe?)


SingularityService/src/main/java/com/hubspot/singularity/data/DisasterManager.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public void updateActiveDisasters(List<SingularityDisasterType> previouslyActiveDisasters, List<SingularityDisasterType> newActiveDisasters) {
+    for (SingularityDisasterType disaster : previouslyActiveDisasters) {
+      if (!newActiveDisasters.contains(disaster)) {
+        removeDisaster(disaster);
+      }
+    }
+
+    for (SingularityDisasterType disaster : newActiveDisasters) {
+      addDisaster(disaster);
```
darcatron: Doesn't this again add previously added disasters?

ssalinas: Ah, yeah I can make it skip those. Originally I was going to store a message in there, so it made sense to call that again in case the message had been updated. Can probably do a `if (!exists(path))` check first, thanks!

ssalinas: updated 👍

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -199,13 +204,22 @@ public void slaveLost(SlaveID slaveIdObj) {
...
+      }
 
       checkRackAfterSlaveLoss(slave.get());
     } else {
       LOG.warn("Lost a slave {}, but didn't know about it", slaveId);
     }
   }
 
+  private void updateDisasterCounter(SingularitySlave slave) {
+    if (slave.getCurrentState().getState() == MachineState.ACTIVE) {
```
darcatron: I might not understand this part fully, but if the slave state is changed to `MachineState.DEAD`, will this ever be true?

ssalinas: Ah, yeah the variable could probably use renaming. The one I'm grabbing is actually the previous slave state, not the one we are about to update to. Essentially I'm looking for any transitions from `ACTIVE` -> `DEAD`.

ssalinas:
Going to merge this into disabled actions since they go together and are now both on staging + qa

PR #1253, opened 2016-08-26
Tailer

wolfd:


SingularityUI/SingularityTailer/src/components/Log.js
```diff
@@ -0,0 +1,296 @@
...
+  fileSize: PropTypes.number,
+  lines: PropTypes.instanceOf(Immutable.List),
+  chunks: PropTypes.instanceOf(Immutable.List),
+  requests: PropTypes.instanceOf(Immutable.Map),
+  config: PropTypes.object.isRequired,
+  lineLinkRenderer: PropTypes.func
+};
+
+Log.defaultProps = {
+  minTailDelayMs: 100,
```
ssalinas: wondering if `100`ms is too low here. To a user, there probably wouldn't be much a perceptible difference between `100` and `500`. Concern here is the load on the api when hitting it every 100ms

ssalinas:
Time to merge this guy 🎉 
Any other updates/fixes to the tailer will be in future PRs

PR #1254, opened 2016-08-29
Ignore empty logs in logfetch

MattCCS:
This PR modifies logfetch to ignore logfiles whose size is 0.  This should prevent logfetch from downloading logs with no content.

/cc @ssalinas @tpetr


ssalinas:
`hasData` shouldn't apply for tailing. It's possible that we are starting to tail a log file that hasn't been written to yet. Other than that looks good

ssalinas:
Looks good @MattCCS

PR #1257, opened 2016-08-29
Handle missing logfetch tail data

MattCCS:
This PR handles the case in which a logfetch tail response has no 'data' field.


ssalinas:
LGTM

PR #1258, opened 2016-08-30
Fix logfetch merge error

MattCCS:
This PR fixes a previous bad merge by removing merge cruft and correcting the import.  It also removes the outdated method which had only been referenced in the import.


ssalinas:
Thanks @MattCCS

PR #1259, opened 2016-08-31
Alternate compression formats and viewing compressed files in UI

ssalinas:
This adds support for other compression formats besides gzip to be used for logrotate. The main target of this PR is support for bzip2, but currently `LogrotateCompressionSettings` is not enforcing any particular extensions, so we can extend to other formats later if needed. I had started adding supported log formats as an enum, but realized I would need the additional fields for executable locations anyways and might as well leave the config more open. 

Changes will be needed in logfetch to handle bz2 files as well and will come in a separate PR.

_note_ : Also added a few more to the not tail-able files list in the ui as part of this too

/cc @tpetr @MattCCS


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/LogrotateCompressionSettings.java
```diff
@@ -0,0 +1,70 @@
...
+                                      @JsonProperty("uncompressCmd") Optional<String> uncompressCmd,
+                                      @JsonProperty("compressOptions") Optional<String> compressOptions,
+                                      @JsonProperty("compressExt") Optional<String> compressExt) {
+    this.compressCmd = compressCmd;
+    this.uncompressCmd = uncompressCmd;
+    this.compressOptions = compressOptions;
+    this.compressExt = compressExt;
+  }
+
+  public static LogrotateCompressionSettings gzip() {
```
tpetr: i'd consider renaming this to `default()`, since it's just an implementation detail of logrotate that it's using `gzip` to compress

ssalinas: default is a reserved word/identifier in java, you can't name a method that. Renamed to `empty()` instead

PR #1260, opened 2016-08-31
logfetch support for bzip2

ssalinas:
Goes with #1259


ssalinas:
Not needed at the moment, going to close for now

PR #1262, opened 2016-09-01
Support for setting user in default executor

### LINKED ISSUE #1261: select user to launch tasks as
Is it possible to select the user used to execute a task as? I am thinking something similar to supervisor's user option in the processes definition which ultimately uses os.setgroups/os.setgid/os.setuid. Thanks.

ssalinas:
/fixes #1261


kchaliki:
@ssalinas thanks for the addition

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -381,6 +381,10 @@ private void prepareCustomExecutor(final TaskInfo.Builder bldr, final Singularit
   private void prepareCommand(final TaskInfo.Builder bldr, final SingularityTaskId taskId, final SingularityTaskRequest task, final Protos.Offer offer, final Optional<long[]> ports) {
     CommandInfo.Builder commandBldr = CommandInfo.newBuilder();
 
+    if (task.getDeploy().getUser().isPresent()) {
+      commandBldr.setUser(task.getDeploy().getUser().get());
```
tpetr: Can we add this corresponding change to `prepareCustomExecutor()` too? I can see some utility in being able to control what user the custom executor is launched as.

tpetr:
LGTM aside from comment above

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
ssalinas: We already had `customExecutorUser` for this purpose, but it had been deprecated for a while now. I've officially removed it in favor of the plain `user` field which will apply to the custom and default executor cases.

PR #1264, opened 2016-09-01
Execution timeout for tasks

darcatron:
Kills tasks after the specified timeout has been reached/exceeded


SingularityBase/src/main/java/com/hubspot/singularity/TaskCleanupType.java
```diff
@@ -3,8 +3,8 @@
 public enum TaskCleanupType {
 
   USER_REQUESTED(true, true), USER_REQUESTED_TASK_BOUNCE(false, false), DECOMISSIONING(false, false), SCALING_DOWN(true, false), BOUNCING(false, false), INCREMENTAL_BOUNCE(false, false),
-  DEPLOY_FAILED(true, true), NEW_DEPLOY_SUCCEEDED(true, false), DEPLOY_STEP_FINISHED(true, false), DEPLOY_CANCELED(true, true), UNHEALTHY_NEW_TASK(true, true),
-  OVERDUE_NEW_TASK(true, true), USER_REQUESTED_DESTROY(true, true), INCREMENTAL_DEPLOY_FAILED(false, true), INCREMENTAL_DEPLOY_CANCELLED(false, true), PRIORITY_KILL(true, true);
```
tpetr: looks like some sort of merge issue here -- we shouldn't remove `PRIORITY_KILL` from this enum

SingularityBase/src/main/java/com/hubspot/singularity/TaskCleanupType.java
```diff
@@ -3,8 +3,8 @@
...
+  DEPLOY_FAILED(true, true), NEW_DEPLOY_SUCCEEDED(true, false), DEPLOY_STEP_FINISHED(true, false), DEPLOY_CANCELED(true, true), DEPLOY_TIMED_OUT(true, true), UNHEALTHY_NEW_TASK(true, true),
```
tpetr: I'm not sure if `DEPLOY_TIMED_OUT` is the right name for this. The "DEPLOY" has the connotation that a deployment took too long (i.e. it took too long for a task to become healthy). `TASK_EXCEEDED_TIME_LIMIT` might be a better option (open to other ideas too)

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -214,6 +214,11 @@
 
   private long warnIfScheduledJobIsRunningForAtLeastMillis = TimeUnit.DAYS.toMillis(1);
 
+  @JsonProperty("taskExecutionTimeout")
+  @Valid
+  @NotNull
+  private long taskExecutionTimeout = TimeUnit.HOURS.toMillis(12);
```
tpetr: we should make this an `Optional<Long>` and default it to `absent()`

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityJobPoller.java
```diff
@@ -69,6 +72,16 @@ public void runActionOnPoll() {
         continue;
       }
 
+      if (start - taskId.getStartedAt() >= configuration.getTaskExecutionTimeout()) {
```
tpetr: we also need to be checking the Request type. As-is, this would kill a web service after running for 12 hours, which isn't what we want -- we should only be checking this for non-long running Requests (i.e. cron jobs, on-demand and run once tasks)

SingularityBase/src/main/java/com/hubspot/singularity/TaskCleanupType.java
darcatron: yeah, I thought something funky happened during that conflict. Will update :arrow_right_hook:

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityJobPoller.java
```diff
@@ -69,6 +72,16 @@ public void runActionOnPoll() {
...
+        taskManager.createTaskCleanup(new SingularityTaskCleanup(
+            Optional.<String>absent(),
+            TaskCleanupType.DEPLOY_TIMED_OUT,
+            start,
+            taskId,
+            Optional.of("Deploy has reached/exceeded the set timeout"),
```
tpetr: Again, the "deploy" wording isn't quite right here. Also, consider updating the message to include the actual timeout value (i.e. something like `Task has exceeded maximum execution time of XX:XX`)

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
ssalinas: As an addition to that, we should probably have a way to override this at the request level (in `SingularityRequest` object) so that an individual request can enforce a shorter/longer execution time limit if it needs to

tpetr: ^ yes

darcatron: @ssalinas that would be another config field, right?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityJobPoller.java
darcatron: So the `SCHEDULED` tasks should be shut down as well. Right now, the method just warns that it's overdue. Should that part be removed since it's redundant or should `SCHEDULED` request types just continue to have overdue alerts while `ON_DEMAND` and `RUN_ONCE` tasks are cleaned up?

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
ssalinas: Yeah, another field on the `SingularityRequest` object

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityJobPoller.java
ssalinas: `getWarnIfScheduledJobIsRunningForAtLeastMillis` can be different than our executionTimeout, so I would think that both would still apply. Warn if it's over the first one, kill if it's over the executionTimeout

darcatron: Follow up to this. Should this conditional have the work of ignoring long running request types or should we have the validator make sure that the config field `taskExecutionTimeLimitMillis` is only set for non-long running tasks? [likely here](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java#L202-L205)? [maybe even both](http://porquenolosdos.com/)?

ssalinas: I would say both, just to be safe

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityJobPoller.java
```diff
@@ -78,11 +82,31 @@ public void runActionOnPoll() {
...
-      if (request == null || !request.getRequest().isScheduled() || taskManager.hasNotifiedOverdue(taskId)) {
+      if (request == null) {
         continue;
       }
 
       final long runtime = start - taskId.getStartedAt();
+
+      if (!request.getRequest().getRequestType().isLongRunning() &&
+          configuration.getTaskExecutionTimeLimitMillis().isPresent() &&
+          runtime >= configuration.getTaskExecutionTimeLimitMillis().get()) {
```
ssalinas: Probably want to check against `request. getTaskExecutionTimeLimitMillis().or(configuration.getTaskExecutionTimeLimitMillis())` here so that the individual request can override the execution time limit

darcatron:
@tpetr @ssalinas okay to merge and verify this in TEST (tomorrow Wed. 14th)?

darcatron:
Accidentally pushed into hs_qa branch, but didn't deploy it yet. I left the hs_qa label because it's on the branch

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityJobPoller.java
```diff
@@ -76,13 +81,34 @@ public void runActionOnPoll() {
...
+          runtime >= request.getTaskExecutionTimeLimitMillis().or(configuration.getTaskExecutionTimeLimitMillis().get())) {
+
+        taskManager.createTaskCleanup(new SingularityTaskCleanup(
+            Optional.<String>absent(),
+            TaskCleanupType.TASK_EXCEEDED_TIME_LIMIT,
+            start,
+            taskId,
+            Optional.of(String.format("Task has run for %s, which exceeds the maximum execution time of %s",
+                DurationFormatUtils.formatDurationHMS(runtime),
+                DurationFormatUtils.formatDurationHMS(configuration.getTaskExecutionTimeLimitMillis().get()))),
```
ssalinas: Just an informational statement, but will want the request.get...().or(config.... here as well for accuracy

ssalinas:
One extra thought on this. We can do it in a separate PR, but we will eventually need to add the extra request field [to the ui form](https://github.com/HubSpot/Singularity/tree/master/SingularityUI/app/components/requestForm) , otherwise the field gets removed when you edit a request in the ui since the field isn't on the json

ssalinas:
👍 LGTM

PR #1265, opened 2016-09-02
Update slave-extras.md

kchaliki:
I was following this and found that without the quote I get a YAML parsing error in Jackson - this makes it go away but to be honest I have not confirmed it actually works.

14:06:02.743 [main] ERROR c.h.s.e.SingularityExecutorRunner - Finished after 00:01.150 with error
com.google.inject.CreationException: Unable to create injector, see the following errors:

1) Error in custom provider, while scanning for the next token
found character % '%' that cannot start any token. (Do not use % for indentation)
 in 'reader', line 14, column 23:
    s3UploaderKeyPattern: %requestId/%Y/%m/%taskId_%index- ... 
                          ^

### DIFF
--- a/Docs/reference/slave-extras.md
+++ b/Docs/reference/slave-extras.md
@@ -65,7 +65,7 @@ logrotateToDirectory: logs
 s3UploaderBucket: bucket-name
 
 # Filename format to use when uploading logs to S3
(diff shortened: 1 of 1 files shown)


tpetr:
You're right, that needs to be wrapped in quotes. Thanks!

PR #1266, opened 2016-09-02
select the correct healthcheck for alert

ssalinas:
Use the last healthcheck in the list, not the first

/cc @tpetr

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskDetail.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskDetail.jsx
@@ -406,7 +406,7 @@ function mapHealthchecksToProps(task) {
   if (!task) return task;
   const { healthcheckResults } = task;
   task.hasSuccessfulHealthcheck = healthcheckResults && healthcheckResults.length > 0 && !!_.find(healthcheckResults, (healthcheckResult) => healthcheckResult.statusCode === 200);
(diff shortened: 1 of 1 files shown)


SingularityUI/app/components/taskDetail/TaskDetail.jsx
```diff
@@ -406,7 +406,7 @@ function mapHealthchecksToProps(task) {
   if (!task) return task;
   const { healthcheckResults } = task;
   task.hasSuccessfulHealthcheck = healthcheckResults && healthcheckResults.length > 0 && !!_.find(healthcheckResults, (healthcheckResult) => healthcheckResult.statusCode === 200);
-  task.lastHealthcheckFailed = healthcheckResults && healthcheckResults.length > 0 && healthcheckResults[0].statusCode !== 200;
+  task.lastHealthcheckFailed = healthcheckResults && healthcheckResults.length > 0 && _.last(healthcheckResults).statusCode !== 200;
```
tpetr: do you know if this was something that was always broken, or did the sorting change at some point?

ssalinas: I think it got messed up in the translation to react. It used to work. You can see we're actually using last already in the TaskAlert to display it, just not here when determining if it failed

PR #1268, opened 2016-09-06
Process status updates in separate thread

tpetr:
This PR adds the option to process Mesos status updates on a thread other than the main Mesos SchedulerDriver thread. This won't speed up status update processing time (they still need to be processed sequentially), but will allow for other events (like resource offers) to contend for control of the global scheduler lock. This should help remedy situations where resource offers take a long time to be handled due to a large backlog of status updates (like during outages or reconciliations, for example).

**Note:** This is disabled by default while we're testing it all out. Enable by setting `processStatusUpdatesInSeparateThread: true` in your config YAML.


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        return maybeSchedulerDriver.get();
+    }
+
+    @Timed
+    public void processStatusUpdate(Protos.TaskStatus status) {
+        schedulerLock.lock();
+        try {
+            final String taskId = status.getTaskId().getValue();
```
wsorenson: sub method will make this try { } finally { readable

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -0,0 +1,299 @@
...
+                return;
+            }
+
+            updateFuture = executorService.submit(new Runnable() {
+                @Override
+                public void run() {
+                    LOG.info("Status update handler thread started");
+                    while (!Thread.currentThread().isInterrupted()) {
+                        try {
+                            final Protos.TaskStatus status = statusUpdateQueue.take();
```
wsorenson: I'm assuming take() blocks on wait?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -0,0 +1,299 @@
...
+        this.logSupport = logSupport;
+        this.scheduler = scheduler;
+        this.stateCacheProvider = stateCacheProvider;
+        this.serverId = serverId;
+        this.schedulerDriverSupplier = schedulerDriverSupplier;
+        this.schedulerLock = schedulerLock;
+        this.handlerStarted = new AtomicBoolean();
+
+        this.statusUpdateQueue = new ArrayBlockingQueue<>(configuration.getStatusUpdateQueueCapacity());
+        this.executorService = Executors.newFixedThreadPool(1);
```
wsorenson: This should be a daemon thread factory, esp. if we aren't calling shutdown on it in stop() - shouldn't we be?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -0,0 +1,299 @@
...
+
+            updateFuture = executorService.submit(new Runnable() {
+                @Override
+                public void run() {
+                    LOG.info("Status update handler thread started");
+                    while (!Thread.currentThread().isInterrupted()) {
+                        try {
+                            final Protos.TaskStatus status = statusUpdateQueue.take();
+                            LOG.info("Handling status update for {} {}", status.getTaskId().getValue(), status.getState());
+                            processStatusUpdate(status);
```
wsorenson: How should this thread handle errors?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -0,0 +1,299 @@
...
+            } catch (InterruptedException ie) {
+                // TODO: what's the best thing to do here?
+            }
+        } else {
+            processStatusUpdate(status);
+        }
+    }
+
+    public void start() {
+        if (processStatusUpdatesInSeparateThread) {
```
wsorenson: If you invert this if it will reduce indentation

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
jhaber: What about using [this](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedScheduledExecutorServiceFactory.java)

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
wsorenson: It probably needs to exit just like it would have on the scheduler thread.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
tpetr: Yep

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
tpetr: Right, will add an explicit abort if the thread encounters an error.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
tpetr: This was just a quick and dirty first stab -- leveraging `SingularityManagedScheduledExecutorServiceFactory` as Haber mentioned probably makes sense.

ssalinas:
@tpetr think this is good to merge if we take care of merge conflicts?

tpetr:
we should bump up the default queue size to 1k, otherwise LGTM

Issue #1214, opened 2016-08-12
runner.sh does not always handle 'taskAppDirectory' properly


ssalinas:
Example, if `taskAppDirectory` is set to `.` , the runner script will still write to `../{logfile}`. This then mismatches with the logrotate.conf which assumes the file is in the `taskAppDirectory`, not 1 directory above

/cc @eherot

ssalinas:
Fixed in 0.11.0 release (#1273)

PR #1273, opened 2016-09-12
resolve the correct log path based on taskAppDirectory

ssalinas:
Attempt at fixing #1214 , still needs a bit more tweaking to work for all cases + some testing

/fyi @eherot

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
@@ -115,6 +115,7 @@ public class SingularityExecutorTaskProcessBuilder implements Callable<ProcessBu
       configuration.getLogrotateToDirectory(),
       executorData.getUser().or(configuration.getDefaultRunAsUser()),
       configuration.getServiceLog(),
+      serviceLogOutPath(),
       task.getTaskId(),
       executorData.getMaxTaskThreads().or(configuration.getMaxTaskThreads()),
       !getExecutorUser().equals(executorData.getUser().or(configuration.getDefaultRunAsUser())),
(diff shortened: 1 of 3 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
```diff
@@ -10,6 +11,7 @@
 import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 import com.hubspot.deploy.ExecutorData;
+import com.hubspot.mesos.MesosUtils;
```
tpetr: unused import?

ssalinas: fixed

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessBuilder.java
```diff
@@ -148,6 +151,16 @@ private ProcessBuilder buildProcessBuilder(TaskInfo taskInfo, ExecutorData execu
     return processBuilder;
   }
 
+  private String serviceLogOutPath() {
+    if (configuration.getTaskAppDirectory().equals(".")) {
+      return configuration.getServiceLog();
+    }
+    Path basePath = task.getTaskDefinition().getTaskDirectoryPath();
+    String appPath = configuration.getTaskAppDirectory().startsWith("./") ? configuration.getTaskAppDirectory().replace("./", "") : configuration.getTaskAppDirectory();
+    Path app = basePath.resolve(appPath);
```
tpetr: you should be able to collapse these two lines to just `basePath.resolve(configuration.getTaskAppDirectory()).normalize()`

PR #1274, opened 2016-09-13
restore default of root directory

### LINKED ISSUE #1270: Task page no longer shows root files before you click in
When you open a task page, it looks like this:
screen shot 2016-09-08 at 2 10 44 pm

You then have to click on the `root` link to actually get it to display the files:
screen shot 2016-09-08 at 2 10 53 pm

I think this is a regression, it used to show them by default.  Looks like it's focusing on some non-existent subdirectory of the sandbox that has the task id?
cc @pennello

ssalinas:
This used to be the default behavior, and I thought we had a config property for it, but I didn't see it anywhere. `defaultToTaskIdDirectory` (defaults to `false`) can be set to `true` to use the taskId directory as the default for those using the SingularityExecutor, otherwise it will use the root of the sandbox as the default directory.

/fixes #1270 
/fyi @stevenschlansker @tpetr

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java
@@ -92,6 +92,9 @@ public class UIConfiguration {
   @NotNull
   private Optional<String> extraScript = Optional.absent();
 
+  @JsonProperty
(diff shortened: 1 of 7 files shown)


ssalinas:
@tpetr , updated to take what the backend gives for currentDirectory with an undefined path as the default

stevenschlansker:
Thanks!

PR #1277, opened 2016-09-13
added threshold for startup tasks during singularity launch

darcatron:
@tpetr @ssalinas @MattCCS

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
@@ -141,10 +141,10 @@ class SingularityStartup {
         case DEPLOYING_TO_UNPAUSE:
           if(checkActiveRequest(requestWithState, deployKeyToPendingTaskId, now)) {
             numPendingStartupRequests++;
-            float startUpTaskPct = (numPendingStartupRequests * 100.0f / totalStartupRequests);
+            double startUpTaskPct = (numPendingStartupRequests * 100.0 / totalStartupRequests);
 
             if (startUpTaskPct > startUpTaskThresholdPct) {
-              LOG.info("%s% exceeded startup task threshold of %s%", startUpTaskPct, startUpTaskThresholdPct);
+              LOG.info("%s%% exceeded startup task threshold of %s%%", startUpTaskPct, startUpTaskThresholdPct);
               abort.abort(AbortReason.EXCEEDED_STARTUP_TASK_THRESHOLD, Optional.<Throwable>absent());
             }
           }
@@ -157,7 +157,8 @@ class SingularityStartup {
     }
   }
 
-  private boolean checkActiveRequest(SingularityRequestWithState requestWithState, Map<SingularityDeployKey, SingularityPendingTaskId> deployKeyToPendingTaskId, final long timestamp) {
+  @VisibleForTesting
+  boolean checkActiveRequest(SingularityRequestWithState requestWithState, Map<SingularityDeployKey, SingularityPendingTaskId> deployKeyToPendingTaskId, final long timestamp) {
     final SingularityRequest request = requestWithState.getRequest();
 
     if (request.getRequestType() == RequestType.ON_DEMAND || request.getRequestType() == RequestType.RUN_ONCE) {
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
@@ -42,7 +42,7 @@ import com.hubspot.singularity.scheduler.SingularityTaskReconciliation;
 
 @Singleton
 class SingularityStartup {
-
+  private static final int THRESHOLD = 128; // todo: put in SingularityConfiguration?
   private static final Logger LOG = LoggerFactory.getLogger(SingularityStartup.class);
 
   private final MesosClient mesosClient;
@@ -115,6 +115,7 @@ class SingularityStartup {
   @VisibleForTesting
   void checkSchedulerForInconsistentState() {
     final long now = System.currentTimeMillis();
+    int numPendingStartupRequests = 0;
 
     final Map<SingularityDeployKey, SingularityPendingTaskId> deployKeyToPendingTaskId = getDeployKeyToPendingTaskId();
 
@@ -123,7 +124,13 @@ class SingularityStartup {
         case ACTIVE:
         case SYSTEM_COOLDOWN:
         case DEPLOYING_TO_UNPAUSE:
-          checkActiveRequest(requestWithState, deployKeyToPendingTaskId, now);
+          if(checkActiveRequest(requestWithState, deployKeyToPendingTaskId, now)) {
+            numPendingStartupRequests++;
+            if (numPendingStartupRequests > THRESHOLD) {
+              LOG.info("Reached startup task threshold of %s", THRESHOLD);
+              return; // todo: abort Singularity??
+            }
+          }
           break;
         case DELETED:
         case PAUSED:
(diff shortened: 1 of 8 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
```diff
@@ -42,7 +42,7 @@
 
 @Singleton
 class SingularityStartup {
-
+  private static final int THRESHOLD = 128; // todo: put in SingularityConfiguration?
```
darcatron: Should this be moved to a field in SingularityConfiguration or elsewhere?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
```diff
@@ -123,7 +124,13 @@ void checkSchedulerForInconsistentState() {
         case ACTIVE:
         case SYSTEM_COOLDOWN:
         case DEPLOYING_TO_UNPAUSE:
-          checkActiveRequest(requestWithState, deployKeyToPendingTaskId, now);
+          if(checkActiveRequest(requestWithState, deployKeyToPendingTaskId, now)) {
+            numPendingStartupRequests++;
+            if (numPendingStartupRequests > THRESHOLD) {
+              LOG.info("Reached startup task threshold of %s", THRESHOLD);
+              return; // todo: abort Singularity??
```
darcatron: I wasn't sure how we want to go about failing.

ssalinas: To abort we'll use the `SingularityAbort` similar to the usage [here](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerDelegator.java#L98). You can add a new `AbortReason` to the enum

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
ssalinas: `SingularityConfiguration` is a good spot for that

tpetr: +1, let's also turn this into a % of active requests

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityStartupTest.java
```diff
@@ -159,4 +167,47 @@ public void testRunOnceDoesntGetRescheduled() {
     // assert that SingularityStartup does not enqueue a SingularityPendingRequest (pendingType=STARTUP) for the RUN_ONCE request
     Assert.assertTrue(requestManager.getPendingRequests().isEmpty());
   }
+
+  @Test
+  public void itFailsIfNumberOfStartupTasksPassThreshold() {
+    int maxPct = 100;
+    Assert.assertTrue("StartUpTaskThresholdPct cannot be above 100%", configuration.getStartUpTaskThresholdPct() <= maxPct);
+
+    // round down to closest multiple of 5 to reduce the number of requests created during this test
```
ssalinas: Why don't you just call `configuration.setStartUpTaskThresholdPct()` and set it to something low at the beginning of this test, then set it back to default value at the end. Should save a bunch of extra math, and allow you to set a lower threshold so you don't need to create a ton of requests for the test

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
```diff
@@ -115,6 +129,8 @@ public void startup(MasterInfo masterInfo, SchedulerDriver driver) throws Except
   @VisibleForTesting
   void checkSchedulerForInconsistentState() {
     final long now = System.currentTimeMillis();
+    int numPendingStartupRequests = 0;
+    int totalStartupRequests = requestManager.getNumRequests();
```
ssalinas: nit on naming, these are startup requests, this is just 'all' requests. Startup is specific to the type of pending request that is created in `checkActiveRequest`

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
```diff
@@ -123,7 +139,15 @@ void checkSchedulerForInconsistentState() {
         case ACTIVE:
         case SYSTEM_COOLDOWN:
         case DEPLOYING_TO_UNPAUSE:
-          checkActiveRequest(requestWithState, deployKeyToPendingTaskId, now);
+          if(checkActiveRequest(requestWithState, deployKeyToPendingTaskId, now)) {
+            numPendingStartupRequests++;
+            double startUpTaskPct = (numPendingStartupRequests * 100.0 / totalStartupRequests);
```
ssalinas: another nit on naming, not working with any tasks here, just requests and pending requests

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
```diff
@@ -133,18 +157,19 @@ void checkSchedulerForInconsistentState() {
     }
   }
 
-  private void checkActiveRequest(SingularityRequestWithState requestWithState, Map<SingularityDeployKey, SingularityPendingTaskId> deployKeyToPendingTaskId, final long timestamp) {
+  @VisibleForTesting
```
ssalinas: you aren't calling this in the test, change back to `private`?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
ssalinas: edit, ah I see the difficulty in trying to come up with a descriptive name for this now that I try and think of one, just don't want it to get confused with actual running tasks at all since we are counting enqueued pending requests

darcatron:
@ssalinas can you give it a look through once you get a chance

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -744,4 +761,19 @@ private int numInstancesExpected(SingularityRequest request, SingularityPendingR
     return Optional.of(nextRunAt);
   }
 
+  private int checkStartupThreshold(SingularityPendingRequest pendingRequest, int totalTasks) {
+    if (pendingRequest.getPendingType() == PendingType.STARTUP) {
+      int numNewStartupTasks = taskManager.getNumActiveTasksForRequest(pendingRequest.getRequestId());
```
ssalinas: this isn't quite what you want. `getActiveTasksForRequest` will get you _all_ active tasks for that request, so we might be including ones that were already running here. We already have the `numScheduledTasks` value a few lines above where this is called in `drainPendingQueue`. That value is just the number of tasks that were started due to this particular request and is what you want to be counting

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityStartup.java
```diff
@@ -81,7 +81,7 @@ public void startup(MasterInfo masterInfo, SchedulerDriver driver) throws Except
 
     MesosMasterStateObject state = mesosClient.getMasterState(uri);
 
-    slaveAndRackManager.loadSlavesAndRacksFromMaster(state);
+    slaveAndRackManager.loadSlavesAndRacksFromMaster(state, true);
```
ssalinas: this seems like it's from another branch. Might want to rebase off master???

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -90,9 +92,20 @@
 
   private final SingularityMailer mailer;
 
+  private final SingularityAbort abort;
+  private final AtomicInteger numStartupRequests = new AtomicInteger(0);
```
ssalinas: This is referring to tasks this time around, not requests (now that I've got my head on straight and am not leading you astray in this PR... 😉  )

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -744,4 +761,19 @@ private int numInstancesExpected(SingularityRequest request, SingularityPendingR
...
 
+  private int checkStartupThreshold(SingularityPendingRequest pendingRequest, int totalTasks) {
+    if (pendingRequest.getPendingType() == PendingType.STARTUP) {
+      int numNewStartupTasks = taskManager.getNumActiveTasksForRequest(pendingRequest.getRequestId());
+      numStartupRequests.set(numStartupRequests.get() + numNewStartupTasks);
+      totalTasks += numNewStartupTasks;
+      double startUpRequestPct = (numStartupRequests.get() * 100.0 / totalTasks);
+
+      if (startUpRequestPct > configuration.getStartUpTaskThresholdPct()) {
+        LOG.info("{}% exceeded startup task threshold of {}%", startUpRequestPct, configuration.getStartUpTaskThresholdPct());
```
darcatron: will this print out correctly? String.format() would take `%%` to print a "%"

ssalinas:
Closing this for now

PR #1293, opened 2016-09-20
Cache files with md5, add more detail for cache misses

ssalinas:
If the md5 of a file changes in s3 without the file name changing, we can run into some issues when we download and try to cache the file. A file with the same name already exists in the cache, but because the md5 does not match we download it again. Some fixes and extra info for clarify:

- [x] Include the reason we attempted to redownloaded an artifact in the exception that gets thrown
- [x] Store files in the cache with md5 when appropriate so same-named files don't overlap

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/deploy/Artifact.java
+++ b/SingularityBase/src/main/java/com/hubspot/deploy/Artifact.java
@@ -2,6 +2,7 @@ package com.hubspot.deploy;
 
 import java.util.Objects;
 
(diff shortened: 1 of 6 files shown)


ssalinas:
Update, added changes to store the file in the cache as `(md5)-(filename)` when the md5 is present on the `Artifact`. This way artifacts with the same name won't overlap in the cache. S3Downloader and the SingularityExecutor will be impacted by this change

PR #1295, opened 2016-09-20
Clean up tests + make travis build more reliable

ssalinas:
- Don't build swagger docs every time, we care mostly about tests
- fail fast (for now, may add the `fae` arg back later)

TODO:
- [x] Find test that is causing forks to time out on occasion (We were ignoring an InterruptedException in a while loop, causing it to hang forever)


ssalinas:
Noting here for reference:

The `finishNewTaskChecks()` method in `SingularitySchedulerTestBase` is what occasionally hangs. When running the test, we can get an `IllegalStateException` from the curator `TestingServer`, which causes the new task checker to trigger an abort and the executor service daemon thread to stop. Subsequent checks all get `RejectedExecutionException`s and since we check in a while loop for the futures to complete, we hang there forever.

Still looking into a fix, but that seems to be the mostly likely candidate for the issue, and is contained to the healthcheck-related tests currently

ssalinas:
Going to merge this to avoid merge conflicts as we write new tests

NOTE: Besides the small twaek to healthchecks methods, no test content or assertions were actually changed, just moved between files

PR #1299, opened 2016-09-22
Duplicate Code refactor

prabh27:
I found out two files which had a method that was 100%cloned. I refactored the clones by adding a new class, and putting the method in that class. Have a look at it.


prabh27:
Same code with 1 import removed.

PR #1300, opened 2016-09-22
abort if the status update thread has stopped running

ssalinas:
@tpetr the addition to the status update changes we were talking about
- Abort if there isn't a status update thread running since status updates are such a core piece of things
- Update the InterruptedException on adding to the queue to just keep going. Since we won't have acked the status update, it will just be sent again and we'll get it on the next go.


ssalinas:
merging this into `status-update-thread` branch now that both are in hs_qa

PR #1302, opened 2016-09-23
duplicate code refactoring

prabh27:
Duplicate code found in 2 classes. Refactored the code.


ssalinas:
Hey @prabh27 , thanks for the input here, but I think we are going to keep it as-is for the moment.

In our opinion it is easier to understand what is happening this way. The ZkDataMigration classes are for one-time changes in zookeeper (think database migrations). So, having one class per migration helps us completely separate what is happening and in what order (`migrationNumber`) without having to dig through other classes

PR #1304, opened 2016-09-27
More even distribution among racks

ssalinas:
This PR tackles 2 issues

1. When `# instances` mod `# racks` was not `0`, we could get very uneven placement (e.g. 3,3,1 for 7 instances on 3 racks. I've updated the `SlaveAndRackManager` to make sure we distribute these cases evenly. In this example, no rack would be allowed a third instance until the others all had at least 2.

2. When scaling down, we always clean up the highest instance numbers. But it is often the case that these are all on the same rack. I added code to the scheduler to (configurably) create extra cleanups when scaling down so that new tasks can be started which will fill in the correct racks based on the ones that have been left as active. While this does create a slight bit of extra churn, I think it is still an easier solution than trying to restrict racks based on instance number as new tasks are being started

/cc @tpetr @darcatron

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
@@ -433,7 +433,7 @@ public class SingularityScheduler {
           }
         }
         if (extraCleanedTasks > 0) {
-          schedule(extraCleanedTasks, matchingTaskIds, request, state, deployStatistics, pendingRequest, maybePendingDeploy);
+          schedule(extraCleanedTasks, remainingActiveTasks, request, state, deployStatistics, pendingRequest, maybePendingDeploy);
         }
       }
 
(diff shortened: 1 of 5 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -419,18 +409,56 @@ private int scheduleTasks(SingularitySchedulerStateCache stateCache, Singularity
         Collections.sort(matchingTaskIds, Collections.reverseOrder(SingularityTaskId.INSTANCE_NO_COMPARATOR)); // clean the highest numbers
       }
 
+      List<SingularityTaskId> remainingActiveTasks = new ArrayList<>(matchingTaskIds);
       for (int i = 0; i < Math.abs(numMissingInstances); i++) {
         final SingularityTaskId toCleanup = matchingTaskIds.get(i);
-
+        remainingActiveTasks.remove(matchingTaskIds.get(i));
```
darcatron: `remainingActiveTasks.remove(toCleanup)`?

ssalinas: 👍 updated

PR #1306, opened 2016-09-27
Smarter Healthchecks

### LINKED ISSUE #1287: Singularity should support healthchecks on statically defined service ports
For some services we need to define the ports statically.  While we can do this with network=BRIDGE mode services with static portMappings, this is not possible with network=HOST mode services since the current healthchecks are based on ports defined in portMappings.  So for network=HOST services it would be nice to be able to specify one or more healthchecks that can include a statically defined port.

ssalinas:
This PR attempts to do a few things regarding our existing healthcheck setup
- Split out the configuration into its own object so we stop cluttering the SingularityDeploy
- Add the concept of a `startup` period (identified primarily by 'connection refused' responses vs actuall http responses)
  - Add a separate `startupDelay` (time before running any checks) and `startupInterval` (how often to check during this time)
  - Add a `startupTimeout` - time limit for app to actually respond to a check
- Make the `healthcheckTotalTimeoutSeconds` a calculated field based on the others present rather than an override
- Ability to specify a specific port number for a healthcheck (/fixes #1287)

Still TODO:
- [x] Add validation on `maxRetries` \* `intervalSeconds` (as a replacement for having a separate static max total timeout)
- [x] Finish wiring up `startupTimeout` in deploy and new task checkers
- [x] Smarter reactions based on status code
- [x] Tests specific to startup period and status codes
- [x] Clean up excess startup healthchecks since they provide no extra information
- [x] Ensure new task checks run at an appropritate interval
- [x] Update deploy form
- [x] Update healthcheck notifications and messages in UI
- [x] Fix healthcheck/deploy checker timeout race condition

/cc @tpetr @darcatron


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityHealthcheckAsyncHandler.java
```diff
@@ -54,13 +56,18 @@ public Response onCompleted(Response response) throws Exception {
...
+    saveResult(Optional.<Integer> absent(), Optional.<String> absent(), Optional.of(String.format("Healthcheck failed due to exception: %s", t.getMessage())), Optional.of(t));
   }
 
-  public void saveResult(Optional<Integer> statusCode, Optional<String> responseBody, Optional<String> errorMessage) {
+  public void saveResult(Optional<Integer> statusCode, Optional<String> responseBody, Optional<String> errorMessage, Optional<Throwable> throwable) {
+    boolean inStartup = false;
+    if (throwable.isPresent() && throwable.get() instanceof ConnectException) {
+      inStartup = true;
+    }
+
```
darcatron: `boolean inStartup = throwable.isPresent() && throwable.get() instanceof ConnectException`

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeployBuilder.java
```diff
@@ -40,16 +41,25 @@
...
+  @Deprecated
   private Optional<Integer> healthcheckPortIndex;
-  private Optional<Boolean> skipHealthchecksOnDeploy;
+  @Deprecated
   private Optional<HealthcheckProtocol> healthcheckProtocol;
-
+  @Deprecated
   private Optional<Integer> healthcheckMaxRetries;
+  @Deprecated
   private Optional<Long> healthcheckMaxTotalTimeoutSeconds;
```
darcatron: `@Deprecated use {@link healthcheck}`?

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
```diff
@@ -49,16 +50,25 @@
...
+  @Deprecated
   private final Optional<Integer> healthcheckPortIndex;
-  private final Optional<Boolean> skipHealthchecksOnDeploy;
+  @Deprecated
   private final Optional<HealthcheckProtocol> healthcheckProtocol;
-
+  @Deprecated
   private final Optional<Integer> healthcheckMaxRetries;
+  @Deprecated
   private final Optional<Long> healthcheckMaxTotalTimeoutSeconds;
```
darcatron: `@Deprecated use {@link healthcheck}`?

ssalinas:
First frontend+backend draft of this is done now (save for the race condition fix). The setup of the new healthchecks looks like this:
screen shot 2016-10-20 at 12 23 21 pm

UI messages are also updated to reflect this. Some examples:

- Failed because the app never responded (never left the startup phase). This was with a 10s `startupDelay` and 15s `startupTimeout` set. Note that only a single healthcheck appreas in the table. This is because startup phase healthchecks are cleaned up after task checks run (except if it is the last healthcheck). This way we can check more often during the startup phase without cluttering healthchecks.
screen shot 2016-10-20 at 10 51 33 am
--------
screen shot 2016-10-20 at 10 51 41 am

- Failed due to a status code in the failureStatusCodes list
screen shot 2016-10-20 at 10 51 26 am
screen shot 2016-10-20 at 10 51 08 am

- Failed due to too many retries
screen shot 2016-10-20 at 10 55 51 am


Open to any feedback on the ui messages, overall setup, or additional features before we merge this into any branches

@tpetr @darcatron

ssalinas:
Since we recently did a release, going to merge this and I will address any bugs we find in future PRs. Been looking good in hs_stable though.

nyarly:
@ssalinas I don't understand this new behavior at all. Is the reasoning documented somewhere?

Specifically, from what I can tell, once we leave startup, it looks like we poll the healthcheck URL looking for failure until the HC timeout elapses. What's the reasoning behind that? i.e. after one successful healthcheck, why not transition to "started"?

stevenschlansker:
Hm, my understanding is that's exactly what happens -- any single successful check turns it to running state and no further checks will be performed.

nyarly:
I skimmed the code and it seemed to be what I was concerned by. In that case, why are there multiple polls then? Because I was pretty sure that a single failure status (as defined by a configuration) marks the service UNHEALTHY, which I assumes means Singularity kills the task.

ssalinas:
There's a few things to divide up here:
- In terms of the task state, we only start health checks once the task enters `TASK_RUNNING`. This just means the process is running not that it is healthy yet
- It is correct that we stop after a single successful check

In the old method of healthchecks there was just a total amount of time/number of checks as the constraint. Now, there is a constraint around startup (i.e. time until we stop getting connection refused responses) and constraint around the actual response (i.e. time until a 'successfull' status code is received)

ssalinas:
A single failure status does _not_ mark the task as unhealthy unless you set it to do so. Once it has received some sort of valid http response, you then have `maxRetries` attempts left for the app to return an successful check

PR #1307, opened 2016-09-27
Fix decom in SingularityClient

pranav:
@ssalinas 
unless there's a better way we want to do this? In chrome, when we hit decom, it sends `{}` as the request body, but when I use the client now, it sends over nothing and the API sends a 415 error


ssalinas:
@pranav closing in favor of #1308

PR #1308, opened 2016-09-28
Get the SingularityClient up to date

ssalinas:
Started with the slave/rack methods for @pranav . Will continue to go through the other endpoints and add them to this PR as well.

### DIFF
--- a/SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
+++ b/SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
@@ -757,10 +757,10 @@ public class SingularityClient {
 
   @Deprecated
   public void decomissionSlave(String slaveId) {
-    decomissionSlave(slaveId, Optional.<SingularityMachineChangeRequest>absent());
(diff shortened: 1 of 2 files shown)


SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
```diff
@@ -774,10 +755,27 @@ public void deleteDeadRack(String rackId) {
     return getSingle(requestUri, "slave", slaveId, SingularitySlave.class);
   }
 
+  @Deprecated
   public void decomissionSlave(String slaveId) {
+    decomissionSlave(slaveId, Optional.<SingularityMachineChangeRequest>absent());
+  }
+
+  public void decomissionSlave(String slaveId, Optional<SingularityMachineChangeRequest> machineChangeRequest) {
```
pranav: spelling `decommission`

ssalinas: updated the new methods, left the old deprecated ones spelled the same since they were already present in the client

PR #1317, opened 2016-10-11
use saveTaskCleanup() instead of createTaskCleanup() for deploys

tpetr:
`createTaskCleanup()` won't override a preexisting SingularityTaskCleanup, which can be problematic for deployments. This PR updates `SingularityDeployChecker.cleanupTasks()` to use `saveTaskCleanup()`, which *will* override cleanups. This prevents race conditions where an old deploy's task sticks around indefinitely while waiting for a bounce to complete.


ssalinas:
Fixed the merge conflict in the tests, going to merge this

PR #1318, opened 2016-10-14
WIP - Search UI Improvements

darcatron:
Issues to fix:
1. Task Search Pagination - when the number of results is divisible by the number of results per page, the pagination buttons display an "extra page" (if we have 20 results with 10 per page on page 2 you can click the ">" but it refreshes the same data) that just refreshes to the same page and then removes the ">" button. This can cause confusion about the results.
2. Date - date parsing fails if there are extra spaces. Date isn't removed when clear form is clicked. If a date filter is used and then removed, it causes a parsing error.
3. UITable uses - other uses of UITable need to be updated to the new api to work properly
- [x] Task Search Pagination
- [ ] Date
- [x] UITable


ssalinas:
This one will need to be redone a bit, closing for now

PR #1319, opened 2016-10-17
Make the history purger query more efficient

ssalinas:
Currently the purger query can grow in an unbounded manner when in the mode to set bytes to `''` rather than deleting, since the query just goes by `updatedAt < x`. This PR updates the `taskHistory` table to have an additional indexed `purged` column. This way we can limit our query to only tasks we have not already purged.

/cc @tpetr @darcatron


ssalinas:
Removed the update statement from the migration. When the purger first runs, the relevant rows will be updated and all subsequent runs of the purger will be faster

PR #1320, opened 2016-10-17
Ability to run shell command before killing task

ssalinas:
On a request pause, request bounce, or task bounce, it can be useful to run something like a heap dump before the task shuts down. This PR adds the ability to request that a shell command run and finish before a task is killed.

Still TODO:
- [x] Add shell command ui options to bounce/pause/kill modals

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskShellCommandUpdate.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskShellCommandUpdate.java
@@ -7,7 +7,17 @@ import com.google.common.base.Optional;
 public class SingularityTaskShellCommandUpdate {
 
   public enum UpdateType {
-    INVALID, ACKED, STARTED, FINISHED, FAILED;
+    INVALID(true), ACKED(false), STARTED(false), FINISHED(true), FAILED(true);
+
+    private boolean finished;
+
+    UpdateType(boolean finished) {
+      this.finished = finished;
+    }
+
+    public boolean isFinished() {
+      return finished;
+    }
   }
 
   private final SingularityTaskShellCommandRequestId shellRequestId;
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskShellCommandUpdate.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskShellCommandUpdate.java
@@ -9,7 +9,7 @@ public class SingularityTaskShellCommandUpdate {
   public enum UpdateType {
     INVALID(true), ACKED(false), STARTED(false), FINISHED(true), FAILED(true);
 
-    private boolean finished;
+    private final boolean finished;
 
     UpdateType(boolean finished) {
       this.finished = finished;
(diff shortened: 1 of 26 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityBounceRequest.java
```diff
@@ -5,23 +5,27 @@
...
 
   private final Optional<Boolean> incremental;
   private final Optional<Boolean> skipHealthchecks;
+  private final Optional<SingularityShellCommand> runBeforeKill;
 
   @JsonCreator
   public SingularityBounceRequest(@JsonProperty("incremental") Optional<Boolean> incremental, @JsonProperty("skipHealthchecks") Optional<Boolean> skipHealthchecks,
-      @JsonProperty("durationMillis") Optional<Long> durationMillis, @JsonProperty("actionId") Optional<String> actionId, @JsonProperty("message") Optional<String> message) {
+      @JsonProperty("durationMillis") Optional<Long> durationMillis, @JsonProperty("actionId") Optional<String> actionId, @JsonProperty("message") Optional<String> message,
+      @JsonProperty("runBeforeKill") Optional<SingularityShellCommand> runBeforeKill) {
```
tpetr: i think it'd be better to more explicit here, perhaps `runShellCommandBeforeKill`

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskShellCommandUpdate.java
```diff
@@ -7,7 +7,17 @@
 public class SingularityTaskShellCommandUpdate {
 
   public enum UpdateType {
-    INVALID, ACKED, STARTED, FINISHED, FAILED;
+    INVALID(true), ACKED(false), STARTED(false), FINISHED(true), FAILED(true);
+
+    private boolean finished;
```
tpetr: `private final boolean`

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -549,4 +561,30 @@ public void checkRequestGroup(SingularityRequestGroup requestGroup) {
...
+  public void checkValidShellCommand(final SingularityShellCommand shellCommand) {
+    Optional<ShellCommandDescriptor> commandDescriptor = Iterables.tryFind(uiConfiguration.getShellCommands(), new Predicate<ShellCommandDescriptor>() {
+      @Override
+      public boolean apply(ShellCommandDescriptor input) {
+        return input.getName().equals(shellCommand.getName());
+      }
+    });
+
+    if (!commandDescriptor.isPresent()) {
+      throw WebExceptions.forbidden("Shell command %s not in %s", shellCommand.getName(), uiConfiguration.getShellCommands());
```
ssalinas: I can update that. FYI, this was copy-pasted from the original shell command code in TaskResource, I just moved it to the validator instead

tpetr: Ah gotcha. Thanks.

ssalinas: updated

PR #1324, opened 2016-10-24
extra filtering on fuzzy match

ssalinas:
Replacement for #1228 . Users are used to the fuzzy matching, I think we just need to slim down the results a bit more. The fuzzy lib has no threshold for score. Instead its condition for including the item is if all characters match somewhere in the string.

I consolidated our fuzzy filter tweaks into utils so we don't duplicate things, and added one additional filter to exclude more low-scoring results, especially when we believe we have a good match.


ssalinas:
Search results are much more targeted now, but you still get what you're looking for with an accidental typo. Think this one is good to merge

PR #1325, opened 2016-10-24
Refresh task lists appropriately

ssalinas:
A different way of approaching #1243 . With a periodic refresh of the task history table, we still aren't guaranteed that the refresh will happen at an optimal time to fix the issue of tasks missing from active and not appearing in history. Since we know the situation where we want to refresh the table, I implemented it in that fashion. This will trigger some function whenever the size of the data in the table changes. In this case, it triggers the refresh of the task history if the number of active tasks changes.

@tpetr @darcatron let me know what you think of this approach

### DIFF
--- a/SingularityUI/app/components/common/table/UITable.jsx
+++ b/SingularityUI/app/components/common/table/UITable.jsx
@@ -28,11 +28,18 @@ class UITable extends Component {
     this.handlePageChange = this.handlePageChange.bind(this);
   }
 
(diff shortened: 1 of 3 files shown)


ssalinas:
Extended this to also refresh active tasks after certain request actions (pause, run now, etc) so that we see an up-to-date list

ssalinas:
This caught all the obvious missing refreshes after actions that I could find, any others I will do in a subsequent PR

PR #1326, opened 2016-10-27
Fix js TypeError on task detail page

ishabalin:
Submitting a deploy with no "resources" constraint would result in a js error on the task detail page

```
TypeError: undefined is not an object (evaluating 't.task.taskRequest.deploy.resources.numPorts')
```


ssalinas:
👍 Thanks for the fix, I'll move this up though our test branches

PR #1331, opened 2016-10-29
Make task destroy work from the ui

ssalinas:
Seems this got lost in the rewrite. 
- Updates the ui to make sure task destroys work correctly to trigger the destroy task framework messages.
- Hides the wait for replacement when clicking destroy task. If you are on the destroy modal, you are likely looking for instant and the wait for replacement being checked can confuse the user as to why nothing is happening.
- Refresh the task history and cleanups immediately after triggering kill

/cc @tpetr @darcatron

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
@@ -429,7 +429,7 @@ public class SingularityExecutorMonitor {
 
     try {
       if (!wasKilled) {
-        task.markKilled();
+        task.markKilled(user);
       }
 
       processBuilderFuture = processBuildingTasks.get(task.getTaskId());
@@ -454,7 +454,7 @@ public class SingularityExecutorMonitor {
         } else {
           task.getLog().info("Destroying process with pid {} for task {}", runningProcess.getCurrentPid(), taskId);
         }
-        task.markForceDestroyed();
+        task.markForceDestroyed(user);
         runningProcess.signalKillToProcessIfActive();
         return KillState.DESTROYING_PROCESS;
       }
@@ -519,6 +519,7 @@ public class SingularityExecutorMonitor {
       public void onSuccess(Integer exitCode) {
         TaskState taskState = null;
         String message = null;
+        Optional<String> maybeKilledBy = task.getKilledBy();
 
         if (task.wasKilledDueToThreads()) {
           taskState = TaskState.TASK_FAILED;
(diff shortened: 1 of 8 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
```diff
@@ -411,11 +411,19 @@ public int getNumHealthchecks(SingularityTaskId taskId) {
...
   public SingularityCreateResult saveTaskHistoryUpdate(SingularityTaskHistoryUpdate taskHistoryUpdate) {
+    return saveTaskHistoryUpdate(taskHistoryUpdate, false);
+  }
+
+  @Timed
+  public SingularityCreateResult saveTaskHistoryUpdate(SingularityTaskHistoryUpdate taskHistoryUpdate, boolean overwrite) {
     singularityEventListener.taskHistoryUpdateEvent(taskHistoryUpdate);
 
-    return create(getUpdatePath(taskHistoryUpdate.getTaskId(), taskHistoryUpdate.getTaskState()), taskHistoryUpdate, taskHistoryUpdateTranscoder);
+    if (overwrite) {
```
ssalinas: As an fyi, I added this in because I had noticed that the history was no longer indicative of what killed the tasks if other manual steps were taken. For example, the cleaning would be started by a decom, but the task would actually be killed from a user request. Only task cleaning updates use the `save` instead of `create`

tpetr: makes sense -- can we further clarify the naming to something like `overwriteExisting`?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
```diff
@@ -532,7 +533,11 @@ public void onSuccess(Integer exitCode) {
 
             message = String.format("Task killed forcibly after waiting at least %s", JavaUtils.durationFromMillis(millisWaited));
           } else if (task.wasForceDestroyed()) {
-            message = "Task killed forcibly after multiple kill requests from framework";
+            if (maybeKilledBy.isPresent()) {
+              message = String.format("Task killed forcibly by %s", maybeKilledBy.get());
```
ssalinas: Because in this case the task was forcibly killed by a user, there were not necessarily multiple requests from the framework. `maybeKilledBy` is not present whe framework actions are killing a task, only when it is initiated by a user

SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
ssalinas: updated

PR #1332, opened 2016-10-31
Ensure quotes and new lines are escaped in echo

ssalinas:
So, we had previously made this a bit more resilient to weird new lines in environment variables not being loaded properly in docker by adding them to command line args instead. But, with the right order or quotes/newlines/etc, we can still run into issues at the `echo` statement before the actual command is executed. This adds escaping for the double quotes as well as calling `echo -e` so that the escaped characters are expanded properly when printed.


ssalinas:
@tpetr , renamed the helper to be more descriptive here

PR #1333, opened 2016-10-31
Remove OOMKiller and LogWatcher Modules

ssalinas:
Both are currently unused

/cc @tpetr @wsorenson


ssalinas:
Also removing the old vagrant stuff and some left over cookbook files as part of this.

@eherot all of the cookbook stuff is in your repo now right? Couldn't find any connection to the last few files in the cookbook folder here, but wanted to double check.

eherot:
Correct as far as I understand it.

eherot:
So I didn't realize this was relevant to us at the time but did you guys replace this LogWatcher stuff with something else or did you just get tired of maintaining it?

PR #1335, opened 2016-11-02
replace instead of ignore so that we can update the offset

tpetr:
@jhaber what do you think about this tweak? using `REPLACE` instead of `IGNORE` allows us to update the `offset` when we skip over a bytes in the beginning of a chunk.

/cc @ssalinas #1334

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
@@ -117,9 +117,9 @@ public class SandboxManager {
   }
 
   /**
-   * this method will first sanitize the input by throwing away invalid UTF8 characters before sending it to
-   * Jackson for parsing. We can get invalid UTF8 data if there is a multibyte character at the boundary
-   * of our fetch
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java
```diff
@@ -115,17 +117,47 @@ public SlaveNotFoundException(Exception e) {
...
-   * of our fetch
+   * This method will first sanitize the input by replacing invalid UTF8 characters with \ufffd (Unicode's "REPLACEMENT CHARACTER")
+   * before sending it to Jackson for parsing. We then strip the replacement characters characters from the beginning and end of the string
+   * and increment the offset field by how many characters were stripped from the beginning.
    */
   @VisibleForTesting
   MesosFileChunkObject parseResponseBody(Response response) throws IOException {
     // not thread-safe, need to make a new one each time;
-    CharsetDecoder decoder = StandardCharsets.UTF_8.newDecoder().onMalformedInput(CodingErrorAction.IGNORE);
+    CharsetDecoder decoder = StandardCharsets.UTF_8.newDecoder().onMalformedInput(CodingErrorAction.REPLACE);
```
jhaber: Maybe call `replaceWith(REPLACEMENT_CHARACTER)` to be more explicit?

PR #1350, opened 2016-11-18
Docs updates

### LINKED ISSUE #384: API Docs for SingularityDeploy should include defaults
I.e. it would be helpful if there were a "default" column in the table.  Examples might also be handy in cases where there are no defaults.

### LINKED ISSUE #967: Swagger docs - invalid $ref fields
There are several fields in both the swagger JSON api documents and the markdown documentation that aren't valid - $ref fields on properties (as I understand it, and per https://github.com/OAI/OpenAPI-Specification/blob/master/versions/1.2.md#433-data-type-fields) have to refer to a model id, and several in the docs are raw Java types. 

For example, https://github.com/HubSpot/Singularity/blob/master/Docs/reference/apidocs/models.md#-commandinfo includes
urisOrBuilderList: List[? extends org.apache.mesos.Protos$CommandInfo$URIOrBuilder]

Maybe that should be "array" with an items field as well? Or else the field should be excluded from the Swagger docs?

### LINKED ISSUE #1288: api-docs link not working
The trailing slash appears to have been dropped from the API Docs link. Consequently the browser tries to load http://singularity.local:7092/strapdown.js instead of http://singularity.local:7092/api-docs/strapdown.js, which doesn't work for obvious reasons.

ssalinas:
Bunch of cleanup in this PR:
- Remove deprecated `SingularityContainerInfo` constructor
- Add a large amount of missing `@ApiModelProperty` annotations
- Add a deploy defaults doc to point out fields that fall back to a system-wide or request-level config setting (/fixes #384)
- Don't try and generate swagger docs for any org.apache.mesos classes (/fixes #967)
- Don't hide overflow of tables in the gitbook 
- Fix default api-docs link (/fixes #1288)

### DIFF
--- a/Docs/reference/api.md
+++ b/Docs/reference/api.md
@@ -1292,6 +1292,29 @@ Update the skipHealthchecks field for the request, possibly temporarily
 
 
 - - -
+#### **DELETE** `/api/requests/request/{requestId}/skip-healthchecks`
+
+Delete/cancel the expiring skipHealthchecks. This makes the skipHealthchecks request permanent.
+
+
+###### Parameters
+**path**
+
+| Parameter | Required | Description | Data Type |
+|-----------|----------|-------------|-----------|
+| requestId | true | The Request ID | string |
+
+###### Response
+[SingularityRequestParent](#model-SingularityRequestParent)
+
+
+###### Errors
+| Status Code | Reason      | Response Model |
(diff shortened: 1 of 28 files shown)


ssalinas:
Going to merge this one and publish an update to the docs

Issue #1323, opened 2016-10-24
Can't deploy services to all hosts when using fixed ports


relistan:
I may just be doing something wrong here so if that's the case, advice is appreciated. But what I'm seeing is that I am not able to deploy services with fixed ports if an existing service is already running. Here's the scenario:

* Service in a Docker container
* Running in HOST networking mode
* Binding to a fixed port on each host where it runs (this is hard requirement)

I want to deploy with the rolling deployment where it handles one instance at a time. But it doesn't seem like I can tell the deployment to bring an existing instance down before trying to start the new instance. When you are using a service with a fixed port, this means the deployment is a non-starter. I currently have these settings:

```json
        "deployInstanceCountPerStep": 1,
        "deployStepWaitTimeMs": 10000,
        "autoAdvanceDeploySteps": true
```

Even without those settings, though, with the non-rolling deploy it of course has the same problem. Is there a setting I'm missing? I've been all over the docs and code and don't see one. Here's the status for the failed deployment:

 > Only 0 of 2 tasks could be launched for deploy, there may not be enough resources to launch the remaining tasks

ssalinas:
@relistan How many slaves are you running vs # of instances in that deploy, and are there any other constraints (attributes/etc) on the placement? My first thought is that you are either generally low on resources, or some other constraint is limiting where the tasks can run in such a way that you can't find enough free space to launch one.

 If you look in the Singularity logs, for each offer that comes in there should be a `TRACE` level line like `Rejecting SEPARATE task {request id} from slave {slave} ...` or any other `TRACE` level log line from the `SingularitySlaveAndRackManager` class that might give some more insight.

relistan:
Thanks @ssalinas . There were two things going on and the resource message was a red herring. However the main issue still remains so I'll try to re-explain the scenario:

 * The service needs to run on all of the slaves or as nearly as possible
 * Needs to bind on a fixed port

The problem is that in this scenario I need to be able to take down an instance _before_ starting a new one. If I don't, the port will not be free on potentially any hosts, and therefore no new instances will be able to start in the new deploy. 

[...]

Issue #469, opened 2015-03-02
investigate a new request type for running a task on every slave


tpetr:
I think this fills an important use case, but we need to figure out:
- how to best prioritize resource allocation between the different request types
- how to best visualize this type of request, especially since it could have a large # of tasks associated with it

/cc @stevenschlansker

stevenschlansker:
Thanks!

apognu:
I would be very interested in that feature, if it's still on the roadmap.

bradjorgensen:
+1

ozayduman:
+1

stevenschlansker:
Is this fixed by #1519 ?

ssalinas:
correct, see a bit more in [this comment](https://github.com/HubSpot/Singularity/pull/1519#issue-225079969) or the docs as to how to use it

PR #1359, opened 2016-12-07
Add support for spread task to all slaves.

felixgborrego:
Still WIP. Add a new slave placement strategy useful when you want to deploy a task to all host. Related with issues [#469](https://github.com/HubSpot/Singularity/issues/469) and [#1323](https://github.com/HubSpot/Singularity/issues/1323)

The new SlavePlacement type `SPREAD_ALL_SLAVES` has the some behaviour as `SEPARATE_BY_REQUEST` but the new poller auto scale up/down the number of instances if `request.instances != currentActiveSlaveCount`


ssalinas:
From the Singularity side of things I think the poller approach is ok, but it would only be a best-effort attempt to schedule the tasks on each slave.

One situation I can see is that we have 3 slaves, but one of them is 'full' (i.e. it offers 10GB mem and all 10 are already allocated to other tasks). We will no longer be getting any offers from this slave with enough memory to launch a new task because there is none left. So, we will have a pending task waiting to be launched until something else moves off that slave and frees up resources. Some possible solutions for this that I can think of:

- Option to move tasks off of a slave to free up memory. However, especially in a multi-framework environment, this still does not guarantee that we will get offered those resources
- Implement use of the [reserved resources functionality in mesos](http://mesos.apache.org/documentation/latest/reservation/). Downside here is that it starts to introduce significant complexity trying to ensure that we have reserved the appropriate amount of resources from each slave for all possible requests with `SPREAD_ALL_SLAVES` (i.e. how to gracefully deploy to all servers since we will need extra memory on each host while the deploy is in-progress)

Other things we'd need for this:
- Validation when updating a request in the RequestResource (either changing placement strategy, or updating scale) that the new scale is correct for a request with `SPREAD_ALL_SLAVES`. Alternatively, ignore the scale as set in the update and set it ourselves based on the presence of `SPREAD_ALL_SLAVES`

I think it's a good start for sure, we'll just need a bit more strategy if we want to guarantee placing a task on each slave vs having it be best-effort

felixgborrego:
Hi @ssalinas,

As you said, this approach based on poller will eventually match number of instances with the number of Mesos slaves if possible, but there is no actual guaranty.  

For our concrete use case, the reserved resources functionality in Mesos sounds like a good alternative. I don't know enough about Singularity but It doesn't sound like a massive change,  for what I can see we need to update `SingularityMesosScheduler.match`  to take the role into account, and configure new slaves with some reserved resources so we can guaranty `SPREAD_ALL_SLAVES` will run the task in all host.

I'll prototype the reserved resources solution to see how it could look like but I rather keep this change simple. As you pointed out, it may add unnecessary complexity. And actually, reserved resources is a nice functionality interesting for other placement strategies too, we could consider adding it in a different PR and live with `best-effort attempt` here.

ssalinas:
Pinging @tpetr here as well in case he has any other ideas on a possible implementation here

felixgborrego:
Thanks @ssalinas @tpetr,

Added logic to auto scale the number of instances when using `SPREAD_ALL_SLAVES` (for creating or updating a request).

As for the usage of Mesos reserved resources, after prototyping it, It looks to me out of the scope of this PR. It's actually a large change, I'm preparing another PR to support reserved resources so we can discuss it separately.

Any other change or considerations `SPREAD_ALL_SLAVES ` may need?

felixgborrego:
Just updated this PR rebasing from master.

Your main concern about 'best-effort attempt' is addressed by #1374. We can guaranty that `SPREAD_ALL_SLAVES` will be able to allocate resources for request with `requiredRole` using Mesos reserved resources.


 Any other feedback on this PR @ssalinas ?

ssalinas:
Hey @felixgborrego , thanks for the effort you put in to this. @tpetr and I are still trying to think if there are cleaner ways of working with our existing pollers that wouldn't require automatically mutating the request instances. While this strategy works, it doesn't follow the approach we normally take where only a user action should update the Singularity request data. Also, the poller gives us a strategy that is 'eventually consistent' once the poller runs, while an ideal strategy would give us a tighter coupling with the lifecycle of the mesos slaves.

We'll start moving along the required roles PR while we think and try to come back with a more final strategy or comments next week.

relistan:
@ssalinas I agree with the first point:
>While this strategy works, it doesn't follow the approach we normally take where only a user action should update the Singularity request data.

It would be great to not have to mutate the request in the background. But as for the second point:
> the poller gives us a strategy that is 'eventually consistent' once the poller runs, while an ideal strategy would give us a tighter coupling with the lifecycle of the mesos slaves.

Anything we do here is going to be eventually consistent since it can't happen prior to the slave being added. By the nature of it, we can't deploy anything to the new slave until it's added. The poller has the advantage of being very simple and using mechanisms that are already in use. Ideally there would be an event we could have a callback on, to remove the delay inherent in the poller timing. But, in both cases the deployment will still have to run async and be eventually consistent.

felixgborrego:
@ssalinas, any thought or update in this one?
If you think it's way to go, I'm open to try to prototype the solution using the lifecycle of the Mesos slaves.

ssalinas:
Hey, sorry for the slow response. I think we can give the poller a try, just need to make sure we document the mechanics and the fact that it isn't necessarily guaranteed.

I'll give this a test in our staging environment on Monday 👍

relistan:
Awesome @ssalinas, this would be great to have in the mainline codebase.

felixgborrego:
@ssalinas updated the script, so it's easier to test as a container.

felixgborrego:
@ssalinas Is there any change to have this merged or reevaluated at some point? we are already running a custom build with this PR on production.
Also, I see 0.14.0 is out, is the support for Mesos Reserved Resources included #1374 on 0.14.0?

felixgborrego:
@ssalinas let me know if you have any feedback or if we can help you in any way to push this PR forward. We have been running this change in production for almost a month so we are still pretty interested in finding the right approach for `SPREAD_ALL_SLAVES` and getting this proposal into Master.

ssalinas:
Hey @felixgborrego apologies for the delay again on this. 
having run it in prod for a while, I'm curious if you have seen any cases where a task will fail to launch on all slaves. Interested to see if there are any edge cases we may be missing that should be addressed

felixgborrego:
@ssalinas the only edge case, I can think of is:
If a server goes down or lost connectivity while a deployment with `SPREAD_ALL_SLAVES` is happening, the system ends up in a situation where the number of pending tasks is bigger than the number of instances of the current deployment.  In this case, we need somehow to cancel the current deploy (or update the numInstances)




SLAVE_SATURATED

ssalinas:
Moved all these commits plus a few more to #1519 which will be part of release 0.15.0

PR #1365, opened 2016-12-13
Fix PATH in API annotations

### LINKED ISSUE #1363: Duplicate link to /api/requests in the documentation pointing instead to the Request Groups doc?
screen shot 2016-12-11 at 18 56 09

This may be a result of a recent API change, but the above looks like some unintended duplication

ssalinas:
Looks like these never got changed when we added the new resources.

/fixes #1363


ssalinas:
merging this, will republish the gitbook for getsingularity.com afterwards as well

PR #1369, opened 2016-12-13
Additional threshold for deleting task history row

ssalinas:
Building off of the previous history purge update PR and I'll merge the two. This separates out settings for deleting task history bytes versus the entire task history row in the db. Rows to delete are checked/purged first, followed by rows to update bytes.

`deleteTaskHistoryBytes...` settings take on the previous default values of 365 days/10k tasks since that was the previous default behavior
`deleteTaskHistory...` settings default to 0 so the rows are not deleted by default


ssalinas:
This PR is into the `task_history_purge_updates` branch, not master. Going to merge it in there now that both are in hs_qa

PR #1373, opened 2016-12-13
customizable --use-compress-program for tar

tpetr:
This PR allows you to override the compression program used by `tar` for decompressing tarballs.


ssalinas:
this has been fine in hs_stable for a while, going to merge

PR #1374, opened 2016-12-14
Add support for Role / Mesos Reserved Resources

felixgborrego:
This PR aims at adding support for Mesos Reserved Resources to Singularity. Related with #1359 

The main change is in `MesosUtils.doesOfferMatchResources` that now takes into account  Singularity `Request.requiredRole` and Mesos Offer `Resources.Role`

The UI is still missing, and I'm not sure if this approach is the right one, but hopefully this PR can help us to discuss it.

We need to think about the semantic of "required role" for a Request.  The current implementation will not accept offers with `*` for a Request with `requiredRole`, but we may want to support other use cases here.

sidekick @ssalinas @relistan

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
+++ b/SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
@@ -9,7 +9,10 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Random;
+import java.util.Set;
+import java.util.Iterator;
 
+import com.google.common.collect.Sets;
 import org.apache.mesos.Protos.MasterInfo;
 import org.apache.mesos.Protos.Offer;
 import org.apache.mesos.Protos.Resource;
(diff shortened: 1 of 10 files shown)


relistan:
That last commit looks good to me @felixgborrego

felixgborrego:
Any feedback for this PR @ssalinas ?

SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
```diff
@@ -39,17 +42,44 @@ private static double getScalar(Resource r) {
...
-      if (r.hasName() && r.getName().equals(name) && r.hasScalar()) {
+      if (r.hasName() && r.getName().equals(name) && r.hasScalar() && isRequiredRole(r, requiredRole)) {
         return getScalar(r);
       }
     }
 
     return 0;
   }
 
+  private static Boolean hasRole(Resource r) {
```
tpetr: can you switch to primitive `boolean`s instead? removes the potential for NullPointerExceptions

SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
```diff
@@ -183,33 +215,65 @@ public static Resource getPortsResource(int numPorts, List<Resource> resources,
...
+    }
+
+    return builder.build();
   }
 
   private static Resource newRange(String name, long begin, long end) {
     return Resource.newBuilder().setName(name).setType(Type.RANGES).setRanges(Ranges.newBuilder().addRange(Range.newBuilder().setBegin(begin).setEnd(end).build()).build()).build();
   }
 
+  public static String getRolesInfo(Offer offer) {
```
tpetr: it looks like you're just using this in logging statements -- I think the `toString()` of `getRoles(offer)` should suffice

tpetr:
Thanks for the PR! For completeness, we should also support setting the framework's desired role in `FrameworkInfo`.

SingularityBase/src/main/java/com/hubspot/mesos/MesosUtils.java
```diff
@@ -39,17 +42,44 @@ private static double getScalar(Resource r) {
...
     }
 
     return 0;
   }
 
+  private static Boolean hasRole(Resource r) {
+    return r.hasRole() && !r.getRole().equals("*");
+  }
+
+  private static Boolean isRequiredRole(Resource r, Optional<String> requiredRole) {
```
tpetr: this method name is a little confusing to me -- it determines whether or not the resource is of the desired role -- would you be open to renaming it to `hasRequiredRole()` or `isResourceMatchesRequiredRole()`?

ssalinas:
After @tpetr 's comments I think this one should be gtg. I'll start moving it into our test environments after that 👍

felixgborrego:
Updated my PR based on comments, but I'm afraid we may need to rethink/reconsider this one.

I run into an *unexpected limitation* while testing this PR.

It looks like role-based reservations are only offers to frameworks with that role but frameworks can only have one single rol, see: https://issues.apache.org/jira/browse/MESOS-1763

felixgborrego:
Just as a clarification, even with the limitation of one role per framework, the support for Mesos Reserved Resources in Singularity is still useful as it can guaranty that some resources will be assigned to a specific Singularity Request (#1359). Just limited only to one single role + "*"

We can always extend this support to multi-role once [#MESOS-1763](https://issues.apache.org/jira/browse/MESOS-1763) is done.

PR #1377, opened 2016-12-15
Report currentActiveInstances on SingularityDeployProgress

darcatron:
This reports that a canary deploy has completed the first step


SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -134,8 +138,13 @@ public SingularityRequestParent deploy(@ApiParam(required=true) SingularityDeplo
 
     SingularityPendingDeploy pendingDeployObj = new SingularityPendingDeploy(deployMarker, Optional.<SingularityLoadBalancerUpdate> absent(), DeployState.WAITING, deployProgress, updatedValidatedRequest);
 
-    checkConflict(deployManager.createPendingDeploy(pendingDeployObj) != SingularityCreateResult.EXISTED,
-        "Pending deploy already in progress for %s - cancel it or wait for it to complete (%s)", requestId, deployManager.getPendingDeploy(requestId).orNull());
+    boolean deployExists = deployManager.createPendingDeploy(pendingDeployObj) == SingularityCreateResult.EXISTED;
+    if (deployExists && forceDeploy.or(false)) {
+      cancelDeploy(requestId, deployManager.getInUseDeployId(requestId).orNull());
```
darcatron: Unsure if there is a restart method or if I need to do something after hitting the cancel

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeployProgress.java
```diff
@@ -56,21 +66,22 @@ public long getTimestamp() {
   }
 
   public SingularityDeployProgress withNewInstances(int instances) {
-    return new SingularityDeployProgress(instances, deployInstanceCountPerStep, deployStepWaitTimeMs, false, autoAdvanceDeploySteps, failedDeployTasks, System.currentTimeMillis());
+    return new SingularityDeployProgress(instances, currentActiveInstances, deployInstanceCountPerStep, deployStepWaitTimeMs, false, autoAdvanceDeploySteps, failedDeployTasks, System.currentTimeMillis());
   }
 
   public SingularityDeployProgress withCompletedStep() {
-    return new SingularityDeployProgress(targetActiveInstances, deployInstanceCountPerStep, deployStepWaitTimeMs, true, autoAdvanceDeploySteps, failedDeployTasks, System.currentTimeMillis());
+    return new SingularityDeployProgress(targetActiveInstances, currentActiveInstances + 1, deployInstanceCountPerStep, deployStepWaitTimeMs, true, autoAdvanceDeploySteps, failedDeployTasks, System.currentTimeMillis());
```
tpetr: is `currentActiveInstances + 1` correct here? wouldn't it be `currentActiveInstances + deployInstanceCountPerStep`?

SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
tpetr: Can we implement force deploy in a separate PR?

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeployProgress.java
ssalinas: This was originally intended to just be the current deploy progress with `stepCompleted` flipped to true. If we want to update the active instance count I'd suggest getting it from the `deployActiveTasks` variable that gets passed around instead. If we pass that to the calls of `markStepFinished`, we can construct the new deploy progress with the actual count of active tasks after each step. `withCompletedStep` should do just what I mentioned above and flip the value of that bool, keeping the rest the same

darcatron:
@tpetr any changes necessary on this one?

tpetr:
Looks good but could you add unit tests for this? Or if there are already tests flexing this functionality, update them to ensure that `currentActiveInstances` is updated properly?

darcatron:
Just added a test following the logic of existing tests, but I can't say I'm super confident that it's the best way to test the counter. 

e.g. I couldn't figure out how to deploy with a target of 4 instances, and check the `currentActiveInstances` one at a time. I used the logic in another test that seemed to update the request each time before checking.

Thoughts? bitmoji

ssalinas:
@darcatron apologies for letting this one sit so long. I think having that current active instances value will still be valuable. In terms of the test, I would check out the example of `testDeployOneInstanceAtATime` in the same test class for how to do one step at a time of an incremental deploy. What that ones does is it:
- launches two initial tasks for wht will be the 'old' deploy
- starts a new deploy with incremental settings (1 instance at a time)
- updates the first task to be running and runs the deploy check to update the deploy progress (which would now have a current active of 1)
- makes sure the first old task gets killed
- does the same for the second task

You could likely extend that test to do three instances instead of just two and add a few more assert statements in there to adequately test the new counter

darcatron:
@ssalinas gave the test another look based on your suggestions and it seems that was the original test I followed. Instead of adding a check for these changes by extending an existing test, this one is separate. I think that's probably better anyway so unless you want to combine them, I'm going to keep it as is.

Cool if I get this into staging?

ssalinas:
I think that's fair, we can try this out in staging

PR #1379, opened 2016-12-16
[WIP] Force Deploy - allow a new deploy to override an existing deploy

darcatron:
- [ ] Need to make sure a deploy is fully cancelled before starting the new one


ssalinas:
Going to close this one for now. We'll come back to it in the future if we still need the functionality

PR #1381, opened 2016-12-16
remove all compressed log viewing code

ssalinas:
Goes with #1380 
/cc @tpetr


ssalinas:
this is in stable now, merging it into the tailer branch

PR #1382, opened 2016-12-19
Add maxTasksPerOffer at request level

ssalinas:
Allow a request to override the max tasks per offer so we can avoid scheduling a large stack of tasks on a single host
/cc @tpetr


tpetr:
Can we update `SingularityValidator` to ensure that `maxTasksPerOffer` is positive?

ssalinas:
Good, point @tpetr , added to the validator

ssalinas:
FYI, added a global setting for the request-level limit as well

PR #1383, opened 2016-12-19
Log start and end time attrs for s3 uploads

ssalinas:
One of the short comings of our current log searching is that we don't have defined start and end timestamps for when the log was written to. Because logrotate is run, we know when the log file was done being written to, and when it got uploaded, but not when it started being written to.

This PR uses linux extended custom file attributes to keep track of start and end times for log files when using log rotate. an attribute is set when a log file is initially being written to (at least for the default service log). When a log is rotated, that start time attribute is transferred to the rotated file (along with setting the end time), and the start time on the original file is reset to the current time.

This is still a POC, but worked pretty well in local testing.


SingularityExecutor/src/main/resources/logrotate.conf.hbs
```diff
@@ -14,15 +14,46 @@ notifempty
...
+    NOW="$(($(date +%s%N)/1000000))"
+    LOGSTART=`getfattr --only-values -n user.logstart {{{ logfile }}}`
+    if [ "$LOGSTART" != "" ]; then
+      for filename in "{{{ rotateDirectory }}}/{{{ logfileName }}}-$(date +{{{ rotateDateformat }}})"*
+      do
+        setfattr -n user.logend -v "$NOW" {{{ rotateDirectory }}}/$filename
+        setfattr -n user.logstart -v "$LOGSTART" {{{ rotateDirectory }}}/$filename
+      done
+    fi
+    setfattr -n user.logstart -v "$NOW" {{{ logfile }}}
```
darcatron: Since this is saving as the start time, does this mean that a log file that doesn't have the extended attribute hasn't been written to?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -678,6 +681,15 @@ public void setCronDirectory(String cronDirectory) {
     this.cronDirectory = cronDirectory;
   }
 
+  public boolean isUseFileAttributes() {
```
ssalinas: names like `isxxx()` and `getxxx()` are compliant with the JavaBeans spec for naming getter methods for booleans. That way, things that use reflection (like jackson) can find teh boolean getters. If we want it to read differently, we can renamed the field to something more compatible with `is<Field Name>`

SingularityExecutor/src/main/resources/logrotate.conf.hbs
ssalinas: Two things on that:
- I'm also setting this in the runner.sh script so a brand new log file would have it set.
- Right now in the S3Uploader I'm only adding the metadata if it is present on the file. The assumption would be that if we can't find a start time, we use the start time of the task itself, since it can be no earlier than that, and the same for end time. That's how the searching script currently works for all logs now

ssalinas:
The file attributes are working and making it to the s3uploader now. The uploader is adding the metadata to the S3Object before upload. For some reason, I am still not seeing any metadata being return, either in the endpoint, or when inspecting headers downloading manually. Still investigating here.

ssalinas:
This feature is turned off by default. Going to merge so there aren't so many horrid conflicts working on #1391

PR #1391, opened 2017-01-04
S3 Search Improvements

ssalinas:
Opening this early for any feedback.

Currently, all data about which files to upload to s3, and any possible bucket or key pattern overrides is stored in the executor configuration. This unfortunately means that SingularityService has no idea if it needs to search a separate bucket/pattern to find logs. It can only search what it knows about from it's defaults and the request group overrides.

Notable changes so far:
- s3 file and file name related configurations are moved from the executor to the service. This way it is available for searching and/or ui use more easily
- `loggingS3Bucket` removed from `ExecutorData` on the deploy. We have overrides by group and by file name for this purpose. If we want a request level override separate from the group one we can add that later
- The executor data passed to the custom executor is now the `SingularityTaskExecutorData` object not the `ExecutorData` from the deploy. This way we can pass additional information to the custom executor from system-wide configuration.
- `S3LogResource` will check the now available s3 additional files to determine if additional buckets need to be searched

Still TODO:
- [x] Optionally skip get/download url generation to improve efficiency of the logs endpoints
- [x] Generate proper prefixes for key pattern overrides
- [x] Add request group as possible param in the uploader key pattern
- [x] Paginate the s3 search endpoint
- [x] Docs for config changes

**Note** - SingularityService needs to be deployed _before_ the executor for this particular change

/cc @tpetr


ssalinas:
FYI, now building this on top of the commits from #1375 . It modifies/moves a bunch of the same files and the merges are a bit horrid otherwise

ssalinas:
Added the docs for this now, going to merge

PR #1394, opened 2017-01-09
jets3t -> aws sdk

ssalinas:
Moving us to the aws sdk. This is partly just to keep up to date, but also so that we can use the new `GetObjectV2` to do pagination when listing buckets in the future.

`aws.sdk.version` is bumped to `1.10.77` for now. `1.10.x` is the first to have the new v2 request (basepom has `1.9.x`). Going all the way to `1.11.x` created too many dependency issues.

PRing this into the other S3 rework  branch.

/cc @tpetr

### DIFF
--- a/SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/config/SingularityS3Configuration.java
+++ b/SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/config/SingularityS3Configuration.java
@@ -45,7 +45,7 @@ public class SingularityS3Configuration extends BaseRunnerConfiguration {
 
   @Min(1)
   @JsonProperty
-  private long s3ChunkDownloadTimeoutMillis = TimeUnit.SECONDS.toMillis(30);
+  private int s3ChunkDownloadTimeoutMillis = (int) TimeUnit.SECONDS.toMillis(30);
 
   @Min(1)
   @JsonProperty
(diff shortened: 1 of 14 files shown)


SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/config/SingularityS3Configuration.java
```diff
@@ -131,7 +131,7 @@ public void setLocalDownloadPath(String localDownloadPath) {
     this.localDownloadPath = localDownloadPath;
   }
 
-  public long getS3ChunkDownloadTimeoutMillis() {
+  public Long getS3ChunkDownloadTimeoutMillis() {
```
tpetr: why `Long` here?

ssalinas: in order to more cleanly call `intValue()` (aws sdk takes an int). Was debating if it was cleaner to keep the `TimeUnit.SECONDS.toMillis(30)`so the default is readable and have to do `long` -> `int`, or to move it all to int. Reading it now the second probably makes more sense, I'll update

tpetr: I'd say let's change this to `int`. The max value of a Java integer in milliseconds is a little more than 24 days, which should be more then enough for this timeout.

ssalinas: updated

tpetr:
LGTM

ssalinas:
This is working fine in staging, going to merge it with the other s3-related PR

PR #1395, opened 2017-01-09
Implement pagination for S3 endpoint using continuation tokens

ssalinas:
This adds additional endpoints for searching s3 logs. This first attempt at pagination utilizes S3's continuation tokens. These tokens are specific to the bucket + prefix being search. So, this pagination would work via the api returning results as well as a list of tokens and their associated prefix/bucket that could be returned to retrieve a subsequent page.

I am not sold that this is the best option, but it is the one that limits the number of calls we make to s3 the best.

An additional option would be to list out all of the object details for everything, but only 'hydrate' (download/get url + extra metadata) those on the current page. This would allow simpler parameters (page + page size) and allow us a total count, but would require us listing all objects for all buckets + prefixes on each new page request.

thoughts @tpetr @darcatron ?

*note* - this pr is also built on top of the existing s3_rework branch

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/api/ContinuationToken.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/api/ContinuationToken.java
@@ -1,0 +1,58 @@
+package com.hubspot.singularity.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.wordnik.swagger.annotations.ApiModelProperty;
+
+public class ContinuationToken {
+  private final String value;
+  private final boolean lastPage;
+
+  @JsonCreator
+  public ContinuationToken(@JsonProperty("value") String value, @JsonProperty("lastPage") boolean lastPage) {
+    this.value = value;
+    this.lastPage = lastPage;
+  }
+
+  @ApiModelProperty(required=true, value="S3 continuation token specific to a bucket + prefix being searched")
+  public String getValue() {
+    return value;
+  }
+
+  @ApiModelProperty(required=true, value="If true, there are no further results for this bucket + prefix")
+  public boolean isLastPage() {
+    return lastPage;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    ContinuationToken token = (ContinuationToken) o;
+
+    if (lastPage != token.lastPage) {
+      return false;
+    }
+    return value != null ? value.equals(token.value) : token.value == null;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = value != null ? value.hashCode() : 0;
+    result = 31 * result + (lastPage ? 1 : 0);
+    return result;
+  }
+
+  @Override
+  public String toString() {
+    return "ContinuationToken{" +
+        "value='" + value + '\'' +
+        ", lastPage=" + lastPage +
+        '}';
+  }
+}
--- a/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityS3SearchRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityS3SearchRequest.java
@@ -1,0 +1,100 @@
+package com.hubspot.singularity.api;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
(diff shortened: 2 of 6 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/api/ContinuationToken.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.wordnik.swagger.annotations.ApiModelProperty;
+
+public class ContinuationToken {
+  String bucket;
+  String prefix;
+  String value;
+  boolean lastPage;
```
tpetr: `private final` ?

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityS3SearchRequest.java
```diff
@@ -0,0 +1,77 @@
...
+  public SingularityS3SearchRequest(@JsonProperty("start") Optional<Long> start,
+                                    @JsonProperty("end") Optional<Long> end,
+                                    @JsonProperty("excludeMetadata") Optional<Boolean> excludeMetadata,
+                                    @JsonProperty("listOnly") Optional<Boolean> listOnly,
+                                    @JsonProperty("maxPerPage") Optional<Integer> maxPerPage,
+                                    @JsonProperty("continuationTokens") Set<ContinuationToken> continuationTokens) {
+    this.start = start;
+    this.end = end;
+    this.excludeMetadata = excludeMetadata.or(false);
+    this.listOnly = listOnly.or(false);
```
tpetr: if you're doing `.or(false)` it'd be cleaner to just have the type be `boolean`

SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java
```diff
@@ -323,23 +348,41 @@ public SingularityS3LogMetadata call() throws Exception {
...
+      return new SingularityS3SearchResult(continuationTokens, isFinalPageForAllPrefixes(continuationTokens), logs);
     } finally {
       executorService.shutdownNow();
     }
   }
 
+  private boolean isFinalPageForAllPrefixes(Set<ContinuationToken> continuationTokens) {
+    boolean finalPage = true;
+    for (ContinuationToken token : continuationTokens) {
+      finalPage = finalPage && token.isLastPage();
```
tpetr:
```java
if (!token.isLastPage()) {
    return false
}
```
would be a little more efficient

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityS3SearchRequest.java
```diff
@@ -0,0 +1,77 @@
...
+    this.maxPerPage = maxPerPage;
+    this.continuationTokens = Objects.firstNonNull(continuationTokens, Collections.<ContinuationToken>emptySet());
```
tpetr: I like the idea of the continuation tokens but having to look up in the set seems a little janky to me -- could we map bucket + prefix to the token as a `Map<String, ContinuationToken>` ?

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityS3SearchRequest.java
ssalinas: Forgot that the null will default to a false with jackson. Had it in my head that it would error. Updating 👍

ssalinas:
@tpetr got this a bit closer now. It's kind of a trade off of concurrency and fast calls vs respecting the `perPage` count. I could probably get it to exactly the right amount, but it would require making a new call with a different limit set for whatever the last bucket/prefix was to add data (so that we have a valid continuation token). Seemed silly to be dumping data we already fetched so I didn't do that yet.

Open to any comments :)

ssalinas:
@tpetr got this to a final(ish) state now.

The `/search` endpoint takes an object that can have a list of tasks/requests+deploys to search, and will find all needed prefixes, then begin getting pages of data. This way we can more easily search across multiple things without having to make an api call for each.

The page size is the -ish part of this still. In order to keep some concurrency/speed, I an utilizing `AtomicInteger`'s `compareAndSet`. However, the last chunk of items to get added to the results may push the total count over the max page size. We can leave any out because the next continuation token is then not valid. We _could_ re-request with a different `maxKeys`, but it seems silly to re-request data we already have.

ssalinas:
With most recent changes this is working well in staging/qa. Going to merge it with the other s3 PR

PR #1396, opened 2017-01-10
[logfetch] Use the new paginated s3 search endpoint

ssalinas:
Goes with #1395


ssalinas:
Holding off in favor of #1408 , closing this for the moment

PR #1397, opened 2017-01-12
Shade com.google.thirdparty

stevegutz:
Guava 16 introduced packages starting with `com.google.thirdparty` which need to be shaded as well. I am augmenting all current shadings of Guava regardless of version to prevent future issues.

@ssalinas @jhaber


ssalinas:
Thanks @stevegutz LGTM

PR #1399, opened 2017-01-17
Add ability to upload files immediately to S3

PtrTeixeira:
Change has two parts: Adds an optional uploadImmediately field to the uploader
metadata, that when present and set will cause the uploader to upload it
immediately on change, and code within the on-detection function
`handleNewOrModifiedS3Metadata` to perform the upload if that field is
detected.

This change in particular involved moving the vast majority of the code in
`checkUploads` into a seperate function.  I wanted to avoid duplicating the
logic in `checkUploads` in `handleNewOrModifiedS3Metadata`. If I directly
called `checkUploads`, however, that would upload all the files that were
queued to upload, rather than just the files marked to upload immediately. So I
made a new function that would just take a map of files to upload and upload
only those files. So when it needed to upload the all queued files, as in
`checkUploads`, it could directly work on the `metadataToUploader` field, and
when it only needs to upload a single block of files, it can only be passed
that single uploader/metadata.


ssalinas:
Commenting here to document a chat we had in-person. Decided to go a slightly different route and have the immediate uploads submit their own future to the executorService, rather than trying to hook something immediate into the runs of the scheduledExecutorService. @PtrTeixeira is making updates and we'll review again once those are up

PtrTeixeira:
cc @ssalinas Could you take a look?

ssalinas:
@PtrTeixeira looks good overall. For ease of debugging when you try this out, I'd add some info or debug logging near the beginning of `performImmediateUpload`. Would give us a better idea that everything is being triggered appropriately rather than just waiting to see if things uploaded or waiting for the callback to trigger.

ssalinas:
FYI @PtrTeixeira I merged #1391 ,should be able to resolve the merge conflicts on this one now

PR #1402, opened 2017-01-19
introduce offer cache to allow better resource allocation

wsorenson:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityOfferCache.java
```diff
@@ -0,0 +1,160 @@
...
+
+    public CachedOffer(Offer offer) {
+      this.offerId = offer.getId().getValue();
+      this.offer = offer;
+      this.offerState = OfferState.AVAILABLE;
+    }
+
+    private void checkOut() {
+      LOG.trace("Checking out offer {}", offerId);
+      Preconditions.checkState(offerState == OfferState.AVAILABLE, "Offer %s was in state %s", offerId, offerState);
```
ssalinas: Saw a sentry come through for this the other day. This exception will bubble all the way back up to the LeaderOnlyPoller and cause us to abort. Should we maybe catch this and just skip the offer in the case that something else has it checked out?

SingularityService/src/main/java/com/hubspot/singularity/mesos/OfferCache.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public void rescindOffer(SchedulerDriver driver, OfferID offerId);
+
+  public void useOffer(CachedOffer cachedOffer);
+
+  public List<CachedOffer> checkoutOffers();
+
+  public void returnOffer(CachedOffer cachedOffer);
+
+  public List<Offer> peakOffers();
```
tpetr: `peekOffers()`?

ssalinas:
Fixed the merge conflicts and re-tested in each environment as well, going to merge this one 👍

miry:
@wsorenson how to disable this this feature?

tpetr:
@miry It looks like the documentation needs to be updated, thanks for the heads up. You can disable offer caching by setting `cacheOffers: false` in your Singularity config YAML.

@ssalinas @wsorenson thoughts on having offer caching disabled by default?

ssalinas:
For the sake of those running in multi-framework clusters it might make sense to have it off by default. I can submit a PR with that and the documentation update

PR #1403, opened 2017-01-19
Import formModal on racks page

darcatron:
`/racks` was busted

@ssalinas


ssalinas:
Thanks @darcatron , LGTM

PR #1405, opened 2017-01-24
Slave Usage Monitoring UI

darcatron:
This is the front-end for the slave usage tracking. Definitely a work in progress, but the general structure is there. Feel free to edit the task lists.

Design Tasks:
- [x] slave button styling
- [x] slave stat menu styling
- [x] aggregate page styling: do we want aggregate info/what do we want here?

Functionality Tasks:
- [ ] what are the thresholds and where are they defined
- [x] over subscription issues
- [x] timestamp display locale
- [x] clicking on a slave stat copies to clipboard
- [ ] how to show dead slaves?
- [ ] show/sort by rack info?
- [ ] filtering/sorting options

Updated - 2/21/2017
image

cc: @ssalinas @wsorenson @tpetr
BE: https://github.com/HubSpot/Singularity/pull/1400


ssalinas:
A few comments based on your task list above:
- There is a method in utils for converting those timestamps to a more human-readble format
- Not sure what you categorize as disabled slaves, but it might be good to separate out at least the 'Dead' ones like on the slaves page
- Would be interesting to be able to aggregate by rack (i.e. one big box of summed up metrics for each rack)
- would be nice to be able to search/filter for a slave in some way. Not sure what the best way for that would be since we don't want those boxes to be text-heavy at all
- For the over-subscription piece, we should add extra attributes that Singularity will look for on the slave. If those attributes are present, it will use that as the 'total' for mem/cpu/etc. If they are not present it will use the values reported by the slave. This will need to be a change on the backend PR

Will look more at the actual js and leave comments there shortly

darcatron:
The actual js has a lot of room for improvement. Modularization is one I'm working on now. I think that review is better saved for once the design is decided since it can change a lot before that.

wsorenson:
cool!

We need to merge the data with the slave APIs so that they should hostname and cpusUsed vs cpusAllocated (as opposed to the raw no.)

Maybe that was obvious / not implemented yet, wanted to make sure!

ssalinas:
Yep we talked about using the attribute values (falling back to the resource ones) as the denominator for getting the usage percentages already

And yes @darcatron , would be nice if we sub that slave ID in with the hostname, would be much easier to reason about/read in the ui that way

SingularityUI/app/actions/api/slaves.es6
```diff
@@ -2,14 +2,14 @@ import { buildApiAction, buildJsonApiAction } from './base';
 
 export const FetchSlaves = buildApiAction(
   'FETCH_SLAVES',
-  {url: '/slaves'}
```
ssalinas: why did you remove all of the leading slashes here?

SingularityUI/app/components/machines/SlaveUsage.jsx
```diff
@@ -0,0 +1,211 @@
...
+
+  const drawSlave = (slave, index) => {
+    var criticalGlyph = 'glyphicon glyphicon-remove-sign';
+    var warningGlyph = 'glyphicon glyphicon-minus-sign';
+    var okGlyph = 'glyphicon glyphicon-ok-sign';
+    var criticalStyle = 'danger';
+    var warningStyle = 'warning';
+    var okStyle = 'success';
+
+    if (isStatCritical(slave, props.cpusUsedStat) || isStatCritical(slave, props.memoryBytesUsedStat) || isStatCritical(slave, props.numTasksStat)) {
```
ssalinas: we should probably catch the exceptions you are throwing in the case where we can't properly find the stats. It'd be better to show less stats (excluding the ones that threw an exception), rather than throwing js errors all the way up

SingularityUI/app/components/machines/SlaveUsage.jsx
```diff
@@ -0,0 +1,211 @@
...
+        } catch (e) {
+          throw 'Could not find resources (memory) for slave ' + slave.slaveId;
+        }
+      default:
+        throw statName + ' is an unsupported statistic';
+    }
+  };
+
+  const slaveWithStats = (slave, index, bsStyle, glyphicon) => (
+    <Dropdown key={slave.slaveId} id={index.toString()}>
```
ssalinas: Any reasoning behind using dropdown here? Seems like an odd component choice

SingularityUI/app/actions/api/slaves.es6
darcatron: the `buildApiAction` already has `/api/` so it's just an extra `/`. `../api//slaves` will work, but it's not proper so I figured i'd remove it. I can drop it from the rest of the files too, but prob better for another pr

SingularityUI/app/components/machines/SlaveUsage.jsx
darcatron: The dropdown is the little stat menu you get when clicking on a slave btn. Using Dropdown over DropdownMenu allows more styling flexibility

SingularityUI/app/rootComponent.jsx
```diff
@@ -36,6 +36,7 @@ const rootComponent = (Wrapped, title, refresh = _.noop, refreshInterval = true,
       }).catch((reason) => {
         // Boot React errors out of the promise so they can be picked up by Sentry
         setTimeout(() => {
+          console.log(reason.stack);
```
ssalinas: don't forget to remove these after you're done testing ;)

darcatron: I'm never done testing 😈

SingularityUI/app/actions/api/slaves.es6
darcatron: turns out this is handled in staging so the `buildApiAction` doesn't return the extra slash  bitmoji

darcatron:
Got some changes in to mimic the mockup. The leftovers are really just style changes which can be adjusted fairly easily.

p.s. i'll take care of the accidental merge from stagging

SingularityUI/app/components/machines/SlaveUsage.jsx
```diff
@@ -0,0 +1,62 @@
...
+  const slaveHealthData = slaveUsages.map((slaveUsage, index) => {
+    const slaveInfo = getSlaveInfo(slaves, slaveUsage);
+    return <SlaveHealth key={index} slaveUsage={slaveUsage} slaveInfo={slaveInfo} />;
+  });
+
+  return (
+    <div id="slave-usage-page">
+      <h1>Slave Usage</h1>
+      <div>
+        <SlaveAggregates totalSlaves={20} totalTasks={200} avgCpu={30} avgMemory={40} />
```
darcatron: realized these are still hard coded. will try to get a fix out for this soon

SingularityUI/app/components/machines/Constants.jsx
```diff
@@ -0,0 +1,38 @@
+export const HUNDREDTHS_PLACE = 2;
+
+export const SLAVE_STYLES = {
+  ok : 'ok-slave',
+  warning : 'warning-slave',
+  critical : 'critical-slave'
+};
+
+export const THRESHOLDS = {
```
ssalinas: Maybe we can load these in from window.config instead?

SingularityUI/app/components/machines/SlaveUsage.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+const refresh = () => (dispatch) =>
+  Promise.all([
+    dispatch(FetchSlaves.trigger()),
+    dispatch(FetchSlaveUsages.trigger()),
+    dispatch(FetchSingularityStatus.trigger())
+  ]);
+
+const initialize = () => (dispatch) =>
+  Promise.all([]).then(() => dispatch(refresh()));
```
ssalinas: Why an empty Promise.all here?

SingularityUI/app/utils.es6
```diff
@@ -45,6 +46,10 @@ const Utils = {
     ));
   },
 
+  humanizeSlaveHostName(longHostName) {
```
ssalinas: I'll check out how this looks live in the ui, but this might be a bit specific to how hubspot does naming. Most are probably going to want to see the full hostname

SingularityUI/package.json
```diff
@@ -89,6 +91,12 @@
     "gulp-util": "^3.0.7",
     "mocha": "^3.1.2",
     "nib": "^1.1.0",
+    "node-sass": "^3.8.0",
+    "react-copy-to-clipboard": "^4.2.3",
+    "react-circular-progressbar": "^0.1.3",
+    "react-hot-loader": "^3.0.0-beta.2",
```
ssalinas: are these all in use here, or were they part of the merge conflicts with hs_staging?

SingularityUI/app/components/machines/SlaveUsage.jsx
darcatron: mostly due to my lack of understanding of which pieces are necessary. Looking at the rootComponent code, it looks like not having an initialize will just default to the refresh which is fine so I'll remove this function

SingularityUI/app/utils.es6
darcatron: if we need to, we can drop it into a boolean config property then

SingularityUI/package.json
darcatron: react-copy-to-clipboard and react-circular-progressbar are the new ones. The others are from hs_staging

SingularityUI/app/components/machines/SlaveUsage.jsx
ssalinas: 👍

darcatron:
@ssalinas got the updates in for the default values. Once you think it's correct for staging, i'll set our config for shortening the slave name to `true` and deploy it

SingularityService/src/main/java/com/hubspot/singularity/views/IndexView.java
```diff
@@ -239,6 +239,10 @@ public String getRedirectOnUnauthorizedUrl() {
     return redirectOnUnauthorizedUrl;
   }
 
+  public boolean isShortenSlaveUsageHostname() {
```
ssalinas: FYI @darcatron , without the getter here, this value is not available when rendering the template, so `{{ shortenSlaveUsageHostname }}` would come back empty in the rendered template

demobox:
Just related to this, or perhaps for potential inspiration: we are currently doing similar monitoring of our Mesos agent usage and utilization via https://github.com/Capgemini/mesos-ui/https://youtu.be/bKbOod8Pn4E

PR #1407, opened 2017-01-26
Forbid more characters from request/deploy IDs

### LINKED ISSUE #1406: Deploy id character blacklist needs expanding
Experimenting with Deploy ids and redeploys (specifically, I was looking for a delimiter to use in machine-generated deploy ids), I found that a Deploy id that Singulary's API accepted still crashed the Executor:

Failed to launch container: Failed to create container: Failed to chown: Failed to execute 'chown -R 0:0 '/mnt/mesos-slave/slaves/f38ff857-4831-46dc-b143-d8aaa0c44d06-S0/frameworks/Singularity/executors/github.com>opentable>sous-demo::ci-sf-<>|=()'_+{}"^&df1d7456ac654e4097ae451b0dc6c0fa-1485370286019-1-[REDACTED]-FIXME/runs/9967afcc-0feb-4421-baa7-094d6d8896fb'' (exit status: 512): Success; Abnormal executor termination

I'm concerned further that one explanation for this problem is that the `'` ends the chown command, and the shell treats the rest as garbage. (I can only guess here: it *looks* like exec(commandString) rather than exec([array, of, command, parts])). If that's the case, there's the possibility of an RCE against Mesos agents for anyone with access to Singularity. It's tricky to test for this, because the ` ` restriction limits us to single word commands... like `halt`.

PtrTeixeira:
In an effort to prevent attempting to create docker containers
with names that docker will refuse, this forbids more characters
from being used in request/deploy IDs. This is actually more of a
whitelist approach than a blacklist approach; it explicitly
says the characters that can be used and forbids anything else.

Closes #1406 
/cc @ssalinas


SingularityService/src/test/java/com/hubspot/singularity/data/SingularityValidatorTest.java
```diff
@@ -0,0 +1,63 @@
+package com.hubspot.singularity.data;
+
+import static org.mockito.Mockito.*;
```
ssalinas: We generally try to stay away from any import `*` if we can. Can probably tune your IDE settings for this

SingularityService/src/test/java/com/hubspot/singularity/data/SingularityValidatorTest.java
```diff
@@ -0,0 +1,63 @@
...
+import com.hubspot.singularity.config.MesosConfiguration;
+import com.hubspot.singularity.config.SingularityConfiguration;
+
+@RunWith(JukitoRunner.class)
+public class SingularityValidatorTest {
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  public static class TestModule extends JukitoModule {
```
ssalinas: We actually have a class set up that handles most of this setup already. If you just add these tests in the `ValidatorTest` class (feel free to rename to `SingularityValidatorTest` for consistency), that class already has all of the guice bits taken care of and you can use the already injected `SingularityValidator`. Saves us from adding the mockito dep for just this class as well ;)

SingularityService/src/test/java/com/hubspot/singularity/data/ValidatorTest.java
```diff
@@ -28,4 +48,100 @@ public void testCronExpressionHandlesDayIndexing() {
     Assert.assertEquals("0 0 12 ? * MON-WED", validator.getQuartzScheduleFromCronSchedule("0 12 * * MON-WED"));
   }
 
+  @Test(expected = WebApplicationException.class)
+  public void itForbidsBracketCharactersInDeployIds() throws Exception {
+    final String badDeployId = "deployKey[[";
+
+    SingularityDeploy singularityDeploy = new SingularityDeploy(badDeployId, badDeployId,
```
ssalinas: There are `SingularityDeployBuilder` and `SingularityRequestBuilder` methods for this so that you don't need to use the giant constructor. For example you could do:
`SingularityDeploy.newBuilder("requestId", "deployId")`

Fix that up and I think this PR is all set 👍

ssalinas:
LGTM

PR #1409, opened 2017-01-27
Destroy task

darcatron:
This fixes an issue with how the form elements were mapped. If a prior form field isn't used in an updated form, it would throw an undefined error when it tried to access the non-existent field. I tried finding a good way to remove the field without disrupting the rest of the uses of the FormModal component, but I could not find one. Right now it just removes as unused fields from the mapping. If you see a better method, I could def use it

@ssalinas

### DIFF
--- a/SingularityUI/app/components/common/modal/FormModal.jsx
+++ b/SingularityUI/app/components/common/modal/FormModal.jsx
@@ -84,7 +84,7 @@ export default class FormModal extends React.Component {
     this.setState({ formState });
   }
 
-  validateForm() {
+  isValidForm() {
     // Check required values
     const errors = {};
     this.props.formElements.forEach((formElement) => {
@@ -141,11 +141,11 @@ export default class FormModal extends React.Component {
     if (event) {
       event.preventDefault();
     }
-    if (this.validateForm()) {
+    if (this.isValidForm()) {
       let formState = this.parseFormState(this.state.formState);
       this.props.onConfirm(formState);
       if (!this.props.keepCurrentFormState) {
-        const formState = {};
+        formState = {};
         this.props.formElements.forEach((formElement) => {
           formState[formElement.name] = formElement.defaultValue;
         });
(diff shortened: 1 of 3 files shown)


SingularityUI/app/components/common/modal/FormModal.jsx
```diff
@@ -119,20 +114,24 @@ export default class FormModal extends React.Component {
...
-          break;
-        case FormModal.INPUT_TYPES.DURATION:
-          if (val) {
-            parsed[key] = juration.parse(val) * 1000;
-          }
-          break;
-        default:
-          parsed[key] = val;
+      if (element === undefined) {
+        delete this.state.formState[key];
```
ssalinas: we don't want to directly manipulate the state here. Instead we need to eventually pass the new data to `setState`. We use the return value from this method to set the state correctly on line 153. So, I think if the element is undefined we can just skip that element and move on. It will then not end up in `parsed` and will be later removed from the state when we do `setState`

SingularityUI/app/components/common/modal/FormModal.jsx
```diff
@@ -142,11 +141,11 @@ export default class FormModal extends React.Component {
     if (event) {
       event.preventDefault();
     }
-    if (this.validateForm()) {
+    if (this.isValidForm()) {
       let formState = this.parseFormState(this.state.formState);
       this.props.onConfirm(formState);
       if (!this.props.keepCurrentFormState) {
-        const formState = {};
+        formState = {};
```
darcatron: removed the const because the variable is block scoped and so it would never be saved properly here

ssalinas: yep, saw that after and deleted my comment, didn't have the diff open wide enough

darcatron:
Updated to ignore unused fields. I tried removing the state entirely and getting it working with just props, but it didn't work easily. It seems possible though. We should re-visit it at another time.

ssalinas:
Looking good on stable, LGTM

PR #1410, opened 2017-02-02
use an ldap cache

wsorenson:


SingularityService/src/main/java/com/hubspot/singularity/auth/datastore/SingularityLDAPDatastore.java
```diff
@@ -143,9 +170,16 @@ public SingularityLDAPDatastore(SingularityConfiguration configuration,
...
+              Optional.fromNullable(Strings.emptyToNull(userEntry.get(configuration.getUserEmailAttribute()).getString())), groups));
+
+          if (ldapCache.isPresent()) {
+            ldapCache.get().put(user, result);
+          }
+
+          return result;
         } finally {
-          LOG.info("Loaded {}'s user data in {}ms", user, System.currentTimeMillis() - startTime);
+          LOG.trace("Loaded {}'s user data in {}", user, JavaUtils.duration(startTime));
```
ssalinas: 💯

ssalinas:
Will be good to have this, especially with the number of calls that things like the log tailer make

PR #1411, opened 2017-02-02
Clean up some logging

ssalinas:
- Strip the new lines and repeated spaces out of the toString for the mesos protos objects
- Add shorter but helpful debug messages where we are now logging entire task objects

@wsorenson wasn't a quick-and-easy way to write a template to not log absent optionals in all the toString methods. But this should still clean things up a good bit. Will add to it a bit more tomorrow.


darcatron:
:ship:

PR #1414, opened 2017-02-06
Add style of change display of overflow text

PtrTeixeira:
Add a style to change the display of the overflow text. This will break
between words if possible, and if not possible will force a line wrap
when it hits the maximum with.

/cc @ssalinas


ssalinas:
will be nice to not have the command line args go off the screen anymore 💃

markhazlewood:
@PtrTeixeira Could we see an example of the new wrapping behavior? In general I'd recommend against having dropdown items take up multiple lines and instead use an `ellipsis w/ tooltip` overflow pattern. But would like to see this in context.

PtrTeixeira:
This is what I had part way through the working on this:
screen shot 2017-02-06 at 12 50 01 pm

I didn't think of using a tooltip to get the full content for the menu items. Will move over to that.

ssalinas:
👍 thanks for the suggestion mark

ssalinas:
@PtrTeixeira on staging right now it seems that additional arguments are not showing up in the history. A previous list of args with > 1 arg seems to only show the first one.

PR #1415, opened 2017-02-06
Replace existing global search with search from requests page

PtrTeixeira:
Change from directly using the fuzzy search API provided by `fuzzy` to
the existing custom one used in the search bar on the Requests page.
This did preserve using the existing `react-typeahead` element, rather
than switching to the `react-select` element. This should make the 
global search a little less helter-skelter in what it matches.

One thing that we lost in this switchover: highlighting of matched
characters. The search on the requests page doesn't use or emit that
data, so it no longer comes into the global search. We also aren't as
yet using the full potential of the global search - for example, you
can't search by the deploy user, and you can't toggle searching by
status.

/cc @ssalinas

### DIFF
--- a/SingularityUI/app/components/globalSearch/GlobalSearch.jsx
+++ b/SingularityUI/app/components/globalSearch/GlobalSearch.jsx
@@ -8,6 +8,7 @@ import { SetVisibility } from '../../actions/ui/globalSearch';
 import { Typeahead } from 'react-typeahead';
 import fuzzy from 'fuzzy';
 import key from 'keymaster';
+import filterSelector from '../../selectors/requests/filterSelector';
 
 class GlobalSearch extends React.Component {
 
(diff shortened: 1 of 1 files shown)


darcatron:
Just a question. I saw that this updated search has a sub filter for task types ( SCHEDULED, ON_DEMAND, etc.). Will this be a possible global search query or is it part of the unused potential you mentioned?

PtrTeixeira:
At the moment, it's unused; those filters are used on the requests page search, but not on the global search.

ssalinas:
@PtrTeixeira , The `t` search doesn't have the full range of state/type filters, I don't think we need to specify that field at all here. `state: 'all'` and the search input should be enough

PtrTeixeira:
`filterSelector` does a [filter](https://github.com/HubSpot/Singularity/blob/master/SingularityUI/app/selectors/requests/filterSelector.es6#L29) by the contents of `subFilter` against `requestType`; if `subFilter` is empty or `undefined`, it will remove everything from what it returns.

ssalinas:
My mistake, I had thought that `subFilter` not present just skipped that filter. Let's give this a try in staging

PR #1417, opened 2017-02-07
Ability to specify cache on mesos artifact

### LINKED ISSUE #1416: Artifacts cache
It seems, that currently the artifacts (specified in `uris` field of a SingularityDeploy) cannot be marked as cacheable by the Mesos fetcher: described in http://mesos.apache.org/documentation/latest/fetcher  (chapter Fetching through the cache)

Are there any plans of implementing this? Or I am able to help with this feature (as we are interested in this functionality) - what would be the best way of doing it? Encapsulating the `uris Array[string]` into an object `uris Array[ArtifactItem]` with fields `uri, executable, extract, cache` ?

ssalinas:
/fixes #1416

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/mesos/SingularityMesosArtifact.java
+++ b/SingularityBase/src/main/java/com/hubspot/mesos/SingularityMesosArtifact.java
@@ -1,0 +1,85 @@
+package com.hubspot.mesos;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SingularityMesosArtifact {
+  private final String uri;
+  private final boolean cache;
+  private final boolean executable;
+  private final boolean extract;
+
+  @JsonCreator
+  public static SingularityMesosArtifact fromString(String uri) {
+    return new SingularityMesosArtifact(uri, false, false, false);
+  }
+
+  @JsonCreator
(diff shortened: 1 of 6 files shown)


bydga:
Hi again, I've just tried the cache from self-compiled singularity (current master). And when I did a deployment with the original approach (no caching, just specifying the artifacts as strings)
```
uris: ["http://some.url/resource.tgz", "s3n://some-other-resource-from-s3.tgz"]
```
None of the resources was unpacked in the sandbox directory - which Mesos does by default when not specifying the `cache` field in the protobuf.

I think the behavior should be the same - see: http://mesos.apache.org/documentation/latest/fetcher/
```
message CommandInfo {
  message URI {
    required string value = 1;
    optional bool executable = 2;
    optional bool extract = 3 [default = true];
    optional bool cache = 4;
    optional string filename = 5;
  }
}```

When I did the deploy with the object specification
```uris: [{uri: "http://resource.tgz", extract: true} ]```

everything worked just fine

ssalinas:
Ah, I had been defaulting it to false for all of those parameters, we can update that

bydga:
Well, since you are breaking previous behaviour - which might cause all tasks that rely on an archive unpacking to fail, plus its against the mesos-default behavior, I would suggest to change the `extract` flag to true :)

ssalinas:
Yep, like I said, we can update. I'll get a PR in before we do the next release

PR #1418, opened 2017-02-09
Fixes for Immediate Uploaders

ssalinas:
@PtrTeixeira after looking at the code for this for a while I realized a few things:
- It's quite messy to deal with the single uploader metric counter when the immediate vs regular uploaders are treated so differently
- We weren't actually able to guarantee that the files eventually got uploaded in the exception case. i.e. if there is an exception from S3, we probably need to retry that still.

So, I updated this to have a map to keep track of the immediate uploaders and have a separate metric counter for them as well. We submit the callables to the executor service immediately, but rather than working with the callbacks, we check the results on our regular polling interval, restarting them if there are exceptions (only retrying for exception, not for no files found right now)

This probably still needs a few tweaks, can you take it from here on this one?


PtrTeixeira:
👀

PR #1423, opened 2017-02-10
Update docker parameters on deploy form

### LINKED ISSUE #1420: Deploy UI ignores Docker parameters entered
I'm running Singularity 0.9.0. When I set the Docker parameters in a deploy, like that shown in this screenshot:

screen shot 2017-02-09 at 1 28 53 pm

 they don't show up in the JSON of the deploy, when I view the deploy JSON via the UI, which looks like this:

```json
"deploy": {
        "requestId": "generic-launcher-svc-test-app-1",
        "id": "0209.7",
        "command": "",
        "containerInfo": {
            "type": "DOCKER",
            "docker": {
                "image": "pdocker.valkyrie.net/odc/generic-launcher-svc-test-app:1.1.2",
                "privileged": false,
                "network": "HOST",
                "portMappings": [],
                "forcePullImage": false,
                "parameters": {}
            }
        },
```

I expect the parameters section to include the parameters I entered. The hope is that this will turn into a docker run command like this:

```
docker run --user 616573731 <image-hash>
```
I get no error in creating the deployment and they are not used when starting up the Docker container. I was hoping they were used like parameters are used in a Docker container deploy via Marathon.

The API for this functionality is a bit confusing as well. It has dockerParameters and parameters. For Marathon the parameters need to be formatted as a list of JSON objects like this:

```
{ "key": "user", "value:, "616573731" }
```

Perhaps I need to enter the '--' with my parameter name? Perhaps I need to add them as key=user and value=616573731, but that wouldn't make much sense to require that in the UI...

Not sure if this is related to my problem, but whenever I visit the page for this deploy I get this error popping up:

screen shot 2017-02-09 at 1 38 31 pm

ssalinas:
/fixes #1420

/cc @billwright

### DIFF
--- a/SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
+++ b/SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
@@ -149,7 +149,6 @@ class NewDeployForm extends Component {
   }
 
   getValueOrDefault(fieldId) {
-    console.log(INDEXED_FIELDS)
     return this.props.form[fieldId] || INDEXED_FIELDS[fieldId].default;
   }
 
@@ -329,8 +328,6 @@ class NewDeployForm extends Component {
 
   copyFieldsToObject(deployObject, fieldsToAdd, valueGetter) {
     for (const fieldId of fieldsToAdd) {
-      console.log(fieldId)
-      console.log(fieldsToAdd)
       if (fieldId.type === 'object') {
         deployObject[fieldId.id] = this.copyFieldsToObject(
           deployObject[fieldId.id] || {},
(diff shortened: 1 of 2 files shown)


SingularityUI/app/components/newDeployForm/NewDeployForm.jsx
```diff
@@ -377,6 +385,14 @@ class NewDeployForm extends Component {
             DOCKER_PORT_MAPPING_FIELDS,
             (id) => portMapping[id] || INDEXED_DOCKER_PORT_MAPPING_FIELDS[id].default));
           deployObject[fieldId.id] = portMappings;
+        } else if (fieldId.type === 'dockerParameters') {
+          console.log(fieldId)
+          debugger
```
billwright: What's this line do? debugger? 
Ignore if this is obvious to JS programmers. New language for me.

PR #1424, opened 2017-02-14
Rename metric for immediate uploaders

PtrTeixeira:
Rename the metric for immediate uploaders. I _think_ that if you ask the `MetricsRegistry` for another counter with the same name, it will just give you back a reference to the original counter.


ssalinas:
Good catch @PtrTeixeira LGTM

ssalinas:
Looks fine in qa, let's move this to stable and merge it into the other PR branch @PtrTeixeira

PR #1428, opened 2017-02-22
Prevent flapping slave from rejoining cluster

PtrTeixeira:
This adds a node to zk (`/singularity/inactive`) that keeps just
contains an array of hosts whose slaves have been marked as inactive.
When Singularity checks if an offer should be accepted, it grabs the
node from zk. If the offer is from a slave on a bad host, Singularity
will discard it.

/cc @ssalinas Is the right way to go about this?


SingularityService/src/main/java/com/hubspot/singularity/data/InactiveSlaveManager.java
```diff
@@ -0,0 +1,53 @@
...
+  public Set<String> getInactiveSlaves() {
+    Optional<String[]> maybeInactiveSlaves = getData(ROOT_PATH, this.inactiveSlaveTranscoder);
+    if (maybeInactiveSlaves.isPresent()) {
+      return new HashSet<String>(Arrays.asList(maybeInactiveSlaves.get()));
+    } else {
+      return new HashSet<String>();
+    }
+  }
+
+  public void deactiveSlave(String slave) {
```
ssalinas: deactivate?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -396,14 +397,19 @@ public void loadSlavesAndRacksFromMaster(MesosMasterStateObject state, boolean i
...
-  public void checkOffer(Offer offer) {
+  public void checkOffer(Offer offer, Set<String> inactiveSlaves) {
     final String slaveId = offer.getSlaveId().getValue();
     final String rackId = slaveAndRackHelper.getRackIdOrDefault(offer);
     final String host = slaveAndRackHelper.getMaybeTruncatedHost(offer);
     final Map<String, String> textAttributes = slaveAndRackHelper.getTextAttributes(offer);
 
     final SingularitySlave slave = new SingularitySlave(slaveId, host, rackId, textAttributes, Optional.<MesosResourcesObject>absent());
 
+    if (inactiveSlaves.contains(host)) {
```
ssalinas: maybe instead of doing this check here, we can do it after we find out if it's a new slave. My idea here would be that we can find the new slave, but if it is in the 'bad hosts' list, move it immediately to decommissioned. The end goal is to not schedule tasks on a slave that might be flapping due to issues. So, we definitely don't want it as active, but if we ignore it or mark it dead we don't have any info bout the fact that it's still (somewhat) present. Moving it to decommissioned right away (maybe with a message set) lets us know it's still there, but without the danger of scheduling tasks on a bad slave.

SingularityService/src/test/java/com/hubspot/singularity/data/InactiveSlaveManagerTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+  @Test
+  public void itShouldContainAnInactiveHostWhenHostDeactivated() {
+    inactiveSlaveManager.deactiveSlave("localhost");
+
+    Assert.assertTrue(inactiveSlaveManager.getInactiveSlaves().contains("localhost"));
+  }
+
+  @Test
+  public void itShouldNotContainHostAfterActivatingHost() {
```
ssalinas: to extend these tests, you can have the class extend `SingularitySchedulerTestBase`. In there is a `resourceOffers` method which mocks the sending of an offer, allowing you to trigger the code path of getting an offer from a new slave.

SingularityService/src/main/java/com/hubspot/singularity/data/InactiveSlaveManager.java
```diff
@@ -0,0 +1,53 @@
...
+                              SingularityConfiguration configuration,
+                              MetricRegistry metricRegistry,
+                              Transcoder<String[]> inactiveSlaveTranscoder) {
+    super(curator, configuration, metricRegistry);
+
+    this.inactiveSlaveTranscoder = inactiveSlaveTranscoder;
+  }
+
+  public Set<String> getInactiveSlaves() {
+    Optional<String[]> maybeInactiveSlaves = getData(ROOT_PATH, this.inactiveSlaveTranscoder);
```
ssalinas: Wondering if it might be easier to just create empty child nodes with the proper name here, rather than continually reading/writing an array. Creation would just be creating a node at `/inactive/(hostname)`, getting all would be a `getChildren()` call, and checking if an individual one would just be an `exists()` call rather than grabbing the whole array.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosScheduler.java
```diff
@@ -135,7 +138,7 @@ public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offers) {
     final Set<Protos.OfferID> acceptedOffers = Sets.newHashSetWithExpectedSize(offers.size());
 
     for (Protos.Offer offer : offers) {
-      slaveAndRackManager.checkOffer(offer);
+      slaveAndRackManager.checkOffer(offer, inactiveSlaveManager.getInactiveSlaves());
```
ssalinas: For efficiency we should probably grab this before looping through offers so we don't grab it 50+ times

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -405,7 +406,14 @@ public void checkOffer(Offer offer) {
     final SingularitySlave slave = new SingularitySlave(slaveId, host, rackId, textAttributes, Optional.<MesosResourcesObject>absent());
 
     if (check(slave, slaveManager) == CheckResult.NEW) {
-      LOG.info("Offer revealed a new slave {}", slave);
+      if (inactiveSlaves.contains(slave.getHost())) {
```
ssalinas: alternatively, if we switch to the method of storing the hostnames I mentioned in the previous comment, we don't need to feed the list to this method and can just check if the individual inactive hostname node exists here

SingularityUI/app/components/machines/Slaves.jsx
```diff
@@ -71,7 +72,48 @@ const Slaves = (props) => {
...
+        name="Mark Inactive"
+        buttonChildren={<Glyphicon glyph="remove-circle" />}
+        action="Mark Host Inactive"
+        onConfirm={() => props.deactivateHost(slave.host)}
+        tooltipText={`Flag host '${slave.host}' as inactive`}
+        formElements={[]}
+      >
+        <p>Are you sure you want to mark the host {slave.host} as inactive?</p>
+        <p>
+          This will decommission every slave on this host until you reactivate
```
ssalinas: For clarity, maybe something like:
```
This will automatically decommission any host that joins with a matching hostname.
```
I don't think we will need to mention the piece about reactivation, since the slave will still appear in the decommissioned list with the reactivate button next to it

SingularityUI/app/components/machines/Slaves.jsx
```diff
@@ -71,7 +72,48 @@ const Slaves = (props) => {
...
+        name="Reactivate Host"
+        buttonChildren={<Glyphicon glyph="ok-circle" />}
+        action="Mark Host Active"
+        onConfirm={() => props.reactivateHost(slave.host)}
+        tooltipText={`Mark host '${slave.host}' as active`}
+        formElements={[]}
+      >
+        <p>Are you sure you want to reactivate host {slave.host}?</p>
+        <p>
+          This will allow new slaves from this host to make offers without
```
ssalinas:
```
New slaves with this hostname will no longer be automatically marked as decommissioned
```

The making offers piece is a bit misleading. For one, we get the offers form the master, not the slave. Also, we will continue to get offers from the master for that slave as long as it is still connected. From Singularity's perspective, the important part is that it is a new slave, not necessarily that it continues to make offers

SingularityUI/app/components/machines/Slaves.jsx
```diff
@@ -344,6 +388,20 @@ const Slaves = (props) => {
...
   const inactiveSlaves = props.slaves.filter(({currentState}) => Utils.isIn(currentState.state, ['DEAD', 'MISSING_ON_STARTUP']));
 
+  const inactiveHostsPanel = (inactiveHosts) => (
+    inactiveHosts && inactiveHosts.length > 0 && (
+      <div className="row">
+        <h3>Inactive Hosts</h3>
+        <p>These hosts are marked as inactive: </p>
+        <ul className="list-group">
+          {inactiveHosts.map((host) => (
+            <li className="list-group-item" key={host}>{host}</li>
```
ssalinas: might be good to be able to remove the host from this list in this location as well.

SingularityService/src/main/java/com/hubspot/singularity/data/InactiveSlaveManager.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.curator.framework.CuratorFramework;
+
+import com.codahale.metrics.MetricRegistry;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import com.hubspot.singularity.config.SingularityConfiguration;
+
+@Singleton
+public class InactiveSlaveManager extends CuratorManager {
+  private static final String ROOT_PATH = "/inactive";
```
ssalinas: nit pick here, but could we have this be `/inactiveSlaves`? when looking at the top level in zk would be easier to tell what it's for

ssalinas:
👍  Let's test this out in hs_staging

PR #1429, opened 2017-02-22
Standardize on toString/hash/equals formats

ssalinas:
Probably long overdue, but there is a mix of guava, java.util, string concat, etc in our override methods. This changes all of them to java.util.Objects for equals/hashCode and plain string concat for toString. This way there won't be any conflicts when guava changes class/method names etc.

replaces #1425 
/cc @stevenschlansker

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/mesos/Resources.java
+++ b/SingularityBase/src/main/java/com/hubspot/mesos/Resources.java
@@ -51,16 +51,6 @@ public class Resources {
   }
 
   @Override
-  public String toString() {
-    return "Resources[" +
-        "cpus=" + cpus +
-        ", memoryMb=" + memoryMb +
-        ", numPorts=" + numPorts +
-        ", diskMb=" + diskMb +
-        ']';
-  }
-
-  @Override
   public boolean equals(Object o) {
     if (this == o) {
       return true;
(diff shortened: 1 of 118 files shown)


SingularityBase/src/main/java/com/hubspot/mesos/Resources.java
```diff
@@ -51,32 +49,49 @@ public double getDiskMb() {
...
+    if (Double.compare(resources.cpus, cpus) != 0) {
+      return false;
+    }
+    if (Double.compare(resources.memoryMb, memoryMb) != 0) {
+      return false;
+    }
+    if (numPorts != resources.numPorts) {
+      return false;
+    }
+    return Double.compare(resources.diskMb, diskMb) == 0;
```
darcatron: is there a reason we broke this apart?

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequestHistory.java
```diff
@@ -93,6 +72,27 @@ public long getCreateTimestampForCalculatingHistoryAge() {
...
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(createdAt, user, eventType, request, message);
+  }
+
+  @Override
   public String toString() {
     return "SingularityRequestHistory [createdAt=" + createdAt + ", user=" + user + ", eventType=" + eventType + ", request=" + request + ", message=" + message + "]";
   }
```
darcatron: nit: should we switch this to: 
```
"SingularityRequestHistory{" +
"createdAt=" + createdAt + 
", user=" + user + 
", eventType=" + eventType + 
", request=" + request+ 
", message=" + message + 
"}";
```
to keep the format consistent with the others? Looks like there are a few other places using `[]` and formatted differently

ssalinas:
Updated for comments, thanks @darcatron

darcatron:
🚢

PR #1437, opened 2017-02-28
Defer loading log files till after page load

PtrTeixeira:
Move the action of loading the log files from S3 from `refresh`, where
the `rootComponent` blocks page load until the Promise resolves, to
`componentDidMount`, where it doesn't block page load. This is intended
as a temporary fix, rather than a best-practice course of action.

/cc @ssalinas

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskDetail.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskDetail.jsx
@@ -132,8 +132,6 @@ class TaskDetail extends Component {
     if (this.props.task.isStillRunning) {
       this.props.fetchTaskStatistics(this.props.params.taskId);
     }
-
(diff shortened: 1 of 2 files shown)


ssalinas:
thanks @PtrTeixeira

ssalinas:
@PtrTeixeira seems I was incorrect about being able to do that. We can take another look at a more permanent solution for this tomorrow

PtrTeixeira:
It might be worth just adding another parameter to `rootComponent` for 'These are things that I'd like loaded, but shouldn't block the page load'

ssalinas:
Looks good, this should be good to go to `hs_stable` @PtrTeixeira

ssalinas:
👍

PR #1442, opened 2017-03-06
Display previous overridden cleanups

ssalinas:
Right now, when a cleanup type changes we override the previous entry to change the type. It is useful to still be able to see the history of what happened. This saves previous entries in a filed on the cleanup and displays them in the ui table. They are added on a field in the entry because it would be a much larger change to try and update the zk node structure to accommodate multiple of the same type of update

/cc @darcatron

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskHistoryUpdate.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskHistoryUpdate.java
@@ -84,9 +84,9 @@ public class SingularityTaskHistoryUpdate extends SingularityTaskIdHolder implem
 
   private Set<SingularityTaskHistoryUpdate> getFlattenedPreviousUpdates(SingularityTaskHistoryUpdate update) {
     Set<SingularityTaskHistoryUpdate> previousUpdates = new HashSet<>();
-    for (SingularityTaskHistoryUpdate preivousUpdate : update.getPrevious()) {
-      previousUpdates.add(preivousUpdate.withoutPrevious());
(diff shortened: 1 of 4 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskHistoryUpdate.java
```diff
@@ -50,14 +55,44 @@ public static SimplifiedTaskState getCurrentState(Iterable<SingularityTaskHistor
...
+  public SingularityTaskHistoryUpdate withPrevious(SingularityTaskHistoryUpdate previousUpdate) {
+    Set<SingularityTaskHistoryUpdate> newPreviousUpdates = getFlattenedPreviousUpdates(this);
+    newPreviousUpdates.add(previousUpdate.withoutPrevious());
+    newPreviousUpdates.addAll(getFlattenedPreviousUpdates(previousUpdate));
+    return new SingularityTaskHistoryUpdate(getTaskId(), timestamp, taskState, statusMessage, statusReason, newPreviousUpdates);
+  }
+
+  private Set<SingularityTaskHistoryUpdate> getFlattenedPreviousUpdates(SingularityTaskHistoryUpdate update) {
+    Set<SingularityTaskHistoryUpdate> previousUpdates = new HashSet<>();
+    for (SingularityTaskHistoryUpdate preivousUpdate : update.getPrevious()) {
```
darcatron: nit: sp (previousUpdate)

ssalinas: fixed

PR #1443, opened 2017-03-06
Ignore 404s for `/priority/freeze` endpoint

PtrTeixeira:
That endpoint returns 404s as part of its standard behavior. It isn't in
this case a user-facing error message.


ssalinas:
+1 for updating prop types and code styles along the way, 🚢

PR #1445, opened 2017-03-07
use implicit acks when not offloading status updates to another thread

tpetr:
@ssalinas @wsorenson


tpetr:
(merged branch into `hs_qa`)

PR #1446, opened 2017-03-07
Disaster mode offers

ssalinas:
@wsorenson @tpetr


ssalinas:
Folded this into #1451 , doing all the disaster action stuff in one palce

PR #1448, opened 2017-03-07
disable email option

tpetr:
@ssalinas is this all that i need?


ssalinas:
👍

ssalinas:
Folded this into #1451 , doing all the disaster action stuff in one palce

PR #1450, opened 2017-03-09
Allow new tabs from global search page

PtrTeixeira:
Allow the links in global search page to be opened in a new tab with
right click > open in new tab. Unfortunately, Meta+click will not work.


ssalinas:
👍

PR #1452, opened 2017-03-10
Add a task credit system

ssalinas:
Adds a task credit system for disaster-type scenarios. By default the task credits are set to `-1` (i.e. feature is not enabled). If you give Singularity credits, the feature is now enabled and the credits are decremented each time a task is launched. If the credits reach zero, no more tasks can be launched until more credits are given, or task credits are disabled (sets it back to `-1`)

Currently this is all done in memory for speed, rather than frequently calling out to zk for something in the disasters nodes. So, all calls that do not hit the leader are proxied to the leader.

Also, fixed a few args to rootComponent that seemed to be off after the addition of the onLoad functionality (/cc @PtrTeixeira )


tpetr:
Disappointed you didn't call it Mesos Bucks.

ssalinas:
reworked to not use the proxy-to-leader piece

PR #1456, opened 2017-03-13
Bump to java 8

ssalinas:
Stream to your heart's content ❤️


ssalinas:
Last issue on this PR is with the MetricRegistry. With the updates to metrics packages, our endpoint to serialize the registry doesn't work like it used it. Currently throws some errors for any Gauges or other metrics whose names are defined as lambdas.

ssalinas:
This has been working well in hs_stable, merging so we can start using java8 features if we want 🎉

PR #1461, opened 2017-03-17
Interchange argument order in rootComponent

PtrTeixeira:
Somehow, in the muddle of branches that is Git, this slipped through
onto `master`. Interchanges the order of arguments to `rootComponent` so
that `onLoad` is at the end, and fix the only caller that uses `onLoad`
to have it in the correct position.


ssalinas:
We should double check this before/after tailer branch. The merge with that messed up a number of things for this

ssalinas:
Thanks @PtrTeixeira , will get this in 0.14.1

PR #1462, opened 2017-03-17
extract defaults to true, not executable

### LINKED ISSUE #1460: Legacy support for URIs as an array of strings changes extraction behaviour
The change in f68615f9d3 from `uris` being an `Array<String>` to `Array<SingularityMesosArtifact>` added a backwards-compatibility option to convert the strings to  `SingularityMesosArtifact`s.

That code has since been removed, but the equivalent migration (?) code that converts existing Deploys to work with 0.14.0 seems to have the same behaviour. This breaks Deploys that use compressed artifacts, since the previous behaviour was for such artifacts to be automatically extracted.

Would it be possible to change that migration logic to set `extract` to `true` for legacy URIs? Are there other suggestions for how to migrate a set of existing Deploys when moving to 0.14.0?

ssalinas:
@demobox , I had defaulted the wrong param to true (http://mesos.apache.org/documentation/latest/fetcher/)

/fixes #1460


demobox:
> I had defaulted the wrong param to true 

Aaaaah! ;-)

PR #1465, opened 2017-03-20
Pass job user as environment variable to task

PtrTeixeira:
Pass the user who clicks the "Run Now" button as the value of the
environment variable `STARTED_BY_USER` in the task environment.  Can provide
more granular, task-level permissions or auditability.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
@@ -184,6 +184,8 @@ class SingularityMesosTaskBuilder {
     setEnv(envBldr, "TASK_ID", taskId.getId());
     setEnv(envBldr, "ESTIMATED_INSTANCE_COUNT", task.getRequest().getInstancesSafe());
 
+    setEnv(envBldr, "JOB_USER", task.getPendingTask().getUser().or(""));
+
     for (Entry<String, String> envEntry : task.getDeploy().getEnv().or(Collections.<String, String>emptyMap()).entrySet()) {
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -184,6 +184,8 @@ private void prepareEnvironment(final SingularityTaskRequest task, SingularityTa
     setEnv(envBldr, "TASK_ID", taskId.getId());
     setEnv(envBldr, "ESTIMATED_INSTANCE_COUNT", task.getRequest().getInstancesSafe());
 
+    setEnv(envBldr, "JOB_USER", task.getPendingTask().getUser().or(""));
```
ssalinas: Two things.
- Wondering if `JOB_USER` is descriptive enough here. This method will get called for services and workers as well. Maybe something more like `STARTED_BY_USER` ?
- In the case where `user` is not present on the pending task, would it make more sent to leave this out rather than have an empty string as the value?

ssalinas:
👍

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilderTest.java
```diff
@@ -103,6 +106,29 @@ public void testShellCommand() {
...
+        .getCommand()
+        .getEnvironment()
+        .getVariablesList();
+
+    boolean success = false;
+    for (Variable environmentVariable : environmentVariables) {
+      success = success || (environmentVariable.getName().equals("STARTED_BY_USER") && environmentVariable.getValue().equals(user));
+    }
+
+    assertTrue("Expected env variable JOB_USER to be set to " + user, success);
```
darcatron: got an extra JOB_USER lurking here 🙂

PR #1467, opened 2017-03-21
fix task direct link to logs

darcatron:
think this is the only usage. tested locally. worked fine
@ssalinas

### DIFF
--- a/SingularityUI/app/components/tasks/Columns.jsx
+++ b/SingularityUI/app/components/tasks/Columns.jsx
@@ -361,9 +361,7 @@ export const LogLinkAndJSON = logPath => (
     cellRender={(taskId, rowData) => (
       <div className="hidden-xs">
         <OverlayTrigger placement="top" id="view-log-overlay" overlay={logTooltip}>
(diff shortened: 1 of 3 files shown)


SingularityUI/app/components/tasks/Columns.jsx
```diff
@@ -361,8 +361,7 @@ export const LogLinkAndJSON = logPath => (
     cellRender={(taskId, rowData) => (
       <div className="hidden-xs">
         <OverlayTrigger placement="top" id="view-log-overlay" overlay={logTooltip}>
-          <Link
-            to={`request/${taskId.requestId}/tail/${logPath}?taskIds=${taskId.id}`}
+          <Link to={`task/${taskId.id}/tail/${Utils.substituteTaskId(config.runningTaskLogPath, taskId.id)}`}
```
ssalinas: why `config.runningTaskLogPath` instead of `logPath` here? `logPath` is supplied as an argument.

darcatron: my mistake, good catch. updated

ssalinas:
👍 LGTM

darcatron:
moved the path into a util func to prevent this issue from happening again

PR #1468, opened 2017-03-21
Transition onLoad to an action dispatcher

PtrTeixeira:
One of the changes that was blocked on the tailer was making `onLoad`
behave like an action dispatcher, rather than a function. So each of the
components treats it as an object (or, at least, the base case is `null`
rather than `_.noop`).  This means that it was misbehaving when it
wasn't explicitly set because pre-tailer it was expected to be a
function.

More specifically, some pages were permanently stuck with the page loaded, but a visible loading spinner still present


ssalinas:
Funny enough I already snuck that into https://github.com/HubSpot/Singularity/pull/1466. Looked like it was already that way on some branches and not others. Will roll it out with that one 👍

PR #1472, opened 2017-03-27
Add spacing line at the bottom of logs

PtrTeixeira:
Add a small amount of padding at the bottom of the logging container.
Now if you scroll all the way to the bottom of log tailer, there should
be a small space between the last log line and the bottom of the screen,
so that the scrollbar (on MacOS) won't overlap the last line of the
text.

### DIFF
--- a/SingularityUI/app/styles/scss/new-tailer.scss
+++ b/SingularityUI/app/styles/scss/new-tailer.scss
@@ -31,6 +31,7 @@
       flex-flow: column;
       width: 100%;
       height: 100%;
+      padding-bottom: 10px;
 
       overflow: scroll;
       // will repaint on scroll if you don't put this in!!!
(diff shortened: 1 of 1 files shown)


darcatron:
does this work okay for multiple instances as well? just wondering since the containers stack when doing more than 3

ssalinas:
Maybe add some screenshots for both single and multiple tail cases here?

PtrTeixeira:
Old: 
screen shot 2017-03-29 at 9 08 58 am
screen shot 2017-03-29 at 9 33 45 am


New:
screen shot 2017-03-29 at 9 53 13 am 2
screen shot 2017-03-29 at 9 49 21 am 2

ssalinas:
I like it, LGTM 🥇

PR #1479, opened 2017-03-29
Reduce timeout on requests to task sandboxes

PtrTeixeira:
Previously there was a timeout associated with http request fetching the log files from the task sandbox, but not one associated with browsing the task's sandbox. This adds the same timeout period to browsing the sandbox. It also slightly drops the timeout period on the request, from 5 seconds to 2 seconds.


ssalinas:
👍 will be good to have a timeout on both in case the slave isn't responding. Most calls to read/browse seem to easily be under 500ms when working properly.

PR #1482, opened 2017-03-30
Better task balancing

darcatron:
:construction: This is a WIP for task balancing.

The general idea is that a pendingTask will calculate the best offer by scoring each offer and choosing the best score. Right now, the top score is 1.00 assuming the slave has nothing on it.

Score is weighted based on 2 real criteria: current resource usage for the same request type and current resource availability. I choose weights based on what I thought might be important, but it can be changed. I thought mem would be the most important so I weighted it a bit higher:

```
requestTypeCpuWeight = 0.20;
requestTypeMemWeight = 0.30;
freeCpuWeight = 0.20;
freeMemWeight = 0.30;
```

This only scores based on what's running on the slave. It does not look at the acceptedPendingTasks for an offer.

@ssalinas


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -118,33 +129,44 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration, Cus
...
+
+          if (configuration.getMaxTasksPerOffer() > 0 && offerHolder.getAcceptedTasks().size() >= configuration.getMaxTasksPerOffer()) {
+            LOG.trace("Offer {} is full ({}) - skipping", offerHolder.getOffer(), offerHolder.getAcceptedTasks().size());
+            continue;
+          }
+
+          double score = score(offerHolder, stateCache, tasksPerOfferPerRequest, taskRequestHolder, getUsagesPerRequestTypePerSlave());
+          if (score > 0) {
+            // todo: can short circuit here if score is high enough
+            scorePerOffer.put(offerHolder, score);
```
darcatron: Thought we might want to have a value that's definitely good enough to just accept instead of continue evaluating

ssalinas:
I like the idea of a scoring system overall. Some comments on specific logic things I'll make later since this is still WIP. Overall comments though:
- The time past due for a task should probably also factor in to the scoring. (i.e. if it was supposed to run 5 minutes ago, the definition of 'good enough' is different than if it was supposed to start 2 seconds ago)
- There should probably also be a measure of how many offers we have looked at while trying to schedule this task. We may only get 1,2 etc offers at a time and not have a wholistic view of resources. So if we have looked at a number of them already, the bar of 'good enough' should start to get lower.
- We should definitely be aware of computation time. The offer processing loop is already one of our slower areas. Anything we can do to pre-process this data and require less calculation at offer evaluation time will be a big help

darcatron:
This has been updated as follows:

### Before
All offers would be scored and of all scores above `0`, the best scored offer would be accepted by the task

### After
All offers are still scored, but the minimum score acceptable depends on the task's overdue milliseconds and number of offers a task has not accepted. 

Currently, the overdue time and offer count have a max of `10 min` and `20 attempts`, respectively. The min score is based on the ratio of `curOverdueTime:maxOverdueTime` and `curAttempts:maxAttempts` 

The offer attempts count is any offer that was considered, not just offers that scored too low. So an offer that didn't have enough resources to satisfy the task will still be counted. 

I picked these numbers based on generalizations. We will likely have to tune these

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -109,32 +109,14 @@ public SlaveMatchState doesOfferMatch(SingularityOfferHolder offer, SingularityT
...
-      LOG.trace("Task requires slave with attributes {}, (slave attributes are {})", taskRequest.getRequest().getRequiredSlaveAttributes().get(), offer.getTextAttributes());
+    if (!isSlaveAttributesMatch(offer, taskRequest)) {
       return SlaveMatchState.SLAVE_ATTRIBUTES_DO_NOT_MATCH;
     }
 
     final SlavePlacement slavePlacement = taskRequest.getRequest().getSlavePlacement().or(configuration.getDefaultSlavePlacement());
 
     if (!taskRequest.getRequest().isRackSensitive() && slavePlacement == SlavePlacement.GREEDY) {
+      // todo: account for this or let this behavior continue?
       return SlaveMatchState.NOT_RACK_OR_SLAVE_PARTICULAR;
```
darcatron: I didn't know if we would need to account for any rack sensitivity outside of the existing checks done in the scheduler

darcatron:
@ssalinas Got some specific logic tests in. Let me know if there's a piece I missed that should be added. I'm going to continue to try to get some full logic tests in as well

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -115,36 +116,57 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration, Cus
...
-      for (SingularityOfferHolder offerHolder : offerHolders) {
-        if (configuration.getMaxTasksPerOffer() > 0 && offerHolder.getAcceptedTasks().size() >= configuration.getMaxTasksPerOffer()) {
-          LOG.trace("Offer {} is full ({}) - skipping", offerHolder.getOffer(), offerHolder.getAcceptedTasks().size());
-          continue;
+      for (SingularityTaskRequestHolder taskRequestHolder : pendingTaskIdToTaskRequest.values()) {
+
+        Map<SingularityOfferHolder, Double> scorePerOffer = new HashMap<>();
+        double minScore = minScore(taskRequestHolder.getTaskRequest(), offerMatchAttemptsPerTask, System.currentTimeMillis());
+
+        LOG.info("Minimum score for task {} is {}", taskRequestHolder.getTaskRequest().getPendingTask().getPendingTaskId().getId(), minScore);
```
ssalinas: probably can be lower than info level here

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -115,36 +116,57 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration, Cus
...
+
+        for (SingularityOfferHolder offerHolder : offerHolders) {
+
+          if (configuration.getMaxTasksPerOffer() > 0 && offerHolder.getAcceptedTasks().size() >= configuration.getMaxTasksPerOffer()) {
+            LOG.trace("Offer {} is full ({}) - skipping", offerHolder.getOffer(), offerHolder.getAcceptedTasks().size());
+            continue;
+          }
+
+          double score = score(offerHolder, stateCache, tasksPerOfferPerRequest, taskRequestHolder, getSlaveUsage(currentSlaveUsages, offerHolder.getOffer().getSlaveId().getValue()));
```
ssalinas: for clarity, maybe something like 'hostScore' here? The score is for the particular slave, not necessarily about the offer

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -157,76 +179,150 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration, Cus
...
+  }
 
-        stateCache.getActiveTaskIds().add(task.getTaskId());
-        stateCache.getActiveTaskIdsForRequest(task.getTaskRequest().getRequest().getId()).add(task.getTaskId());
-        addRequestToMapByOfferId(tasksPerOfferPerRequest, offerId, taskRequest.getRequest().getId());
-        stateCache.getScheduledTasks().remove(taskRequest.getPendingTask());
+  @VisibleForTesting
+  double minScore(SingularityTaskRequest taskRequest, Map<String, Integer> offerMatchAttemptsPerTask, long now) {
+    double minScore = 0.80;
+    int maxOfferAttempts = 20;
```
ssalinas: another that would be nice to have ocnfigurable

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -157,76 +179,150 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration, Cus
...
-          taskRequestHolder.getRequestedPorts());
-      final SlaveMatchState slaveMatchState = slaveAndRackManager.doesOfferMatch(offerHolder, taskRequest, stateCache);
+    return 0;
+  }
 
-      if (matchesResources && slaveMatchState.isMatchAllowed()) {
-        final SingularityTask task = mesosTaskBuilder.buildTask(offerHolder.getOffer(), offerHolder.getCurrentResources(), taskRequest, taskRequestHolder.getTaskResources(), taskRequestHolder.getExecutorResources());
+  @VisibleForTesting
+  double score(Offer offer, SingularityTaskRequest taskRequest, Optional<SingularitySlaveUsageWithId> maybeSlaveUsage) {
+    double requestTypeCpuWeight = 0.20;
```
ssalinas: Let's make these configurable, maybe another object in the configuration yaml?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -157,76 +179,150 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration, Cus
...
-      final boolean matchesResources = MesosUtils.doesOfferMatchResources(taskRequest.getRequest().getRequiredRole(), taskRequestHolder.getTotalResources(), offerHolder.getCurrentResources(),
-          taskRequestHolder.getRequestedPorts());
-      final SlaveMatchState slaveMatchState = slaveAndRackManager.doesOfferMatch(offerHolder, taskRequest, stateCache);
+    return 0;
+  }
 
-      if (matchesResources && slaveMatchState.isMatchAllowed()) {
-        final SingularityTask task = mesosTaskBuilder.buildTask(offerHolder.getOffer(), offerHolder.getCurrentResources(), taskRequest, taskRequestHolder.getTaskResources(), taskRequestHolder.getExecutorResources());
+  @VisibleForTesting
+  double score(Offer offer, SingularityTaskRequest taskRequest, Optional<SingularitySlaveUsageWithId> maybeSlaveUsage) {
```
ssalinas: Let's go over this one in-person, think we are getting close, just easier to chat than typing a novel in github ;)

SingularityBase/src/main/java/com/hubspot/singularity/SingularitySlaveUsage.java
```diff
@@ -1,22 +1,40 @@
...
+import java.util.Optional;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class SingularitySlaveUsage {
 
+  public static final String CPU_USED = "cpusUsed";
+  public static final String MEMORY_BYTES_USED = "memoryRssBytes";
+  public static final long BYTES_PER_MEGABYTE = 1024L * 1024L;
```
darcatron: Yeah, I was sad to see there wasn't a lib method for this too 😢

SingularityBase/src/main/java/com/hubspot/singularity/SingularitySlaveUsage.java
```diff
@@ -35,6 +53,30 @@ public int getNumTasks() {
...
+  public Optional<Double> getCpuTotal() {
+    return cpuTotal;
+  }
+
+  public Map<RequestType, Map<String, Number>> getUsagePerRequestType() {
+    return usagePerRequestType;
+  }
+
+  public double getCpusUsedForRequestType(RequestType type) {
+    return usagePerRequestType.get(type).get(CPU_USED).doubleValue();
```
darcatron: agreed, mapping would be clearer then too 👍

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -26,23 +33,33 @@
...
     this.mesosClient = mesosClient;
     this.usageManager = usageManager;
     this.exceptionNotifier = exceptionNotifier;
+    this.requestManager = requestManager;
   }
 
   @Override
   public void runActionOnPoll() {
     final long now = System.currentTimeMillis();
+    Map<RequestType, Map<String, Number>> usagesPerRequestType = new HashMap<>();
```
darcatron: This should be per slave. This poller loops through each slave and creates a new `SingularitySlaveUsage` with the stats for that slave

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
darcatron: Yup, I was in progress on this (now committed), but I kept the fields under SingularityConfiguration since I saw a lot of other stuff in there as well (e.g. caching). We could pull it into an OfferConfiguration file if you think that'd be better for organization

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
darcatron: I'm not sure about the naming here. We do look at the slave's utilization to score the offer, but we are still scoring the offer itself since offers aren't uniquely 1:1 for a slave (e.g. 2 offers for the same slave). 

The slave utilization weight will be the same for all offers on the same slave, but the offer resources will be different per offer.  So, it seems to me that we're scoring the offer in this class rather than the slave itself

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -86,7 +91,9 @@ public void runActionOnPoll() {
 
             double taskCpusUsed = ((usage.getCpuSeconds() - lastUsage.getCpuSeconds()) / (usage.getTimestamp() - lastUsage.getTimestamp()));
 
-            updateUsagesPerRequestType(usagesPerRequestType, getRequestType(taskUsage), usage.getMemoryRssBytes(), taskCpusUsed);
+            if (getRequestType(taskUsage).isLongRunning() ||  isConsideredLongRunning(taskUsage)) {
+              updateLongRunningTasksUsage(longRunningTasksUsage, usage.getMemoryRssBytes(), taskCpusUsed);
+            }
```
darcatron: Added this to also include the usage for tasks that are non long running, but run for a considerable amount of time to warrant adding their usage

darcatron: This is not outdated GitHub..... 
bitmoji

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -258,22 +261,61 @@ private double score(SingularityOfferHolder offerHolder, SingularitySchedulerSta
...
+
+      if (usedResourceWeight == maxNonLongRunningUsedResourceWeight) {
+        return scoreLongRunningTask(longRunningMemUsedScore, memTotalScore, longRunningCpusUsedScore, cpusTotalScore);
+      }
+      freeResourceWeight = 1 - usedResourceWeight;
+    }
+
+    // usage reduces score
+    return calculateScore(longRunningMemUsedScore, memTotalScore, longRunningCpusUsedScore, cpusTotalScore, freeResourceWeight, usedResourceWeight * -1);
+  }
```
darcatron: These changes will be easiest to discuss in person.

Essentially, long running tasks consider an offer 50:50 for usage and free space. non long running offers vary depending on how long they've been running with a default at the middle (25:75). non long running offers favor free space and will score lower if they run for a large amount of time **and** there are long running tasks on the  offered machine

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosOfferSchedulerTest.java
```diff
@@ -49,68 +62,239 @@ public SingularityMesosOfferSchedulerTest() {
...
+  public void itCorrectlyScoresEmptySlaveNonLongRunningTasks() {
+    String slaveId = "slave";
+    Map<ResourceUsageType, Number> longRunningTasksUsage = new HashMap<>();
+    setRequestType(RequestType.ON_DEMAND);
+
+    // new slave (no resources used) -> near perfect score
+    setDeployStatistics(TimeUnit.MINUTES, 5);
+    longRunningTasksUsage.put(ResourceUsageType.CPU_USED, 0);
+    longRunningTasksUsage.put(ResourceUsageType.MEMORY_BYTES_USED, 0);
+    assertScoreIs(.993, scheduler.score(getOffer(10, 10, slaveId), taskRequest, Optional.of(getUsage(10, 10, longRunningTasksUsage))));
```
darcatron: something interesting to note here. A _perfect_ offer for non long running tasks would be a task that runs for 0 seconds on an empty slave. Anything after that will reduce the score. This doesn't hurt us b/c it'll score each offer in the same regard, but it will likely mean that non long running tasks should have a lower minimum score. They are less picky than long running tasks

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -188,6 +188,26 @@
 
   private int maxTasksPerOfferPerRequest = 0;
 
+  private double minOfferScore = 0.80;
+
+  private int maxOfferAttemptsPerTask = 20;
+
+  private long maxMillisPastDuePerTask = TimeUnit.MINUTES.toMillis(10);
```
ssalinas: thinking something shorter would be a better default. We can check in testing, but I think after even 5 minutes we shouldn't be worrying about score and just scheduling asap

SingularityService/src/main/java/com/hubspot/singularity/data/UsageManager.java
```diff
@@ -31,22 +32,30 @@
 
   private static final String SLAVE_PATH = ROOT_PATH + "/slaves";
   private static final String TASK_PATH = ROOT_PATH + "/tasks";
+  private static final String USAGE_SUMMARY_PATH = ROOT_PATH + "/summary";
```
darcatron: @ssalinas can you take a look at this piece? Just wanna make sure I got this set up right

SingularityService/src/main/java/com/hubspot/singularity/data/UsageManager.java
```diff
@@ -123,6 +140,10 @@ public SingularityCreateResult saveSpecificTaskUsage(String taskId, SingularityT
     return save(getSpecificTaskUsagePath(taskId, usage.getTimestamp()), usage, taskUsageTranscoder);
   }
 
+  public SingularityCreateResult saveSpecificClusterUtilization(SingularityClusterUtilization utilization) {
+    return save(getSpecificClusterUtilizationPath(utilization.getTimestamp()) , utilization, clusterUtilizationTranscoder);
```
ssalinas: is there any point at which we would want a history of these? If we don't need a history of summaries, we might as well `save` the data to the summary path (without timestamp) and just overwrite when there is new data.

SingularityService/src/main/java/com/hubspot/singularity/data/UsageManager.java
```diff
@@ -95,6 +104,10 @@ private String getCurrentSlaveUsagePath(String slaveId) {
     return ZKPaths.makePath(getSlaveUsagePath(slaveId), CURRENT_USAGE_NODE_KEY);
   }
 
+  private String getSpecificClusterUtilizationPath(long timestamp) {
```
ssalinas: why `SpecificClusterUtilization` instead of just `getUsageSummaryPath` or something like that?

SingularityService/src/main/java/com/hubspot/singularity/data/UsageManager.java
darcatron: I didn't see a reason to at this point. Saw the others were saving them so it seemed okay since we only save up to 5 points. I think it's safe to just have the one point. Keeps it simpler

SingularityService/src/main/java/com/hubspot/singularity/data/UsageManager.java
darcatron: the "specific" keyword was the pattern the other items were using so I kept that since the timestamp specified which one to grab. I can drop the historical data and then rename it

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -51,6 +88,13 @@ public void runActionOnPoll() {
 
         for (MesosTaskMonitorObject taskUsage : allTaskUsage) {
           String taskId = taskUsage.getSource();
+          SingularityTaskId task;
+          try {
+            task = SingularityTaskId.valueOf(taskId);
+          } catch (InvalidSingularityTaskIdException e) {
+            LOG.error("Couldn't get SingularityTaskId for {}", taskUsage);
+            continue;
+          }
```
darcatron: added this try catch for the potentially incorrect taskId types

darcatron:
### Before
Min score was configurable by the user and would be reduced as tasks were delayed and offer match attempts were rejected. 

### After
Min score is calculated based on the overall utilization of the cluster. `memUsed / memTotal` and `cpuUsed / cpuTotal`. This will help give us a min score closer to the actual offer scores. Task delays and match attempts will still reduce the min score

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -130,12 +131,13 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
 
     while (!pendingTaskIdToTaskRequest.isEmpty() && addedTaskInLastLoop && canScheduleAdditionalTasks(taskCredits)) {
       addedTaskInLastLoop = false;
+      double maxTaskMillisPastDue = maxTaskMillisPastDue(SingularityScheduledTasksInfo.getInfo(taskManager.getPendingTasks(), configuration.getDeltaAfterWhichTasksAreLateMillis()).getMaxTaskLagMillis());
```
darcatron: I think we'll need a better name for this variable. It's the max possible task lag before we decide we're going to take any we can get. I feel `maxTaskMillisPastDue` has some confusing overlap with `maxTaskLag` which is the current highest lag time for pending tasks

darcatron:
#### Before

We had a configured value for the max possible task lag before we accepted any offer that matched a task. The default was 5 minutes.

#### After
The max possible task lag (`maxTaskMillisPastDue`) is determined by a simple point-slope formula that uses the `maxTaskLag` and the existing `deltaAfterWhichTasksAreLateMillis`. 
`maxTaskMillisPastDue = (-180 000 / deltaAfterWhichTasksAreLateMillis) * maxTaskLag + 180 000` where 180,000 is 3 min in milliseconds

A task with no lag will result in a 3 min `maxTaskMillisPastDue`. Any lag will linearly reduce the `maxTaskMillisPastDue`. The minimum possible `maxTaskMillisPastDue` is `1 millisecond`. 

Since we reduce the `minScore` by how long a task is past due: higher `maxTaskLag` => lower `maxTaskMillisPastDue` => lower `minScore`

It's also important to note that we will actually start accepting any matching offer (minScore == 0) before the `maxTaskLag` reaches `deltaAfterWhichTasksAreLateMillis`. We can use this formula to see what the values would be for different deltas.
`180 000 = l + (180 000 / d) * l` where `l = the task lag at which we'll accept any offer` and `d = deltaAfterWhichTasksAreLateMillis`

ssalinas:
@darcatron I think this one is good to merge. Going to get this in master so we can continue to build off of the resource usage updates without endless merge conflicts 👍  . Thanks for all the work on this one

PR #1483, opened 2017-03-31
switch not loaded to loading

darcatron:
I've found this to be confusing since it'll show `0` if it's empty or load the contents soon after. I instinctively refresh because it seems like a failure message

@ssalinas


ssalinas:
👍  Keep in mind we'll need to bump the tailer version and publish to pull this in

PR #1487, opened 2017-04-04
Scale message

darcatron:
Adds the instance number change to the scale history message.

Also removed user from the bounce after scale message since the user already shows up in the user column

@ssalinas

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/helpers/RequestHelper.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/helpers/RequestHelper.java
@@ -102,7 +102,7 @@ public class RequestHelper {
               System.currentTimeMillis(), Optional.<Boolean> absent(), newRequest.getId(), Optional.of(maybeDeployId.get()), skipHealthchecks, message, actionId, maybeBounceRequest.get().getRunShellCommandBeforeKill()));
 
           if (createResult != SingularityCreateResult.EXISTED) {
-            requestManager.bounce(newRequest, System.currentTimeMillis(), user, Optional.of(String.format("Bouncing due to bounce after scale by %s", user)));
+            requestManager.bounce(newRequest, System.currentTimeMillis(), user, Optional.of("Bouncing due to bounce after scale"));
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/helpers/RequestHelper.java
```diff
@@ -102,7 +102,7 @@ private void checkReschedule(SingularityRequest newRequest, Optional<Singularity
               System.currentTimeMillis(), Optional.<Boolean> absent(), newRequest.getId(), Optional.of(maybeDeployId.get()), skipHealthchecks, message, actionId, maybeBounceRequest.get().getRunShellCommandBeforeKill()));
 
           if (createResult != SingularityCreateResult.EXISTED) {
-            requestManager.bounce(newRequest, System.currentTimeMillis(), user, Optional.of(String.format("Bouncing due to bounce after scale by %s", user)));
```
ssalinas: why remove the user here?

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -581,6 +581,12 @@ public SingularityRequestParent scale(@ApiParam("The Request ID to scale") @Path
     validator.checkScale(newRequest, Optional.<Integer>absent());
 
     checkBadRequest(oldRequest.getInstancesSafe() != newRequest.getInstancesSafe(), "Scale request has no affect on the # of instances (%s)", newRequest.getInstancesSafe());
+    String scaleMessage = String.format("Scaling from %d instance(s) to %d instances", oldRequest.getInstancesSafe(), newRequest.getInstancesSafe());
```
ssalinas: given the size of the area it is displayed in the ui, this can probably be even shorter. i.e. `from %d to %s` or even `%d -> %d`

SingularityService/src/main/java/com/hubspot/singularity/helpers/RequestHelper.java
darcatron: It's already in the user column so it seemed repetitive

ssalinas:
👍

PR #1488, opened 2017-04-04
Extend the leader and web caching

ssalinas:
Still a WIP, but this will extend the leader and web cache to other frequently used pieces of information. To start with I extended it to requests (`SingularityRequestWithState`)

/cc @wsorenson

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
@@ -354,7 +354,7 @@ public class RequestManager extends CuratorAsyncManager {
   }
 
   public void activateLeaderCache() {
-    leaderCache.cacheRequests(getRequests());
+    leaderCache.cacheRequests(fetchRequests());
   }
 
   public List<SingularityRequestWithState> getRequests() {
@@ -369,7 +369,7 @@ public class RequestManager extends CuratorAsyncManager {
     if (useWebCache && webCache.useCachedRequests()) {
       return webCache.getRequests();
     }
-    List<SingularityRequestWithState> requests = getAsyncChildren(NORMAL_PATH_ROOT, Optional.of(requestIdCache), requestTranscoder);
+    List<SingularityRequestWithState> requests = fetchRequests();
 
     if (useWebCache) {
       webCache.cacheRequests(requests);
@@ -377,7 +377,9 @@ public class RequestManager extends CuratorAsyncManager {
     return requests;
   }
 
-
+  public List<SingularityRequestWithState> fetchRequests() {
+    return getAsyncChildren(NORMAL_PATH_ROOT, Optional.of(requestIdCache), requestTranscoder);
+  }
 
   public Optional<SingularityRequestWithState> getRequest(String requestId) {
     return getData(getRequestPath(requestId), requestTranscoder);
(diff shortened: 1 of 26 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -282,6 +287,24 @@ public SingularityDeleteResult markDeleted(SingularityRequest request, long time
...
+    if (leaderCache.active()) {
+      return leaderCache.getRequests().stream().filter((r) -> requestIds.contains(r.getRequest().getId())).collect(Collectors.toList());
+    }
+
+    if (useWebCache) {
+      if (webCache.useCachedRequests()) {
+        return webCache.getRequests().stream().filter((r) -> requestIds.contains(r.getRequest().getId())).collect(Collectors.toList());
+      } else {
+        List<SingularityRequestWithState> requests = getRequests(true);
+        webCache.cacheRequests(requests);
```
wsorenson: Aren't these requests specific?

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -510,7 +510,7 @@ public void testOneOffsDontMoveDuringDecomission() {
     requestResource.postRequest(bldr.build());
     deploy("d2");
 
-    requestResource.scheduleImmediately(requestId);
+    requestResource.scheduleImmediately(requestId, Optional.absent());
```
wsorenson: would it have made sense to create a method here that didn't take the second parameter?

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -208,7 +213,7 @@ public SingularityCreateResult update(SingularityRequest request, long timestamp
 
   public SingularityCreateResult save(SingularityRequest request, RequestState state, RequestHistoryType eventType, long timestamp, Optional<String> user, Optional<String> message) {
     saveHistory(new SingularityRequestHistory(timestamp, user, eventType, request, message));
-
+    leaderCache.putRequest(new SingularityRequestWithState(request, state, timestamp));
```
wsorenson: shouldn't there be a delete? is this the only modify call?

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
ssalinas: I was debating if it was helpful to repopulate the cache here. This fetches all the requests then the returned values are the filtered ones. This particular method is called a bunch by the authorization helper to fetch requests allowed to be viewed by a user. We would previously only fetch the relevant ones from zk (still a large list on average), but that's not enough to repopulate the cache since it isn't the full list

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
ssalinas: The `DELETE` for a request right now is actually just a history update, we never actually remove the node. All the getRequests calls are fetching 'deleted' things still, they are just generally filtered out based on their current state

ssalinas:
Going to merge this into the other leader cache PR so changes are all in one place

PR #1493, opened 2017-04-07
Immediately decline offers from decommissioned slaves

ssalinas:


wsorenson:
Does this cover frozen/decommissioned/decommissioning?

ssalinas:
I had forgotten frozen, takes into account all of the decommissioning states though.

ssalinas:
and another into leader cache pr, now that frozen slave check has been added

PR #1494, opened 2017-04-07
Use web cache for request groups

ssalinas:


ssalinas:
merging this to the leader cache pr as well

PR #1495, opened 2017-04-07
use a zk cache for deploy data

ssalinas:


ssalinas:
one more into leader cache pr

PR #1496, opened 2017-04-10
leader cache for task history updates

ssalinas:


ssalinas:
consolidating into leader cache as well

PR #1497, opened 2017-04-10
Delay pollers when under heavy load

ssalinas:
Still working out if this is the right strategy. This would allow each poller to determine if it was previously under a heavier than normal load, and then delay it's next run such that other items contending for the scheduler lock (e.g. status updates) have time to get in. Implemented just cleanup poller to start, based on number of cleanup tasks

/cc @wsorenson


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleanupPoller.java
```diff
@@ -20,21 +21,36 @@
 
   private final SingularityCleaner cleaner;
   private final DisasterManager disasterManager;
+  private final int delayCleanupWhenAboveTasks;
+  private final long delayCleanupForMs;
+
+  private AtomicLong nextRunAfter = new AtomicLong(0);
```
wsorenson: this should be final

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityLeaderOnlyPoller.java
```diff
@@ -35,6 +36,8 @@
   private SingularityAbort abort;
   private SingularityMesosSchedulerDelegator mesosScheduler;
 
+  private AtomicLong nextRunAfter = new AtomicLong(0);
```
wsorenson: this isn't used

ssalinas:
Sry for unused vars and such, hadn't finished this one yet. I'll clean it up in the morning

ssalinas:
Changed the strategy on this one. Singularity will now keep track of the average status update delta over the last 30s (avg delta is exposed in the state as well). If the delta is over a configured amount (default of 30s), the major scheduler-lock-consuming pollers will all short circuit and offer processing will immediately decline all offers until the delta is back under the configured value. The delta is updated on a 5 second interval by a separate poller.

ssalinas:
merging this into the other leader cache PR

PR #1498, opened 2017-04-10
Forbid health checks longer than kill time

PtrTeixeira:
Singularity is configured to automatically kill tasks which are
nonresponsive for a certain amount of time after startup
(`killAfterTasksDoNotRunDefaultSeconds`, set in the config file).  If a
task is configured such that it won't start accepting healthchecks for
longer than that interval, Singularity will just kill it off, without
ever sending health checks. This makes it so that that task
configuration will be rejected ahead of time by Singularity.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
@@ -313,7 +313,7 @@ public class SingularityValidator {
       int startUpDelay = deploy.getHealthcheck().get().getStartupDelaySeconds().get();
 
       checkBadRequest(startUpDelay < defaultKillAfterNotHealthySeconds,
-          String.format("Health check startup delay time must be less than kill after wait time %s (was %s)", defaultKillAfterNotHealthySeconds, startUpDelay));
+          String.format("Health check startup delay time must be less than %s (was %s)", defaultKillAfterNotHealthySeconds, startUpDelay));
     }
 
     checkBadRequest(deploy.getCommand().isPresent() && !deploy.getExecutorData().isPresent() ||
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -302,10 +304,18 @@ public SingularityDeploy checkDeploy(SingularityRequest request, SingularityDepl
...
+
       checkBadRequest((startupTime + ((httpTimeoutSeconds + intervalSeconds) * attempts)) > maxTotalHealthcheckTimeoutSeconds.get(),
         String.format("Max healthcheck time cannot be greater than %s, (was startup timeout: %s, interval: %s, attempts: %s)", maxTotalHealthcheckTimeoutSeconds.get(), startupTime, intervalSeconds, attempts));
     }
 
+    if (deploy.getHealthcheck().isPresent() && deploy.getHealthcheck().get().getStartupDelaySeconds().isPresent()) {
+      int startUpDelay = deploy.getHealthcheck().get().getStartupDelaySeconds().get();
+
+      checkBadRequest(startUpDelay < defaultKillAfterNotHealthySeconds,
+          String.format("Health check startup delay time must be less than kill after wait time %s (was %s)", defaultKillAfterNotHealthySeconds, startUpDelay));
```
ssalinas: probably not as important for the user to know that we call it the 'kill after wait time'. Maybe just `Health check startup delay time must be less than {time} seconds (was {time} seconds)`

Other than that LGTM

SingularityService/src/test/java/com/hubspot/singularity/data/ValidatorTest.java
```diff
@@ -44,14 +50,34 @@ public void itForbidsBracketCharactersInDeployIds() throws Exception {
...
+        .build();
+    SingularityDeploy deploy = SingularityDeploy
+        .newBuilder("1234567", "1234567")
+        .setHealthcheck(Optional.of(healthCheck))
+        .build();
+    SingularityRequest request = new SingularityRequestBuilder("1234567", RequestType.SERVICE).build();
+
+    WebApplicationException exn = (WebApplicationException) catchThrowable(() -> validator.checkDeploy(request, deploy));
+    assertThat((String) exn.getResponse().getEntity())
+        .contains("Health check startup delay");
```
darcatron: whoa, this catch is cool

PR #1500, opened 2017-04-11
Fix proxy to leader calls

ssalinas:
Singularity was having trouble shutting down after the proxy to leader additions. From thread dumps the http client seemed to be the culprit. Since we already have the managed SingularityAsyncHttpClient, I'm updating to use that rather than have a separate one. Also removes the dep on horizon.


ssalinas:
merging this into the base leader cache PR

PR #1501, opened 2017-04-11
Allow copying from the JSON button dialog

PtrTeixeira:
Previously the copy button on the JSON modal would not copy the text of
the modal into the clipboard. This was because by default Bootstrap
Modals demand focus, and clipboard js works by creating a hidden text
field and copying the text out of that.  The fix is allowing the modal
to yield focus to the hidden text field.

There was an additional performance problem that went into the problem.
In particular, attaching the clipboard to a class attaches the clipboard
to every modal on the page, even when the modal isn't open. Tis was
fixed by starting the clipboard only when the modal is open and
destroying it after the modal is closed, to prevent attaching many
handlers that accumulate on the page.

### DIFF
--- a/SingularityUI/app/components/common/JSONButton.jsx
+++ b/SingularityUI/app/components/common/JSONButton.jsx
@@ -25,11 +25,7 @@ export default class JSONButton extends Component {
       modalOpen: false
     };
 
-    this.showJSON = this.showJSON.bind(this);
-    this.hideJSON = this.hideJSON.bind(this);
-
-    this.attachClipboard = this.attachClipboard.bind(this);
-    this.removeClipboard = this.removeClipboard.bind(this);
+    _.bindAll(this, 'showJSON', 'hideJSON', 'attachClipboard', 'removeClipboard');
   }
 
   showJSON() {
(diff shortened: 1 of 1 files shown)


ssalinas:
+1 for adding thing performance fix in as well 😄 LGTM

SingularityUI/app/components/common/JSONButton.jsx
```diff
@@ -27,10 +27,9 @@ export default class JSONButton extends Component {
 
     this.showJSON = this.showJSON.bind(this);
     this.hideJSON = this.hideJSON.bind(this);
-  }
 
-  componentDidMount() {
-    this.clipboard = new Clipboard('.copy-btn');
+    this.attachClipboard = this.attachClipboard.bind(this);
```
andyhuang91: You can bind the functions in the constructor along with the `showJSON` and `hideJSON` functions. There's no reason to do this in componentDidMount. If you have lots of functions that you need to bind, you can use `_.bindAll`

ssalinas: @andyhuang91 , it is in the constructor, if you look a bit up in the diff, you'll see he's removing the `componentDidMount() {` line

PR #1503, opened 2017-04-12
Don't show logs panel if task never running

PtrTeixeira:
If a task is currently stopped and never reached the RUNNING state,
don't show the "Logs" section, or include a link to
"tail_of_finished_service.log".

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskState.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskState.jsx
@@ -1,0 +1,48 @@
+import React, { PropTypes } from 'react';
+import Utils from '../../utils';
+import TaskStatus from './TaskStatus';
+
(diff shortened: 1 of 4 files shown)


SingularityUI/app/components/taskDetail/TaskState.jsx
```diff
@@ -0,0 +1,48 @@
+import React, { PropTypes } from 'react';
+import Utils from '../../utils';
+import TaskStatus from './TaskStatus';
+
+const labelText = (status, currentState, cleanupType) => {
+  if (status === TaskStatus.NEVER_RAN) {
+    return 'Task aborted';
```
ssalinas: 👍  for this, I think this will make things a lot clearer for users

ssalinas:
Good to go ahead and test in staging whenever you like @PtrTeixeira

PR #1505, opened 2017-04-12
don't include cleaning tasks in instance count

darcatron:
Tasks that are cleaning should not be part of the instance count when checking for over provisioned requests

@ssalinas

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/mesos/CounterMap.java
+++ b/SingularityBase/src/main/java/com/hubspot/mesos/CounterMap.java
@@ -17,11 +17,11 @@ public class CounterMap<K> {
   private final Map<K, Counter> map;
 
   public CounterMap() {
-    this.map = new HashMap<K, Counter>();
+    this.map = new HashMap<>();
   }
 
   public CounterMap(int initialCapacity) {
-    this.map = new HashMap<K, Counter>(initialCapacity);
+    this.map = new HashMap<>(initialCapacity);
   }
 
   public Map<K, Long> toCountMap() {
@@ -52,7 +52,7 @@ public class CounterMap<K> {
   }
 
   public void incr(K key, long amount) {
-    Counter c = map.putIfAbsent(key, new Counter());
+    Counter c = map.computeIfAbsent(key, k -> new Counter());
 
     c.count += amount;
   }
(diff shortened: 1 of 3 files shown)


SingularityBase/src/main/java/com/hubspot/mesos/CounterMap.java
```diff
@@ -52,12 +52,7 @@ public String toString() {
   }
 
   public void incr(K key, long amount) {
-    Counter c = map.get(key);
-
-    if (c == null) {
-      c = new Counter();
-      map.put(key, c);
-    }
+    Counter c = map.computeIfAbsent(key, k -> new Counter());
```
ssalinas: we are not using the key in order to determine the value here, should use `putIfAbsent` instead of `computeIfAbsent`. Same for the one below

SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
```diff
@@ -133,6 +134,10 @@ public void save(SingularityHostState hostState) throws InterruptedException {
       numTasks.incr(pendingTaskId.getRequestId());
     }
 
+    for (SingularityTaskId cleaningTaskId : taskManager.getCleanupTaskIds()) {
+      numTasks.decr(cleaningTaskId.getRequestId());
```
ssalinas: Do we still want to include cleaning when checking for under provisioned, this will exclude it for both under _and_ over? Trying to make sure we avoid any false alerts if possible

darcatron: I believe so. I'm imagining a request that should have 3 instances, but has 2 running and 1 cleaning. Since we're still counting pending tasks, if one task is pending, then the count is correct, otherwise, it's under provisioned. I think counting cleaning tasks may actually have made us miss potential under provisioned alerts in the past

ssalinas: 👍  cool, let's give this a go in staging/qa then

SingularityBase/src/main/java/com/hubspot/mesos/CounterMap.java
darcatron: small gotcha for putIfAbsent: turns out putIfAbsent returns the old value (null) so computeIfAbsent is what we want here (returns existing or computed) 
bitmoji

ssalinas: 👍  good find, didn't realize that one

ssalinas:
👍 lets get this into stable

ssalinas:
Just verifying the merge conflicts got resolved correctly and will merge this one too 👍

PR #1513, opened 2017-04-20
Write files in subdirectories to splat path

PtrTeixeira:
When the page was loaded the first time, the call to get the files to
the directory wasn't being called with the directory splat. It was
therefore always fetching the files from the top-level directory. When
linked to the top level directory, this wasn't notable, but links into
the subdirectory of the file browser would seem as if the directory was
empty until going back to the root directory.


ssalinas:
Gotta love when one missing variable can ruin the day... LGTM

PR #1514, opened 2017-04-21
Update README.md

lorcanthrope:
Modify the default command for grep that's specified

@ssalinas


ssalinas:
👍 thanks for the update

PR #1515, opened 2017-04-25
Add flag to run scheduled task on deploy

PtrTeixeira:
Adds a field in the deploy, `runImmediately`, which takes a
`SingularityRunNowRequest`. If that field is present, then the scheduled
task (or on-demand task) will be run immediately after deploying.

/cc @ssalinas


ssalinas:
A few things on this one:
- When doing a regular run now, we do a [number of validations](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java#L250-L292) on the run now request. We should imitate this to some extent in the deploy case as well, probably split between validations when we receive the request and when we try to enqueue it in the pending queue
  - If the scheduled job is already running we shouldn't be enqueueing another pending request. I don't think this needs to fail the deploy though. Maybe we can name it something like `attemptToRunImmediately` or find some way of making it clear it is a 'best effort'
  - We should hydrate the run now request with an id if one is not set
  - A number of the `badRequest` checks can be moved to the validator and we can reuse that code to check the run now request on the deploy object if it is present (like the length of runId one)

ssalinas:
+1 for consolidating in the validator. Let's fix the merge conflicts and give this a try in staging

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -319,6 +335,106 @@ private void finishDeploy(SingularityRequestWithState requestWithState, Optional
...
+          message);
+      return Optional.of(pendingRequest);
+    } else if (!request.isAlwaysRunning()
+        && maybeRunNowRequest.isPresent()
+        && request.getInstances().isPresent()
+        && (activeTasks.size() + pendingTasks.size() < request.getInstances().get())) {
+      SingularityRunNowRequest runNowRequest = maybeRunNowRequest.get();
+      pendingType = PendingType.ONEOFF;
+      runId = runNowRequest.getRunId().or(Optional.of(UUID.randomUUID().toString()));
+      message = runNowRequest.getMessage()
```
ssalinas: might be going a bit line-break happy here/above/below in this method ;) can get a bit hard to read

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -265,12 +268,25 @@ private void finishDeploy(SingularityRequestWithState requestWithState, Optional
...
-      // TODO should this override? What if someone has mucked with the pending queue for this deploy ?
-      requestManager.addToPendingQueue(new SingularityPendingRequest(request.getId(), pendingDeploy.getDeployMarker().getDeployId(), deployResult.getTimestamp(),
-        pendingDeploy.getDeployMarker().getUser(), deployResult.getDeployState() == DeployState.CANCELED ? PendingType.DEPLOY_CANCELLED : PendingType.NEW_DEPLOY,
-        deploy.isPresent() ? deploy.get().getSkipHealthchecksOnDeploy() : Optional.<Boolean> absent(), pendingDeploy.getDeployMarker().getMessage()));
+    if (deploy.isPresent() && deploy.get().getRunImmediately().isPresent()) {
+      Optional<SingularityPendingRequest> maybePendingRequest = buildPendingRequest(request,
+          pendingDeploy,
+          deployResult,
+          deploy.get().getRunImmediately());
+      if (maybePendingRequest.isPresent()) {
```
ssalinas: for this and the one below. We should always make sure _some_ pending request gets enqueued for a `!deployable && !oneOff` case. If it doesn't get enqueued it's possible we could miss scheduling the next run of a cron or something like that.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -355,13 +364,117 @@ public SingularityDeploy checkDeploy(SingularityRequest request, SingularityDepl
...
+  private SingularityDeploy checkImmediateRunDeploy(SingularityRequest request,
+                                                    SingularityDeploy deploy,
+                                                    SingularityRunNowRequest runNowRequest,
+                                                    List<SingularityTaskId> activeTasks,
+                                                    List<SingularityPendingTaskId> pendingTasks) {
+    if (!request.isScheduled() && !request.isOneOff()) {
+      throw badRequest("Can not request an immediate run of a non-scheduled / always running request (%s)", request);
+    }
+
+    boolean canRunImmediately =
```
ssalinas: do we want to validate this at the time the deploy is enqueued? We re-check similar conditions at the time the pending request is created when the deploy finishes, and the data could have changed by then. Checking here also causes a few extra zk calls (activeTaskIds/pendingTaskIds) that could possibly be hit on the non-leader instance

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
PtrTeixeira: Just couldn't wait to get started with immutables c:

PR #1516, opened 2017-04-26
Include deploy marker and oldest deploy step in state

ssalinas:
/cc @darcatron

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
@@ -297,11 +297,18 @@ public class StateManager extends CuratorManager {
 
     int numDeploys = 0;
     long oldestDeploy = 0;
+    long oldestDeployStep = 0;
     List<SingularityDeployMarker> activeDeploys = new ArrayList<>();
     final long now = System.currentTimeMillis();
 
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
```diff
@@ -296,9 +297,18 @@ public SingularityState generateState(boolean includeRequestIds) {
...
+      activeDeploys.add(pendingDeploy.getDeployMarker());
+      if (pendingDeploy.getDeployProgress().isPresent() && !pendingDeploy.getDeployProgress().get().isStepComplete()) {
+        long deployStepDelta = now - pendingDeploy.getDeployProgress().get().getTimestamp();
+        if (deployStepDelta > oldestDeployStep) {
+          oldestDeployStep = deployStepDelta;
+        }
+      }
       long delta = now - pendingDeploy.getDeployMarker().getTimestamp();
       if (delta > oldestDeploy) {
         oldestDeploy = delta;
```
darcatron: not sure if I'm following the logic correctly here, but should this be in an `else` clause?

ssalinas: two different deltas. I'm adding additional fields since it is likely some things still use the old `oldestDeploy` one, so still calculating that one as normal

PR #1520, opened 2017-04-28
Deleting request

darcatron:
The main thing here is that I moved the check to re-enqueue the cleaning request outside of the if block `else if (shouldKillTask(cleanupTask, activeTaskIds, cleaningTasks, incrementalCleaningTasks) && checkLBStateAndShouldKillTask(cleanupTask))`

That should allow it to check if it needs to re-enqueue even if the task fails and doesn't clean up properly. Added some tests for this as well


ssalinas:
LGTM

PR #1523, opened 2017-05-01
sync branches with master

darcatron:
@ssalinas thought this might be helpful in getting the other branches easily synced up with master. I believe the push should fail if there are merge conflicts, otherwise I can drop the push command and keep everything local.

### DIFF
--- a/sync_with_master.sh
+++ b/sync_with_master.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+
+git checkout hs_staging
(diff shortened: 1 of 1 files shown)


sync_with_master.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
```
ssalinas: Probably want a `set -e` in here so the script stops on first failure

PR #1526, opened 2017-05-04
Add skip lb removal flag to DeleteRequestRequest

PtrTeixeira:
Add a flag that can be set on the request deletion request to avoid
removing the basepath associated with the service from the load
balancer. The suggested use case is when the application associated with
the service is moving out of Singularity, but will still be running and
should remain load-balanced.

/cc @ssalinas

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
@@ -235,11 +235,10 @@ public class RequestResource extends AbstractRequestResource {
     final String deployId = getAndCheckDeployId(requestId);
 
     checkConflict(!(requestManager.markAsBouncing(requestId) == SingularityCreateResult.EXISTED), "%s is already bouncing", requestId);
-    Optional<Boolean> removeFromLoadBalancer = Optional.absent();
 
     requestManager.createCleanupRequest(
         new SingularityRequestCleanup(JavaUtils.getUserEmail(user), isIncrementalBounce ? RequestCleanupType.INCREMENTAL_BOUNCE : RequestCleanupType.BOUNCE,
-            System.currentTimeMillis(), Optional.<Boolean> absent(), removeFromLoadBalancer, requestId, Optional.of(deployId), skipHealthchecks, message, actionId, runBeforeKill));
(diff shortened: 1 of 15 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -235,10 +235,11 @@ public SingularityRequestParent bounce(String requestId, Optional<SingularityBou
     final String deployId = getAndCheckDeployId(requestId);
 
     checkConflict(!(requestManager.markAsBouncing(requestId) == SingularityCreateResult.EXISTED), "%s is already bouncing", requestId);
+    Optional<Boolean> removeFromLoadBalancer = Optional.absent();
```
ssalinas: can probably just put `Optional.absent()` in the args rather than setting it here

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCleaner.java
```diff
@@ -659,7 +661,7 @@ private void cleanupRequestIfNoRemainingTasks(SingularityTaskCleanup cleanupTask
       requestManager.createCleanupRequest(
           new SingularityRequestCleanup(
               cleanupTask.getUser(), RequestCleanupType.DELETING, System.currentTimeMillis(),
-              Optional.of(Boolean.TRUE), requestId, Optional.<String> absent(),
+              Optional.of(Boolean.TRUE), Optional.absent(), requestId, Optional.<String> absent(),
```
ssalinas: I think we will end up losing our value in the flow here. For a delete, we end up doing something like
enqueue delete cleanup -> delete cleanup is removed and task cleanups are enqueued -> last task cleanup re-enqueues delete cleanup
The value will be there on the first delete cleanup, but we've lost it when re-enqueued (where the actual load balancer delete would happen

darcatron: yeah, it would always be absent (and then default to true it looks like) for every request. You'd prob have to add the request option configuration to the the SingularityTaskCleanup since that's the only living memory of the request so you can do something like `cleanupTask.getRemoveFromLb`

PtrTeixeira: 👍 Missed this in the test because I forgot to actually start a task after the deploy. Will push a fix

ssalinas:
👍  Let's give this one a go in staging

PR #1528, opened 2017-05-05
rephrase flagging host

darcatron:
Thought this was a little misleading. Happy to change if this is still confusing too

@ssalinas


ssalinas:
👍

PR #1530, opened 2017-05-08
Fix for Optional request bodies

### LINKED ISSUE #1525: singularity 0.15.0 NPE on DELETE /tasks/task/<id> with no body
It looks like
https://github.com/HubSpot/Singularity/blob/358f40b116f6f1289fa57c6b2e4980d3935ffcf9/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java#L297

expects an absent `Optional` when there's no body, but instead it gets a null and quickly NPEs:

```
ERROR [2017-05-03 18:46:11,922] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: 347e4c0962cda4bd
! java.lang.NullPointerException: null
! at com.hubspot.singularity.resources.TaskResource.killTask(TaskResource.java:297)
```

Workaround is to send an JSON body with message.

cc @dstryker @pennello

ssalinas:
Seems that after the jersey/dropwizard upgrade the Optional bodies are not coming through correctly and are `null` instead of `absent()` when the body is not present. Updated the impacted endpoints (anything with an `Optional` body) to take the object (nullable) and construct the `Optional` afterwards. Actually made the tests look a bit nicer not wrapping everything in `Optional.of()` too.

I can test this out and try to get a 15.1 out if it looks good.
/cc @darcatron @stevenschlansker @dstryker @pennello @sannessa
/fixes #1525

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/DisastersResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/DisastersResource.java
@@ -107,7 +107,8 @@ public class DisastersResource {
   @POST
   @Path("/disabled-actions/{action}")
   @ApiOperation(value="Disable a specific action")
-  public void disableAction(@PathParam("action") SingularityAction action, Optional<SingularityDisabledActionRequest> maybeRequest) {
+  public void disableAction(@PathParam("action") SingularityAction action, SingularityDisabledActionRequest disabledActionRequest) {
+    final Optional<SingularityDisabledActionRequest> maybeRequest = Optional.fromNullable(disabledActionRequest);
     authorizationHelper.checkAdminAuthorization(user);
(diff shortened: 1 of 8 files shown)


pennello:
Thanks @ssalinas.  The test cleanup is nice, but it's a disappointing the injection is now a little more cumbersome.  I wonder if this is actually intended given the jersey/dropwizard upgrade, or if it's an upstream bug.

ssalinas:
@pennello , as far as I could tell it was never really documented that an `Optional` request body was something that was expected to work (at least I couldn't find it on the dropwizard or jersey docs easily). There is a lot around options `*Param` and the validation annotations, but not much around actual request bodies.

stevenschlansker:
Yeah, I'll second the surprise that the container doesn't do this for you (maybe there's a fix that should go upstream to Jersey?) but obviously I'm in favor of fixing the issue despite uglifying the code :)

PR #1531, opened 2017-05-08
Bump version numbers in documentation

PtrTeixeira:
Some of the documentation still referred to old dependencies, in
particular Java 7 and Dropwizard 0.x. Either removed or bumped these to
refer to current versions.

/cc @ssalinas


ssalinas:
Thanks for this, completely forgot we had that section of docs. LGTM

PR #1538, opened 2017-05-10
Allow immediate run in pending queue with deploy

PtrTeixeira:
Give immediate run requests in the pending queue a separate key, so they
are permitted to be in the pending queue at the same time as a new
deploy for the same requestID / deployID. This means that a pending
deploy will not block an immediate run from being enqueued.

/cc @ssalinas

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
@@ -25,6 +25,7 @@ import com.hubspot.singularity.SingularityCreateResult;
 import com.hubspot.singularity.SingularityDeleteResult;
 import com.hubspot.singularity.SingularityDeployKey;
 import com.hubspot.singularity.SingularityPendingRequest;
+import com.hubspot.singularity.SingularityPendingRequest.PendingType;
 import com.hubspot.singularity.SingularityRequest;
 import com.hubspot.singularity.SingularityRequestCleanup;
 import com.hubspot.singularity.SingularityRequestHistory;
@@ -139,7 +140,12 @@ public class RequestManager extends CuratorAsyncManager {
 
   private String pendingQueueKey(SingularityPendingRequest pendingRequest) {
     SingularityDeployKey deployKey = new SingularityDeployKey(pendingRequest.getRequestId(), pendingRequest.getDeployId());
-    return String.format("%s%s", deployKey.toString(), pendingRequest.getTimestamp());
+    if (pendingRequest.getPendingType() == PendingType.ONEOFF
+        || pendingRequest.getPendingType() == PendingType.IMMEDIATE) {
+      return String.format("%s%s", deployKey.toString(), pendingRequest.getTimestamp());
+    } else {
+      return deployKey.toString();
+    }
   }
 
   private String getCleanupPath(String requestId, RequestCleanupType type) {
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
@@ -133,20 +133,22 @@ public class RequestManager extends CuratorAsyncManager {
     return ZKPaths.makePath(PENDING_PATH_ROOT, new SingularityDeployKey(requestId, deployId).getId());
   }
 
-  private String getPendingPath(SingularityPendingRequest pendingRequest) {
-    String nodeName = pendingQueueKey(pendingRequest);
+  private String getPendingPath(SingularityPendingRequest pendingRequest, boolean forceImmediate) {
+    String nodeName = pendingQueueKey(pendingRequest, forceImmediate);
     return ZKPaths.makePath(PENDING_PATH_ROOT, nodeName);
   }
 
-  private String pendingQueueKey(SingularityPendingRequest pendingRequest) {
+  private String pendingQueueKey(SingularityPendingRequest pendingRequest, boolean forceImmediate) {
     SingularityDeployKey deployKey = new SingularityDeployKey(pendingRequest.getRequestId(), pendingRequest.getDeployId());
     if (pendingRequest.getPendingType() == PendingType.ONEOFF) {
       return String.format("%s%s", deployKey, pendingRequest.getTimestamp());
     } else if (pendingRequest.getPendingType() == PendingType.IMMEDIATE) {
       Optional<SingularityPendingRequest> existingRequest = getPendingRequest(pendingRequest.getRequestId(), pendingRequest.getDeployId());
-      if (existingRequest.isPresent() &&
-          (existingRequest.get().getPendingType() == PendingType.TASK_DONE
-          || existingRequest.get().getPendingType() == PendingType.NEW_DEPLOY)) {
+      boolean markImmediate = forceImmediate
+          || (existingRequest.isPresent()
+              && (existingRequest.get().getPendingType() == PendingType.NEW_DEPLOY
+      || existingRequest.get().getPendingType() == PendingType.TASK_DONE));
+      if (markImmediate) {
         return String.format("%s%s", deployKey, "-immediate");
       } else {
         return deployKey.toString();
@@ -173,7 +175,13 @@ public class RequestManager extends CuratorAsyncManager {
   }
 
   public SingularityDeleteResult deletePendingRequest(SingularityPendingRequest pendingRequest) {
-    return delete(getPendingPath(pendingRequest));
+    SingularityDeleteResult deleteResult = delete(getPendingPath(pendingRequest, false));
+    if (deleteResult == SingularityDeleteResult.DIDNT_EXIST
+        && pendingRequest.getPendingType() == PendingType.IMMEDIATE) {
+      return delete(getPendingPath(pendingRequest, true));
+    } else {
+      return deleteResult;
+    }
   }
 
   public SingularityDeleteResult deleteHistoryParent(String requestId) {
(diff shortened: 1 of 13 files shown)


ssalinas:
Thanks for the additional test, let's give this a go in staging 👍

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -133,20 +133,22 @@ private String getPendingPath(String requestId, String deployId) {
...
     SingularityDeployKey deployKey = new SingularityDeployKey(pendingRequest.getRequestId(), pendingRequest.getDeployId());
     if (pendingRequest.getPendingType() == PendingType.ONEOFF) {
       return String.format("%s%s", deployKey, pendingRequest.getTimestamp());
     } else if (pendingRequest.getPendingType() == PendingType.IMMEDIATE) {
       Optional<SingularityPendingRequest> existingRequest = getPendingRequest(pendingRequest.getRequestId(), pendingRequest.getDeployId());
-      if (existingRequest.isPresent() &&
-          (existingRequest.get().getPendingType() == PendingType.TASK_DONE
-          || existingRequest.get().getPendingType() == PendingType.NEW_DEPLOY)) {
+      boolean markImmediate = forceImmediate
+          || (existingRequest.isPresent()
```
ssalinas: I think this may still be causing us some issues. I've seen a scheduled request or two pop through that never got a pending task in place after a successful deploy, I think due to the two possible deletes and the fact that they could be enqueued between those attempts. We also have a weird bit going on with the `getPendingPath(String requestId, String deployId)` implementation which returns a different version of the path when we call `getPendingRequest`

I'm wondering about the consequences of returning `String.format("%s%s", deployKey, pendingRequest.getTimestamp())` for all cases. Since we sort the pending requests by timestamp, we will still be able to put multiple in the queue. The work then falls to the scheduler to de-dupe and process appropriately. We should check out where else we might be relying on the `EXISTED` response to see if this is possible

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -133,12 +133,15 @@ private String getPendingPath(String requestId, String deployId) {
...
-    String nodeName = String.format("%s%s",
-      new SingularityDeployKey(pendingRequest.getRequestId(), pendingRequest.getDeployId()),
-      pendingRequest.getPendingType().equals(PendingType.ONEOFF) ? pendingRequest.getTimestamp()  : "");
+    String nodeName = pendingQueueKey(pendingRequest);
     return ZKPaths.makePath(PENDING_PATH_ROOT, nodeName);
   }
 
+  private String pendingQueueKey(SingularityPendingRequest pendingRequest) {
+    SingularityDeployKey deployKey = new SingularityDeployKey(pendingRequest.getRequestId(), pendingRequest.getDeployId());
+    return String.format("%s%s", deployKey.toString(), pendingRequest.getTimestamp());
```
ssalinas: Only immediate thing I see here is that we'd need to handle any pending requests that are currently in the queue when singularity starts. Otherwise we will not delete them correctly since we will calculate the wrong path, but we will continually process them since it's just a getChildren call. Should be able to write a simple zk migration for this that reads them all writes to the correct path and deletes the old path on startup

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -214,66 +217,81 @@ public void checkForDecomissions(SingularitySchedulerStateCache stateCache) {
...
-      }
+      Optional<SingularityRequestDeployState> maybeRequestDeployState = deployManager.getRequestDeployState(requestId);
+      Optional<SingularityPendingDeploy> maybePendingDeploy = deployManager.getPendingDeploy(requestId);
+      List<SingularityTaskId> matchingTaskIds = getMatchingTaskIds(stateCache, maybeRequest.get().getRequest(), deployKey);
 
-      final List<SingularityTaskId> matchingTaskIds = getMatchingTaskIds(stateCache, updatedRequest, pendingRequest);
+      List<SingularityPendingRequest> effectivePendingRequests = new ArrayList<>();
+      pendingRequestsForDeploy.sort(Comparator.comparingLong(SingularityPendingRequest::getTimestamp));
+      for (SingularityPendingRequest pendingRequest : pendingRequestsForDeploy) {
+        final SingularityRequest updatedRequest = updatedRequest(maybePendingDeploy, pendingRequest, maybeRequest.get());
```
ssalinas: We should move this earlier on and make sure all methods that use the SingularityRequest object are using this one. i.e. in the old code `getMatchingTaskIds` was also called with the updated one instead of `maybeRequest.get().getRequest()`. Maybe put this up right after the isRequestActive check?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -214,66 +217,81 @@ public void checkForDecomissions(SingularitySchedulerStateCache stateCache) {
...
+          effectivePendingRequests.add(pendingRequest);
+        } else {
+          // Any other subsequent requests are not honored until after the pending queue is cleared.
+        }
 
-      if (numScheduledTasks == 0 && !matchingTaskIds.isEmpty() && updatedRequest.isScheduled() && pendingRequest.getPendingType() == PendingType.NEW_DEPLOY) {
-        LOG.trace("Holding pending request {} because it is scheduled and has an active task", pendingRequest);
-        heldForScheduledActiveTask++;
-        continue;
+        requestManager.deletePendingRequest(pendingRequest);
```
ssalinas: for any pending requests that we are scheduling tasks for, the delete should come _after_ the scheduleTasks call. If we are interrupted between here and scheduleTasks we lose a pending request and could get into an inconsistent state

ssalinas: second thing on this, if we hit the `else` case there, should we really be deleting the pending request? In this particular case if we had a BOUNCE and a NEW_DEPLOY in the queue, we'd take the bounce and end up deleting the new deploy

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -214,66 +217,81 @@ public void checkForDecomissions(SingularitySchedulerStateCache stateCache) {
...
-      totalNewScheduledTasks += numScheduledTasks;
+      int scheduledInstances = 0;
+      for (SingularityPendingRequest effectivePendingRequest : effectivePendingRequests) {
+        RequestState requestState = checkCooldown(maybeRequest.get().getState(), maybeRequest.get().getRequest(), deployStatistics);
+        scheduledInstances += scheduleTasks(stateCache, maybeRequest.get().getRequest(), requestState,
+            deployStatistics, effectivePendingRequest, matchingTaskIds, maybePendingDeploy);
+      }
 
-      requestManager.deletePendingRequest(pendingRequest);
+      totalNewScheduledTasks += scheduledInstances;
```
ssalinas: let's keep the variable as `numScheduledTasks` since we call it tasks not instances everywhere else in this class

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -398,8 +416,30 @@ private void deleteScheduledTasks(final Collection<SingularityPendingTask> sched
     }
   }
 
-  private int scheduleTasks(SingularitySchedulerStateCache stateCache, SingularityRequest request, RequestState state, SingularityDeployStatistics deployStatistics,
-    SingularityPendingRequest pendingRequest, List<SingularityTaskId> matchingTaskIds, Optional<SingularityPendingDeploy> maybePendingDeploy) {
+  private List<SingularityTaskId> getMatchingTaskIds(SingularitySchedulerStateCache stateCache, SingularityRequest request, SingularityDeployKey deployKey) {
```
ssalinas: is there a reason you duplicated this method?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
PtrTeixeira: The `getMatchingTaskIds` method really only cares about the `requestId` and whether the request is long-running, which I don't _think_ can change per pending request. The reason that I'm pushing back here is that if it needs to be run on the updated request, then it needs to be run in the inner loop, rather than the outer one.

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
PtrTeixeira: If only immediate / one-off requests have a different pending path now, do we still need the zk migration?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
ssalinas: Ah that's right we've got the inner loop for the individual pending requests now, 👍

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
ssalinas: We most likely do. This will still change the path of things that might be currently in the pending queue on startup. So we'd need to rewrite those to their new path

ssalinas:
👍 looks to go give this another go with the zk migration in there now

ssalinas:
this has looked good with the new updates, merging. Thanks @PtrTeixeira

PR #1541, opened 2017-05-11
Revert breaking change in enum deserialization

PtrTeixeira:
There appears to have been a change right around Dropwizard 1.0
in how enums are deserialized; my guess is that somehow `@JsonCreator`
cascades and prevents the `FuzzyEnumModule` from being applied properly,
but I'm not really sure. The result was that `SingularityContainerType`
wasn't accepting lowercase `mesos`, `docker` as valid values, which was
a breaking change from before - and in particular, prevented creating
deploys from the UI.

See also dropwizard/dropwizard#1985, which is kind-of related, I think

/cc @ssalinas

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/mesos/SingularityContainerType.java
+++ b/SingularityBase/src/main/java/com/hubspot/mesos/SingularityContainerType.java
@@ -1,5 +1,12 @@
 package com.hubspot.mesos;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+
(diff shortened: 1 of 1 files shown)


ssalinas:
@PtrTeixeira I can't seem to reproduce this one. I tried the same fields via ui in our staging setup and a lower case `docker` was correctly read as the `SingularityContainerType.DOCKER`. Same with a number of other fields. Going to close this one for the moment, if you have more specific examples of failures feel free to reopen

PR #1543, opened 2017-05-17
(WIP) Immutables

ssalinas:
Because plane rides are long... and so is a constructor with 30 arguments...

This PR updates most objects in SingularityBase to use [Immutables](https://immutables.github.io/). What this means in general is that we can write far less code that is more readable for all of these POJOs we need. For a new class, you simply need:
```java
@Immutable
@SingularityStyle
@JsonDeserialize(as = MyPojo.class) // Only if this is going to be written to/from json. Don't need for internal-to-app-olny use
public interface MyPojoIF {
  String getField();
}
```

which will generate the `MyPojo` class complete with builders and all the needed bits for jackson.

The classes can either be written as interfaces named like `MyClassIF` or abstract classes named like `AbstractMyClass`. This naming convention and other settings to use with immutables are all encapsulated in the `@SingularityStyle` annotation.

For the transition and some semblance of backwards compatibility:

- I currently have the code generation set up to produce a public constructor (with all arguments), although I would plan on removing these in future releases
- Any deprecated constructors are removed
- Any non-deprecated constructors that do not have all arguments are available as a static method `of`. For example `new MyClass(arg1, arg2)` becomes `MyClass.of(arg1, arg2)`, although these would also be considered deprecated and it is preferred to use the new builder style from immutables (`MyClass.builder().setField1(arg1).setField2(arg2).build()`)
- All `*Builder` classes are removed. You can access the same methods via `*.builder()` instead (i.e. `new SingularityRequestBuilder()` becomes `SingularityRequest.builder()`)

Current TODO is to straighten out the use of abstract classes that are extended a few times, and finish converting the rest of `com.hubspot.singularity` in SingularityBase. Plan for this would likely be to get it merged in immediately following a release so that externally, as well as internally on our branches we can make any needed changes to users of the client along with the version bump.


PtrTeixeira:
`+2,797 -7,725` 😱  👍

darcatron:
bitmoji

ssalinas:
While this was a good POC, I can see a number of edge cases that would be hard to find in here, mostly stemming from the fact that the immutables are much more strict on what can and can't be null. I'll submit a smaller PR that takes care of the more commonly used objects in the client and the larger objects like SingularityState that desperately need nice builders. The remainder we can continue to update as we go along. Closing this out, but keeping the branch

PR #1548, opened 2017-05-21
Fix markdown formatting

peterlebrun:
Two of the markdown headers were not rendering properly in Github because there was no newline between the anchor tag and the markdown formatting.


ssalinas:
Thanks for the fix 👍

PR #1558, opened 2017-06-01
Remove unnecessary </a>s

matteofigus:
I found this in the docs, so I decided to make a PR. I hope it helps.

Thanks for working on Singularity ❤️

### DIFF
--- a/Docs/reference/api.md
+++ b/Docs/reference/api.md
@@ -232,7 +232,7 @@ Disable a specific action
 
 | Parameter | Required | Description | Data Type |
 |-----------|----------|-------------|-----------|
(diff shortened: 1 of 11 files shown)


ssalinas:
Hmm, interesting. These docs are actually auto-generated by swagger. I wonder if there is a flag we might be missing to get those out of there

matteofigus:
Ok, in that case, feel free to close and I'll reopen if I manage to fix the command to generate them correctly ;)

ssalinas:
👍  Thanks for the PR though 😃 

can attempt to just generate the docs running something like `mvn -Pbuild-swagger-documentation -DskipTests=true -B -q -fae install` . Config is mostly under the `swagger-maven-plugin` section of the plugins in `SingularityService` pom

PR #1560, opened 2017-06-06
separate timeouts for health check and task running

darcatron:
Looks like the max time for a task to get to TASK_RUNNING and pass health checks was set by `killHealthcheckAfterDefaultSeconds`. Since we were previously using `System.currentTimeMillis() - task.getTaskId().getStartedAt()` we used the same clock for both parts.

This separates the checks for the max time it can take to get a task running and then the max time it can take to pass health checks

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
@@ -113,7 +113,7 @@ public class SingularityNewTaskChecker {
   }
 
   private long getKillAfterHealthcheckRunningForMillis() {
-    return TimeUnit.SECONDS.toMillis(configuration.getKillHealthcheckAfterDefaultSeconds());
+    return TimeUnit.SECONDS.toMillis(configuration.getKillTaskIfNotHealthyAfterSeconds());
   }
 
   private int getDelaySeconds(SingularityTask task, Optional<SingularityRequestWithState> requestWithState) {
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
@@ -23,6 +23,7 @@ import com.google.inject.Inject;
 import com.google.inject.name.Named;
 import com.hubspot.baragon.models.BaragonRequestState;
 import com.hubspot.mesos.JavaUtils;
+import com.hubspot.singularity.ExtendedTaskState;
 import com.hubspot.singularity.LoadBalancerRequestType;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
```diff
@@ -368,16 +382,27 @@ CheckTaskState getTaskState(SingularityTask task, Optional<SingularityRequestWit
     return Optional.absent();
   }
 
-  private boolean isOverdue(SingularityTask task) {
-    final long taskDuration = System.currentTimeMillis() - task.getTaskId().getStartedAt();
+  private boolean isHealthcheckOverdue(SingularityTask task) {
+    final long healthcheckDuration = taskManager.getLastHealthcheck(task.getTaskId()).get().getDurationMillis().or(0L);
```
darcatron: I wasn't sure of the best route here since we don't want to use the time the task was launched for health checks. If we reach this logic, that means there is a health check so we can for sure have that, but the duration millis can be absent. If the duration never gets updated, this __could__ lead to the case of a health check running forever

ssalinas: we should be able to use the time since the task entered the running state here

ssalinas:
should be gtg to staging once the comments above are addressed 👍

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
darcatron: which time are you referring to? `task.getTaskid().getStartedAt()` looked like launch time. i couldn't find anything that shows the time is updated once it starts running. I also don't see any other times except for system time

ssalinas: in task history updates, there will be an update with status of `TASK_RUNNING` that has an associated timestamp

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
```diff
@@ -370,21 +381,50 @@ CheckTaskState getTaskState(SingularityTask task, Optional<SingularityRequestWit
...
+        return taskRunningState.get().getTimestamp();
+      }
+
+      LOG.error("Could not find time when task {} reached TASK_RUNNING state", task);
+    } else {
+      LOG.error("Could not find task history for {}", task);
+    }
+
+    return System.currentTimeMillis();
+  }
```
darcatron: had it fall back to the current time if it doesn't find the state info. could run over in this case, but we should be notified by one of those error messages

PR #1561, opened 2017-06-07
Combine offers to schedule tasks more efficiently

ssalinas:
This updates the scheduler to group offers by host and use the total resources available on all offers when scheduling tasks. Especially when using the offer cache, it has been common for us to have 10+ offers in the cache at a time from a single host. For tasks that require larger amounts of resources, this should get them scheduled much more quickly.

As far as I can tell, the mesos master will add up all resources on it's side when a task or tasks are scheduled using multiple offers. So, this code is not currently making an effort to match each task with a portion of the resources available, it is launching all tasks with all offers for the host. It could be a future improvement to save chunks of unused offers so the offer cache could hold on to them for the next run.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityOfferHolder.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityOfferHolder.java
@@ -4,6 +4,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.mesos.Protos;
 import org.apache.mesos.Protos.Resource;
@@ -13,7 +14,6 @@ import org.apache.mesos.SchedulerDriver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.hubspot.mesos.JavaUtils;
 import com.hubspot.mesos.MesosUtils;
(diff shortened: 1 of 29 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityOfferHolder.java
```diff
@@ -25,25 +26,31 @@
...
     this.rackId = rackId;
-    this.offer = offer;
-    this.acceptedTasks = Lists.newArrayListWithCapacity(taskSizeHint);
-    this.currentResources = offer.getResourcesList();
+    this.slaveId = slaveId;
+    this.hostname = hostname;
+    this.offers = offers;
+    this.roles = MesosUtils.getRoles(offers.get(0));
+    this.acceptedTasks = Lists.newArrayListWithExpectedSize(taskSizeHint);
+    this.currentResources = offers.size()  > 1 ? MesosUtils.combineResources(offers.stream().map(Protos.Offer::getResourcesList).collect(Collectors.toList())) : offers.get(0).getResourcesList();
```
darcatron: Wonder if it's possible to have two currentResources for this? One for the individual offers (as originally given by mesos) and one for the collection of same host offers? The check could then try all the single offers and check the combined offers if no single offers were good enough. 

It could help launch more tasks if the smaller offers were used instead of declined. The tough part would be that the combinedCurrentResources would need to be calculated each time before it's checked since some single offers could be utilized for a task in a later iteration.

ssalinas: I think a more general solution might be to sort out which offers to accept/decline/cache when launching the tasks at the end. i.e. if we have an offers of 2,3 and 4 cpus, and matched tasks that use 3 and 3, we can hold on to the 3 cpu offer, and use the other two. No reason to calculate that as we go along since the total resources is the same, but we can certainly do it at the end with the goal of maximizing usage of the current offers.

#lets-do-the-math

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityOfferHolder.java
```diff
@@ -124,7 +124,20 @@ public void launchTasks(SchedulerDriver driver) {
       LOG.trace("Launching {} mesos task: {}", task.getTaskId(), MesosUtils.formatForLogging(task.getMesosTask()));
     }
 
-    Status initialStatus = driver.launchTasks(offers.stream().map(Protos.Offer::getId).collect(Collectors.toList()), toLaunch);
+    // At this point, `currentResources` contains a list of unused resources, because we subtracted out the required resources of every task we accepted.
+    // Let's try and reclaim offers by trying to pull each offer's list of resources out of the combined pool of leftover resources.
+    List<Offer> neededOffers = offers.stream().filter(o -> {
+      List<Resource> remainingAfterSavingOffer = MesosUtils.subtractResources(currentResources, o.getResourcesList());
+      if (MesosUtils.allResourceCountsNonNegative(remainingAfterSavingOffer)) {
+        cache.cacheOffer(driver, System.currentTimeMillis(), o); // TODO: do we need this timestamp to be something specific, e.g. the time at which we got the offer originally?
```
ssalinas: keep in mind here, the offer may already be from the cache to start with, we might need to be returning the offer rather than caching it here

PR #1562, opened 2017-06-08
Retry failed client requests

PtrTeixeira:
In situations when the client fails with certain error codes,
it should be able to retry the request. Ideally, it would
retry the request to a different host so that in the event that
an instance malfunctioned or was removed, the request wasn't
consistently failing.

/cc @ssalinas

### DIFF
--- a/SingularityClient/pom.xml
+++ b/SingularityClient/pom.xml
@@ -22,6 +22,11 @@
     </dependency>
 
     <dependency>
+      <groupId>com.github.rholder</groupId>
+      <artifactId>guava-retrying</artifactId>
+    </dependency>
+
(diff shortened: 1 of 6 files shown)


ssalinas:
The guava retrying looks pretty straightforward. Some things to keep in mind:
- the HttpClient generally also has its own retries built in as well. We should make sure to keep that in mind when configuring how many times/how often to retry.
- `getHost` is currently a random selection. To make this particular case more useful, we could probably update it to exclude a host that has already been tried in the case that a previous host has failed.
- retry attempts and strategy should be configurable. We can put the methods to set them and add defaults in the SingularityClientProvider class

PR #1563, opened 2017-06-08
Scheduled on-demand requests

baconmania:


SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
```diff
@@ -15,15 +15,21 @@
...
 
   @JsonCreator
-  public SingularityRunNowRequest(@JsonProperty("message") Optional<String> message, @JsonProperty("skipHealthchecks") Optional<Boolean> skipHealthchecks,
-      @JsonProperty("runId") Optional<String> runId, @JsonProperty("commandLineArgs") Optional<List<String>> commandLineArgs, @JsonProperty("resources") Optional<Resources> resources) {
+  public SingularityRunNowRequest(@JsonProperty("message") Optional<String> message,
+                                  @JsonProperty("skipHealthchecks") Optional<Boolean> skipHealthchecks,
+                                  @JsonProperty("runId") Optional<String> runId,
+                                  @JsonProperty("commandLineArgs") Optional<List<String>> commandLineArgs,
+                                  @JsonProperty("resources") Optional<Resources> resources,
+                                  @JsonProperty("runAt") Optional<Long> runAt) {
```
ssalinas: I believe Orion api and some others use this method. Until we get something like #1543 merged in, we should keep the deprecated constructor as well

baconmania: Gotcha, good catch.

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -1390,6 +1390,33 @@ public void testRunNowScheduledJobDoesNotRetry() {
...
+            Optional.of(requestedLaunchTime)
+        )
+    );
+
+    scheduler.drainPendingQueue(stateCacheProvider.get());
+
+    SingularityPendingTaskId task = taskManager.getPendingTaskIds().get(0);
+    long runAt = task.getNextRunAt();
+
+    Assert.assertTrue(requestedLaunchTime == runAt);
```
PtrTeixeira: Why `assertTrue(...)` instead of `assertEquals(requestedLaunchTime, runAt)`?

ssalinas:
We should probably set some sane limit on the amount of time in the future an on demand task can be scheduled to catch things like extra zero typos. We can validate this in the method in the `RequestResource` and return a bad request if the criteria is not met. Maybe something like 1 month in advance???

Also, this makes me realize that we currently have no way of cancelling an individual pending task. We can pause which removes all of them, but not just a single one. Can address that in a separate PR, but we would probably want this for on `ON_DEMAND`s at some point.

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
baconmania: Good catch, fix'd

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -297,13 +298,21 @@ public SingularityPendingRequestParent scheduleImmediately(@ApiParam("The reques
...
 
     if (maybeRunNowRequest.isPresent()) {
       message = maybeRunNowRequest.get().getMessage();
       runId = maybeRunNowRequest.get().getRunId();
       skipHealthchecks = maybeRunNowRequest.get().getSkipHealthchecks();
       commandLineArgs = maybeRunNowRequest.get().getCommandLineArgs();
       resources = maybeRunNowRequest.get().getResources();
+      runAt = maybeRunNowRequest.get().getRunAt();
+
+      if (runAt.isPresent() && runAt.get() > (System.currentTimeMillis() + TimeUnit.DAYS.toMillis(30))) {
```
ssalinas: Should have been more specific in my comment. With any hard limits, we always try to make them configurable via parameters somewhere in `SingularityConfiguration`, that way if for some reason we need to override, we can stick it in the yaml and be gtg

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -297,13 +298,21 @@ public SingularityPendingRequestParent scheduleImmediately(@ApiParam("The reques
...
+        throw badRequest("Task launch delay can be at most 30 days from now.");
+      }
+
+      LOG.info("runAt: " + runAt.toString());
```
ssalinas: forgot to remove this?

baconmania:
See https://github.com/HubSpot/Singularity/pull/1565 instead.

PR #1564, opened 2017-06-12
Fix values for the container type field so they deserialize properly.

baconmania:


PtrTeixeira:
See also https://github.com/HubSpot/Singularity/pull/1541 for more context here

ssalinas:
Yeah this is a temporary fix for the ui, but we need to figure out why `FuzzyEnumModule` suddenly stopped working

PR #1569, opened 2017-06-19
Remove extra scheduled entries from pending queue

PtrTeixeira:
There are now - deliberately - cases when multiple scheduled tasks can
enter the pending queue at once. However, scheduled tasks need to be
globally unique - there should only ever be one between the pending
queue and the active tasks list. Currently any duplicates that reach the
pending queue are held there. This instead clears them out when they are
duplicated, relying instead on the task being started on `TASK_DONE`.

/cc @ssalinas

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
@@ -25,11 +25,9 @@ import org.slf4j.LoggerFactory;
 
 import com.codahale.metrics.annotation.Timed;
 import com.google.common.base.Optional;
-import com.google.common.base.Predicates;
 import com.google.common.base.Throwables;
 import com.google.common.collect.HashMultiset;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.ListMultimap;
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -290,6 +290,12 @@ public void drainPendingQueue(final SingularitySchedulerStateCache stateCache) {
           scheduledTasks += scheduleTasks(stateCache, maybeRequest.get().getRequest(), requestState,
               deployStatistics, pendingRequest, matchingTaskIds, maybePendingDeploy);
           requestManager.deletePendingRequest(pendingRequest);
+        } else if (updatedRequest.isScheduled()
+            && (pendingRequest.getPendingType() == PendingType.NEW_DEPLOY
+            || pendingRequest.getPendingType() == PendingType.TASK_DONE)) {
+          // If we are here, there is already an immediate of run of the scheduled task launched. Drop anything that would
```
ssalinas: Is is possible to write a unit test to prove out this particular case? i.e. what happens if the immediate is first vs the NEW_DEPLOY/TASK_DONE comes first in the queue

ssalinas:
Thanks for adding the test, let's give this one a go in staging

PR #1570, opened 2017-06-20
resource usage endpoint

darcatron:
- [x] add some unit tests

Added endpoint to get resource usage data for each tracked request and the cluster as a whole.
Increased the number of points collected form 5 -> 15 and increased the interval between each point (poll frequency x interval) so we have a wider span of data


SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
```diff
@@ -562,6 +566,26 @@ public SingularityState getState(Optional<Boolean> skipCache, Optional<Boolean>
     return Optional.of(response.getAs(SingularityTaskReconciliationStatistics.class));
   }
 
+  public SingularityClusterUtilization getClusterUtilization() {
```
ssalinas: I would follow the format of the other get calls here. You'll see there is a `getSingle` method where the http call building and deserialization is done for you

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -190,4 +206,88 @@ private void updateLongRunningTasksUsage(Map<ResourceUsageType, Number> longRunn
...
+        numRequestsWithUnderUtilizedMemBytes, totalUnderUtilizedCpu, totalOverUtilizedCpu, totalUnderUtilizedMemBytes, avgUnderUtilizedCpu,
+        avgOverUtilizedCpu, avgUnderUtilizedMemBytes, maxUnderUtilizedCpu, maxOverUtilizedCpu, maxUnderUtilizedMemBytes, minUnderUtilizedCpu,
+        minOverUtilizedCpu, minUnderUtilizedMemBytes, totalMemBytesUsed, totalMemBytesAvailable, totalCpuUsed, totalCpuAvailable, now);
+  }
+
+
+  @VisibleForTesting
+  void clearOldUsage(List<SingularityTaskUsage> taskUsages, String taskId) {
+    if (taskUsages.size() + 1 > configuration.getNumUsageToKeep()) {
+      long minMillisApart = configuration.getUsageIntervalMultiplier() * configuration.getCheckUsageEveryMillis();
```
ssalinas: not sure of the purpose for this one, are you trying to make sure to keep a certain time period of data points rather than a count?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -190,4 +206,88 @@ private void updateLongRunningTasksUsage(Map<ResourceUsageType, Number> longRunn
...
+
+          numRequestsWithOverUtilizedCpu++;
+          totalOverUtilizedCpu += overusedCpu;
+          maxOverUtilizedCpu = Math.max(overusedCpu, maxOverUtilizedCpu);
+          minOverUtilizedCpu = Math.min(overusedCpu, minOverUtilizedCpu);
+        } else {
+          includeUtilization = false;
+        }
+
+        if (unusedMemBytes / memoryBytesReserved >= minUnderUtilizedPct) {
```
ssalinas: is there a downside to return all the data values instead of having mins that have to be hit? I feel like we'd want this to be a full report

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -190,4 +206,88 @@ private void updateLongRunningTasksUsage(Map<ResourceUsageType, Number> longRunn
...
+
+      if (maybeDeploy.isPresent() && maybeDeploy.get().getResources().isPresent()) {
+        boolean includeUtilization = true;
+        long memoryBytesReserved = (long) (maybeDeploy.get().getResources().get().getMemoryMb() * SingularitySlaveUsage.BYTES_PER_MEGABYTE);
+        double cpuReserved = maybeDeploy.get().getResources().get().getCpus();
+
+        double unusedCpu = cpuReserved - utilization.getAvgCpuUsed();
+        long unusedMemBytes = memoryBytesReserved - utilization.getMemBytesTotal();
+
+        if (unusedCpu / cpuReserved >= minUnderUtilizedPct) {
```
ssalinas: similar here, why not just report the over/under for everything rather than filtering it down?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -190,4 +206,88 @@ private void updateLongRunningTasksUsage(Map<ResourceUsageType, Number> longRunn
...
+          includeUtilization = false;
+        }
+
+        if (unusedMemBytes / memoryBytesReserved >= minUnderUtilizedPct) {
+          numRequestsWithUnderUtilizedMemBytes++;
+          totalUnderUtilizedMemBytes += unusedMemBytes;
+          maxUnderUtilizedMemBytes = Math.max(unusedMemBytes, maxUnderUtilizedMemBytes);
+          minUnderUtilizedMemBytes = Math.min(unusedMemBytes, minUnderUtilizedMemBytes);
+        } else if (!includeUtilization) {
+          it.remove();
```
ssalinas: not filtering things down also means you'd get to remove this. Generally better practice and more readable in code to not add or remove from the collection your iterating over.

SingularityBase/src/main/java/com/hubspot/singularity/RequestUtilization.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public double getCpuTotal() {
+    return cpuTotal;
+  }
+
+  public int getNumTasks() {
+    return numTasks;
+  }
+
+  public double getAvgMemBytesUsed() {
```
ssalinas: having these getters here means they would be included as json fields. Since they are easily calculated, and not used anywhere else in the code maybe exclude these? (think two extra fields x 3000+ requests all in one json blob)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
darcatron: Yeah, the goal here is to increase the interval between each point (poll frequency * interval) so we have a wider span of data. I thought that would be more telling of usage since a task could be in a bad state (fail, be modified, etc.) for a few consecutive runs. The interval defaults to 3 so that's essentially 45 minutes of data (15 points 3 min apart) rather than 15 minutes (15 points 1 min apart)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
darcatron: I added this min percentage to give us a goal of where we want our utilization to be. It's defaulted to 5% so that means any task using less than 95% of resources allocated would be considered in our report. 

Without a min percent, it'd skew the counts and averages. Every task not using all the requested resources, no matter how small, will be counted as under-utilized. If some task is using 98 mb out of the 100 mb they requested, that seems very reasonable so it didn't make sense to count it as an under-utilized task.

With that said, I'm okay with changing it to consider all tasks if you think that would be more useful

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
darcatron: yeah, i wasn't a fan of the iterator here either, but that was the only way to safely remove the item from the list in the loop. If we decide to keep the min percentage, I could do a second cleanup loop rather than do it within the loop

SingularityBase/src/main/java/com/hubspot/singularity/RequestUtilization.java
darcatron: these are actually used to determine each unused resource:
`long unusedMemBytes = (long) (memoryBytesReserved -
 utilization.getAvgMemBytesUsed());`

I could drop the fields and do an inline calculation though?:
`long unusedMemBytes = (long) (memoryBytesReserved -
 (utilization.getMemBytesTotal() / utilization.getNumTasks()));`

darcatron:
@PtrTeixeira I've fixed some of those tests locally already and am working on getting mine written and the others fixed as well

SingularityBase/src/main/java/com/hubspot/singularity/RequestUtilization.java
PtrTeixeira: Could just do `@JsonIgnore`and keep them on the object, but not the serialization

darcatron:
@ssalinas looking good in qa. okay to get it into prod?

ssalinas:
Going to merge this one. Any further updates can be done in future PRs

PR #1572, opened 2017-06-21
Fix bug when 'maxTotalHealthcheckTimeout' set

PtrTeixeira:
The comparison was flipped in the validator the max total health check
timeout seconds, where when it was set it would reject anything _less_
than that period, rather than anything greater than it. We never noticed
the bug because internally this isn't a field that we set.

/cc @ssalinas


ssalinas:
👍  another that can go to staging

PR #1573, opened 2017-06-21
Bug: Ignore healthcheck config when skipping checks

SegFaultAX:
Fix deployment failures when deploying to a request with no ports
assigned (and therefore no valid value for healthcheck port index) when
a healthcheck URI is supplied.

Example:

```json
{
  "requestId": "my-test-service",
  "healthcheck": {
    "uri": "/health"
  },
  "resources": {
    "cpus": 0.1,
    "memoryMb": 128,
    "numPorts": 0
  }
}
```

This configuration currently fails even when skipHealthchecks is true.
Ideally Singularity should either

1. Error immediately because supplying healthcheck config for a request
with checks disabled is logically invalid, or

2. Ignore healthcheck metadata entirely when the request isn't going to
use it anyway.

This adds the latter solution, but I believe both are sensible options.

### DIFF
--- a/SingularityService/src/test/java/com/hubspot/singularity/data/ValidatorTest.java
+++ b/SingularityService/src/test/java/com/hubspot/singularity/data/ValidatorTest.java
@@ -12,6 +12,7 @@ import com.google.common.base.Optional;
 import com.google.inject.Inject;
 import com.hubspot.deploy.HealthcheckOptions;
 import com.hubspot.deploy.HealthcheckOptionsBuilder;
+import com.hubspot.mesos.Resources;
 import com.hubspot.singularity.RequestType;
 import com.hubspot.singularity.SingularityDeploy;
 import com.hubspot.singularity.SingularityRequest;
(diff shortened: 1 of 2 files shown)


SingularityService/src/test/java/com/hubspot/singularity/data/ValidatorTest.java
```diff
@@ -80,4 +81,35 @@ public void itForbidsHealthCheckStartupDelaysLongerThanKillWait() {
...
+    SingularityDeploy deploy = SingularityDeploy
+        .newBuilder("1234567", "1234567")
+        .setHealthcheck(Optional.of(healthCheck))
+        .setResources(Optional.of(resources))
+        .setSkipHealthchecksOnDeploy(Optional.of(Boolean.TRUE))
+        .build();
+    SingularityRequest request = new SingularityRequestBuilder("1234567", RequestType.SERVICE).build();
+    WebApplicationException exn = (WebApplicationException) catchThrowable(() -> validator.checkDeploy(request, deploy));
+    assertThat((String) exn.getResponse().getEntity())
+        .doesNotContain("Either an explicit port number, or port resources and port");
```
SegFaultAX: This is a rather loose sort of assertion since the test does error, but lower down in `checkDeploy`. Unfortunately that method is complex and fairly poorly tested as is. I did my best to capture a reasonable-ish approximation.

ssalinas:
Thanks for the PR @SegFaultAX . I agree the Validator class could certainly use more tests. In the past we have tended to go with your first solution. On first look it doesn't seem there would be any way to get in a spot where the invalid health check was activated and used. But, I still think error out early would be preferable. It's always easier to make future changes when you can be more certain of the state of the data that is already in the system as well.

ssalinas:
Feedback has gone unanswered and there are a number of merge conflicts now on this. Going to close it out

PR #1574, opened 2017-06-25
Support for gpus

kchaliki:
Hello, I appreciate this is not well tested etc but I figured its mostly a plumbing change and the resource management is already well tested but happy to add whatever is expected


kchaliki:
Clearly a github amateur, I will sort out conflicts

ssalinas:
Thanks for the PR! You are correct that it is mostly plumbing that needs to be added. However, there are a few caveats to be aware of with the gpu support:

- gpus can only be in whole numbers (fraction cpus are allowed, but not gpus)
- when testing this, the `filesystem/linux,cgroups/devices,gpu/nvidia` isolators must be specified on the agent is order for it to properly isolate gpu usage
- There is a capability called `GPU_RESOURCES` that must be added in the `FrameworkInfo` when the framework first registers. This is what will allow us to start receiving gpu offers from the master.

Feel free to ping here if you need any help on the PR. I'm planning on an 0.16 release soon, so this willy likely make it into 0.17 instead

kchaliki:
@ssalinas does this mean gpus should generally be an int? As far as I can tell scalars are doubles in mesos. Can I just let mesos error out per [this](http://mesos.apache.org/documentation/latest/gpu-support)? I'll try and make sense of the rest.

ssalinas:
Whenever possible we should catch errors like that before we interact with mesos. If we already know mesos will error on the task, there is no reason to send it off.

You are correct that we can probably use an `int` for gpus

ssalinas:
Closing this as there are quite a few merge conflicts now. We can revisit gpus at a later date

PR #1576, opened 2017-06-28
Evenly-spread task placement

baconmania:


SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityMachineStatesTest.java
```diff
@@ -333,7 +333,8 @@ public void testFrozenSlaveCanBeDecommissioned() {
 
     saveAndSchedule(request.toBuilder().setSlavePlacement(Optional.of(SlavePlacement.OPTIMISTIC)).setInstances(Optional.of(2)));
```
ssalinas: feel free to change this to GREEDY or SEPARATE so that the frozen slave test isn't relying on calculations form optimistic

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -208,13 +210,23 @@ public SlaveMatchState doesOfferMatch(SingularityOfferHolder offerHolder, Singul
         }
         break;
       case OPTIMISTIC:
-        final double numPerSlave = numDesiredInstances / (double) stateCache.getNumActiveSlaves();
+        Collection<SingularityTaskId> currentlyActiveTasksForRequestClusterwide = stateCache.getActiveTaskIdsForRequest(taskRequest.getRequest().getId());
 
-        final boolean isSlaveOk = numOnSlave < numPerSlave;
+        // If no tasks are active for this request yet, we can fall back to greedy.
+        if (currentlyActiveTasksForRequestClusterwide.size() > 0) {
```
ssalinas: would it make sense to make this more explicit with a == 0 check + early return rather than having to add a comment?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -208,13 +210,23 @@ public SlaveMatchState doesOfferMatch(SingularityOfferHolder offerHolder, Singul
...
+          Set<String> currentHostsForRequest = currentlyActiveTasksForRequestClusterwide.stream()
+              .map(SingularityTaskId::getSanitizedHost)
+              .collect(Collectors.toSet());
+
+          final double numPerSlave = currentlyActiveTasksForRequestClusterwide.size() / (double) currentHostsForRequest.size();
+
+          final boolean isSlaveOk = numOnSlave <= numPerSlave;
+
+          if (!isSlaveOk) {
+            LOG.trace("Rejecting OPTIMISTIC task {} from slave {} ({}) due to numOnSlave {}", taskRequest.getRequest().getId(), slaveId, host, numOnSlave);
```
ssalinas: might want to add the other values you're using in he calculation in here for debugging purposes

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
baconmania: Hmm, so the general pattern in this method is to return only if the `SlaveMatchState` is determined to be something other than `OK`. Otherwise, flow is allowed to trickle down to that final `return SlaveMatchState.OK` statement.

PR #1583, opened 2017-07-05
Track average scheduling delay when accepting tasks.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -373,9 +374,36 @@ private SingularityTask acceptTask(SingularityOfferHolder offerHolder, Singulari
...
+    );
+
+    if (statistics.isPresent()) {
+      SingularityDeployStatisticsBuilder bldr = statistics.get().toBuilder();
+
+      long dueTime = taskRequestHolder.getTaskRequest().getPendingTask().getPendingTaskId().getNextRunAt();
+      long now = System.currentTimeMillis();
+
+      if (bldr.getAverageSchedulingDelayMillis().isPresent()) {
+        long newAverageSchedulingDelayMillis = (bldr.getAverageSchedulingDelayMillis().get() * bldr.getNumTasks() + (now - dueTime)) / (bldr.getNumTasks() + 1);
```
ssalinas: this could get funky along with tracking the avgRunTime. We would be incrementing the numTasks when the tasks were finished, but updating this based on when they were started. Might want to track a separate numTasks for this

ssalinas:
👍

PR #1584, opened 2017-07-07
Multi Cluster Coordination

ssalinas:
This adds a new module, the `SingularityClusterCoordinator`. The idea here is that this service implements the same API as Singularity and can be aware of multiple clusters (data centers) each with their own Singularity+Mesos+ZK. The coordinator's job is to then consolidate results (for things like listing slaves/racks/requests) or route to the correct dataCenter (based on request id). The UI can then be pointed to the proxy and everything will look as if it is in one large cluster.

Some simplifying assumptions for v1 of this:
- Request IDs are unique across all clusters as are rack IDs. The proxy will throw an error if it find a duplicate. Currently this would be open to a race condition of the same request being created in two places at the same time. But the cross-data-center coordination of proxy instances won't be tackled in V1
- The proxy has a concept of the 'default' data center. All requests for things like user information/settings and any read request where no data center can be determined will be routed there. Writes where the data center cannot be determined will error

TODOs:
- [x] Scaffold out the api in the new module, attempting to be as d.r.y. as possible
- [x] Finish implementation of the proxy code
  - [x] Implement routing by request id
  - [x] Implement merging of results from multiple sources
  - [x] determine for each endpoint if it should be a merge or a route
- [x] Determine an identifier to force a request to one data center or another and implement its usage in the SingularityClient
- [x] Determine if it makes sense for the proxy to serve up the ui (probably does)
- [x] Docs

/cc @tpetr @baconmania @darcatron


darcatron:
:ship:

darcatron:
:ship:

PR #1586, opened 2017-07-11
Refactor tailer instance selection dropdown

kwm4385:
The instance select dropdown has been refactored into a class and to use react-bootstrap components.

This also fixes two issues:

- The dropdown will not close when selecting or deselecting an instance, now multiple can be checked without reopening the dropdown.
- A Select All option has been added which will either select all instances or deselect all but the first instance (at least one instance always needs to be selected or the tailer UI will break).

Caveats: 

- The dropdown will still close when transitioning between one instance selected and greater than one. This is because the header needs to re-render with the appropriate buttons for tailing a single instance.

### DIFF
--- a/SingularityUI/app/components/logs/NewTasksDropdown.jsx
+++ b/SingularityUI/app/components/logs/NewTasksDropdown.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
-import classNames from 'classnames';
+import Checkbox from 'react-bootstrap/lib/Checkbox';
 
-class NewTasksDropdown extends React.Component  {
+class NewTasksDropdown extends React.Component {
 
   renderTaskItems() {
     if (!this.props.ready || !this.props.runningTasks) {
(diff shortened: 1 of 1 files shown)


ssalinas:
Nice, looks pretty smooth in staging. Isn't an easy way to get around the re-render when going to/from 1 instance. Not worried about that one for now.

Only question is around the limit on selecting. I think the tailer limits to no more than six instances tailed at once, are we enforcing that in any way when clicking select all? (i.e. selects only the first 6 or something?)

kwm4385:
There doesn't appear to be anything stopping me from tailing more than 6 instances in the current implementation. I know it's not very useful from a readability/performance standpoint though. However it could be more confusing on the user's end to get capped at a seemingly arbitrary number (think about those ultrawide early adopters!). Also if all Select All does it toggle the first six instances I think it loses much of it's usefulness.

ssalinas:
Fair enough on usability. Mostly don't want to crash anyones browser from accidentally trying to tail 30 instances or something.

kwm4385:
True. How about we take a middle ground and cap it at the first 8 instances when selecting all? I'm able to open that many without any immediate performance issues in my browser.

ssalinas:
Sounds like a plan 👍

PR #1587, opened 2017-07-11
Fix key error on requests page

kwm4385:
Removes an unnecessary `<div>` around `FormModal.INPUT_TYPES.MULTIINPUT` to stop React complaining about the `key` attribute missing at the top level.


ssalinas:
👍 , must have previously been multiple elements there to necessitate the `div` or something

PR #1589, opened 2017-07-12
Keep the state cache up to date within runs of the poller

ssalinas:
Right now old pending tasks aren't deleted properly for a scheduled request if there is a new deploy pending request immediately followed by an immediate pending request within the same run of the poller. The state cache does not contain the new pending task from the first pending request, so it is not deleted during the second. This updates the `schedule` method to keep the state cache up to date within a run of the poller.

/cc @darcatron


ssalinas:
@baconmania going to close this one in favor of your upcoming changes to exclusively use the leader cache and remove the state cache

PR #1590, opened 2017-07-12
package.json tune-up

kwm4385:
Part of ongoing effort to bring our dependencies up to date and optimize by removing unused libraries on combining them in cases where we are using more than one for the same purpose.

There are several cases where deps are not updated due to more extensive code refactoring required (these will require their own PRs).


kwm4385:
Gonna leave this in staging for a bit and keep my eyes on sentries.

PR #1593, opened 2017-07-14
More flaky tests

ssalinas:
slowly combating things that break the travis build...

Changed these over to use conditions with timeouts so the build will still pass on a system where it may take more than the previous 50ms pause to get things going.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,16 @@ language: java
 jdk:
   - oraclejdk8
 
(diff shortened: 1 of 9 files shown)


ssalinas:
FYI @darcatron @baconmania I've added a hard time limit of 30 seconds on all tests. This should help us weed out the one test that is occasionally hanging (think it is health check related) as well as forcing us to clean up the test framework if things get too slow 😉

PR #1595, opened 2017-07-14
Only write to the leader cache when it's active.

baconmania:


ssalinas:
Ideally these should never happen, but good to be consistent with the other write actions in the cache, LGTM 👍

PR #1592, opened 2017-07-14
Resource Usage UI

kwm4385:
## New Features
- Cluster-level utilization data added to `/utilization` page (formerly `/slave-usage`). Max over/under CPU and memory numbers are linked to the corresponding request. Colored sections of the breakdown link to the appropriate filter on the Requests page.
local hubteamqa com-3334-singularity-ui-utilization
- Resource Usage section on Request Detail page added underneath Task History. Will be shown open by default if CPU has been over-allocated in the past 24 hours.
image
- Over/under CPU/memory filters on the Requests page allow searching of requests based on resource utilization.

## Modifications
- `/slave-usage` page has been renamed to `/utilization`. The tiles within the Slave Health section have had their details replaced by popovers to eliminate them being covered by the tooltip component. Overall layout was improved and code refactored to better align with project standards.

## TODO:
- [x] Add filters for utilization levels on Requests page
- [x] Display 24-hr total CPU/mem stats on utilization page

PR #1596, opened 2017-07-17
Fix check if redux devtools exist

kwm4385:
Fixes page not loading when the extension isn't installed/updated.


ssalinas:
👍

kwm4385:
The change that originally broke this was made in https://github.com/HubSpot/Singularity/pull/1592 so I'm baking the change into that PR from now on and closing this.

PR #1597, opened 2017-07-19
Also relocate guava-retrying in SingularityClient

ssalinas:
The interactions of unshaded retryer with shaded guava were causing issues. Shading retryer as well in SingularityClient

/cc @stevegutz


stevegutz:
I didn't get an email when you tagged me initially for some reason. LGTM

PR #1600, opened 2017-07-21
Report disk usage

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -113,6 +113,11 @@ public SingularityTask buildTask(SingularityOfferHolder offerHolder, List<Resour
 
     if (desiredTaskResources.getDiskMb() > 0) {
       bldr.addResources(MesosUtils.getDiskResource(desiredTaskResources.getDiskMb(), requiredRole));
+    } else if (MesosUtils.getDisk(offerHolder.getCurrentResources(), Optional.absent()) >= 1.0) {
+      // If this offer contains 1MB of disk resources, claim it to enable disk usage reporting.
+      // This is just a temporary hack to enable disk usage reporting where we can, not an actual way to match task disk requirements to offers.
```
ssalinas: Before we roll this out anywhere else, it would probably be better to set something more like a configurable default disk (similar to the default mem/cpus). If someone were to have disk isolation enforced by default, this would be very problematic

ssalinas:
🚢

PR #1601, opened 2017-07-21
(Mesos 1) Better backpressure and abort logic

ssalinas:
Two things in this PR:

- We weren't setting up the appropriate backpressure for the mesos master events observable. Now we will buffer up to the configured number of events. If the buffer overflows we will bail. By default this is set to 5000, which should be plenty to handle the surges of events from most reconciliation calls
- Smoother aborts without the need for SingletonCloser class. While it closed everything, we were closing the zk connection before the server was appropriately shut down, causing 'instance not started' exceptions for a few requests. With previous changes in place, closing the leader latch will allow the scheduler to shut down gracefully, so I am only explicitly running that step before shutting down the server

PRed to mesos_1 branch, will merge once this is also in hs_qa


ssalinas:
changes are in hs_qa, merging this into the mesos_1 branch

PR #1603, opened 2017-07-27
Reorganize request group view

kwm4385:
Switches to using a side nav instead of tabs for better usability with large request groups. Also makes the feature more noticeable. 

## Desktop

image

## Mobile

local hubteamqa com-3334-singularity-ui-group-kafkaopssignals iphone 6 plus

## Menu
(Also has `Apply to all requests in group` tooltip)

image


kwm4385:
Resolves https://git.hubteam.com/HubSpot/PaaS/issues/2543

ssalinas:
I definitely like the side menu layout here. Two questions/comments:

- Does it collapse? (Not sure if we need it to or not since the width is about the same after taking the margins away)
- The current UI will show the error banner if one request in the group is not found or is failed on fetch. Would it be possible to maybe display this status in the sidebar as opposed to having the warning banner? Something like a different color/tooltip/etc if we couldn't fetch the data for that request?

kwm4385:
- The sidebar doesn't collapse. I don't think it's needed since the request section still has the same available width.

- I added a label to the menu if a request has been deleted: 

image

The only caveat is that it won't show up until the user selects it for the first time. Otherwise we'd have to preload all the requests, which slows things down significantly. Not sure if that's a good tradeoff.

ssalinas:
I think that's an ok tradeoff. Just to clarify, by 'load all the requests', your talking about individual calls for each in the group, not the /requests endpoint right?

kwm4385:
Correct, I mean the request details for each.

ssalinas:
👍

PR #1609, opened 2017-08-14
Use HostAndPort#getHostText instead of HostAndPort#getHost

stevegutz:
`HostAndPost#getHost` replaced `HostAndPort#getHostText` in Guava 20 and `HostAndPort#getHostText` was removed in Guava 22. This PR should have no semantic effect on your code. [Java doc](http://google.github.io/guava/releases/21.0/api/docs/com/google/common/net/HostAndPort.html#getHostText--)

I had to upgrade to Guava 20.0 to get the new functionality and if I go to 21 we might have trouble because the version of curator this is on is not compatible with that version. This codepath should never be hit outside of `SingularityService` and `EmbedSingularityExample`, so don't rush to merge this if you're worried about the version change.

@ssalinas @baconmania @darcatron


ssalinas:
🚢

stevegutz:
🎄

PR #1610, opened 2017-08-15
Highlight new files

kwm4385:
Users want a way to determine if a file is currently being written to, especially in cases where downloading an incomplete file such as a heap dump would be undesirable. 

image

The task file browser will highlight files that have been modified in the last 60 seconds.

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskFileBrowser.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskFileBrowser.jsx
@@ -71,7 +71,7 @@ function TaskFileBrowser (props) {
   }
 
   function recentlyModifiedTooltip() {
-    return <Tooltip id="tooltip">This file is currently being written to.</Tooltip>;
+    return <Tooltip id="tooltip">File is currently being written to</Tooltip>;
   }
 
   return (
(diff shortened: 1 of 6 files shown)


SingularityUI/app/components/taskDetail/TaskFileBrowser.jsx
```diff
@@ -64,18 +70,33 @@ function TaskFileBrowser (props) {
...
+    return <Tooltip id="tooltip">File is currently being written to</Tooltip>;
+  }
+
   return (
     <div>
       <Breadcrumbs items={pathItems} />
       <UITable
         data={getFiles() || []}
         keyGetter={(file) => file.name}
+        rowClassName={({mtime}) => { return isRecentlyModified(mtime) ? 'bg-info-light' : null; }}
```
andyhuang91: This will only update if something else causes this component to re-render. Is this the behavior that you want?

SingularityUI/app/components/taskDetail/TaskFileBrowser.jsx
```diff
@@ -64,18 +70,33 @@ function TaskFileBrowser (props) {
...
         emptyTableMessage="No files exist in this directory"
         defaultSortBy="name"
+        striped={false}
       >
         <Column
+          className="icon-column"
+          label=""
+          id="icon"
+          key="icon"
+          cellData={(file) => isRecentlyModified(file.mtime) &&
```
andyhuang91: Because this depends on the current time, you should evaluate this once at the beginning of the render function instead of calling it 3 times.

SingularityUI/app/components/taskDetail/TaskFileBrowser.jsx
kwm4385: The app globally refreshes this data every 60 seconds, and when the window regains focus, so this shouldn't get out of sync.

PR #1612, opened 2017-08-23
Display global message if maxTaskLag > 3min

kwm4385:
screen shot 2017-08-23 at 11 46 05 am


kwm4385:
Future note: We'll have to gate this behind a configuration option for the UI so it doesn't show up to open source users without email notifications set up.

ssalinas:
🚢

PR #1613, opened 2017-08-29
Serve UI configuration via SingularityService.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/resources/StateResource.java
```diff
@@ -56,4 +59,11 @@ public SingularityState getState(@QueryParam("skipCache") boolean skipCache, @Qu
   public Optional<SingularityTaskReconciliationStatistics> getTaskReconciliationStatistics() {
     return stateManager.getTaskReconciliationStatistics();
   }
+
+  @GET
+  @Path("/ui-configuration")
+  @ApiOperation("Retrieve information about the deployed UI configuration")
+  public UIConfiguration getUIConfiguration() {
```
ssalinas: When we create the index view, there is a slightly more hydrated version of ui configuration that gets sent. We would probably want a new pojo that mimics that. It includes things like shell commands/etc from different ares of singularity configuration

ssalinas:
@baconmania I made a few updates to serve up the full IndexView, but as json. This is the entire context needed to render out the index.html normally and should have all the info we need. @kwm4385 can do the ui sections to consume it in a future pr, then we can start to trim down the information we need to send to the ui

ssalinas:
Were a bunch of changes in these files since this was opened. Going to close for now and we can come back to this later

PR #1615, opened 2017-09-05
Add visibility around getChildren() calls.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/data/CuratorManager.java
```diff
@@ -123,6 +123,7 @@ protected boolean exists(String path) {
   }
 
   protected List<String> getChildren(String root) {
+    LOG.debug("Preparing to call getChildren() on {}", root);
```
tpetr: it this information still useful for us? If yes: I'd suggest making it `trace()` instead (it's pretty verbose). If not: please delete.

PR #1616, opened 2017-09-05
Remove swagger jackson version override

stevegutz:
It seems as though this was added when the Jackson version was very old and it should no longer be necessary, but I don't really know how to confirm that myself.

This came up in https://github.com/HubSpot/basepom/pull/20

@ssalinas @tpetr @darcatron @baconmania @jhaber


ssalinas:
🚢

PR #1626, opened 2017-10-06
Expose instance counts for singularity requests

ssalinas:
This adds an endpoint that exposes the counts of tasks that are active/pending/healthy/unhealthy. Debating changing it to be the list of task ids rather than just the counts so we could then use it to possibly display which tasks are healthy in the ui.

/cc @darcatron


ssalinas:
Also fyi @kwm4385 , this would give us information for the request detail page so we could display which of the active tasks are currently healthy. It would be a single endpoint that could replace the pending/cleaning/active calls

darcatron:
worked fine on my local testing against qa 🚢

PR #1628, opened 2017-10-06
Preserve instance id order when scaling down

ssalinas:
Example situation:
- Tasks with instance number 1,2,4,5 are active and there is a single pending task
- request is scaled down to 4
- pending task is deleted and no tasks are cleaned, remaining instances are 1,2,4,5

This PR updates the scheduler to preserve the instance numbers in order since a number of systems depend on them being contiguous. This causes slightly more task churn in a few cases in return for consistency. If the task churn is an issue, we can always put it behind a configuration flag.

tests added, haven't added the config flag just yet

@darcatron


darcatron:
🚢

PR #1630, opened 2017-10-11
Mesos backpressure

ssalinas:
Diff will look nicer once the other mesos_1 PR is merged. Relevant commit is https://github.com/HubSpot/Singularity/commit/d15af0f7b0148388ae6bc5ba5a1cb61e59227eab

Moves the backpressure to the mesos client itself now that 0.2.0 of mesos-rxjava allows us to set backpressure behaviors. Also doubles the default event buffer size to handle large influxes of status updates from reconciliation

/cc @darcatron @baconmania

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
@@ -45,6 +45,7 @@ import com.mesosphere.mesos.rx.java.SinkOperations;
 import com.mesosphere.mesos.rx.java.protobuf.ProtobufMesosClientBuilder;
 import com.mesosphere.mesos.rx.java.util.UserAgentEntries;
 
+import rx.BackpressureOverflow;
 import rx.Observable;
 import rx.subjects.PublishSubject;
 import rx.subjects.SerializedSubject;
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
```diff
@@ -146,7 +147,14 @@ private void connect(URI mesosMasterURI, FrameworkInfo frameworkInfo, Singularit
...
         .mesosUri(mesosMasterURI)
-        .applicationUserAgentEntry(UserAgentEntries.userAgentEntryForMavenArtifact("com.hubspot.singularity", "SingularityService"));
+        .applicationUserAgentEntry(UserAgentEntries.userAgentEntryForMavenArtifact("com.hubspot.singularity", "SingularityService"))
+        .onBackpressureBuffer(
+            scheduler.getEventBufferSize(),
+            () -> {
+              String message = String.format("Overflow of event buffer (%s), singularity could not keep up!", scheduler.getEventBufferSize());
+              scheduler.onUncaughtException(new EventBufferOverflowException(message));
+            },
+            BackpressureOverflow.ON_OVERFLOW_ERROR);
```
baconmania: So just to confirm, if the set of pending events overflows the max buffer size, we get a top-level exception and abort?

ssalinas: yes, it'll cause an abort if the buffer overflows. I felt that was more in line with our idea of keep state consistent than dropping things. There is an alternative for using an unbounded buffer, but I figured it'd be better to abort with an exception than to oom

baconmania: Agreed.

baconmania:
🚢

PR #1631, opened 2017-10-11
Support subdirectories in the s3 uploader

ssalinas:
Adds a field in the upload metadata you can specify to have the uploader check all subdirectories for matching files as well


baconmania:
🚢

PR #1632, opened 2017-10-16
More lenient OPTIMISTIC placement.

baconmania:
This change sets the scheduler up to consider both pending *and* active
tasks when deciding what the optimal number of tasks is on each machine.


ssalinas:
Looks right to me, would be worth a quick unit test just to prove it out + make sure we don't break it in the future

baconmania:
Closing this in favor of https://github.com/HubSpot/Singularity/pull/1633.

PR #1633, opened 2017-10-16
Make the leniency of OPTIMISTIC tunable.

baconmania:
This changes the OPTIMISTIC placement strategy to become increasingly lenient as the number of pending tasks for a given request grows. That leniency can be tuned by a configurable coefficient.


ssalinas:
🚢

PR #1634, opened 2017-10-19
Catch errors in FetchRequestArgHistory

ssalinas:
This is an optional bit to fill in the args for a task, we can catch errors for this endpoint. There error banner is more worrying than helpful in the run now modal.

/cc @kwm4385


kwm4385:
🚢

PR #1635, opened 2017-10-20
Webhook auth

ssalinas:
This adds a way for Singularity to use an external system to provide authorization via json web tokens. Singularity is told where to look for a token, and given a url to check if that token is valid. It will cache responses for a certain time so that the validation sytem is not overwhelmed.

This follows a similar setup to how kubernetes does their external jwt auth.

TODOs:
- [x] Update the UI to extract the value of a cookie and set it as the auth header when configured to do so

/cc @baconmania @darcatron


tpetr:
Excited to see us tackling this, thanks! Copying Kubernetes' webhook strategy for auth sounds 👍. I'd caution against using cookies as Singularity's sole authentication mechanism, though -- it opens you up to CSRF attacks (and implementing CSRF tokens to guard against this is kind of a pain). Instead, you could have Singularity use the `Authorization` header for authentication, and make SingularityUI set that header based on the contents of your desired authentication cookie.

We might also want to consider supporting multiple authentication strategies to ease our migration off of `HEADER_PASSTHROUGH`.

tpetr:
Also, I know this is a super annoying nitpick (so apologies in advance), but I don't think there is anything JWT-specific about this approach -- Singularity is delegating the auth to the other service and isn't actually doing anything with the auth token at all. What do you think about naming it something a little more generic, like `WEBHOOK`?

ssalinas:
Yeah, multiple strategies was definitely on my mind for easing migration. You're correct that it isn't jwt specific. I had first started writing it with the thought of two modes, one external system, and one where singularity does the jwt portion (for outside users of Singularity to have built-in auth) and hadn't renamed the classes.

I can move that portion to the UI to grab the token, couldn't remember what our other UIs were doing there. Thanks for the feedback!

ssalinas:
@tpetr addressed the backend part of the feedback. The authenticators are now bound using a multibinder, so we'll keep trying the configured ones until one works, or we run out of ones to try. Still need to do the UI portion.

ssalinas:
As an FYI I've changed quite a bit here. The previous way we were doing auth was not very flexible and there was some weirdness around injecting the Optional<SingularityUser> into each of the resources (i.e. an absent user ended up as a 500 ProvisionException, not a 401).

So, I've moved us to a model of using ContainerRequestFilters and actual dropwizard authenticators. The `SingularityMultiMethodAuthenticator` allows us to put in any list of possible `SingularityAuthenticator` implementations, and it will attempt each in order to provide a SingularityUser. This would make it very easy to eventually add something like basic auth (or any other auth method for that matter).

The majority of changes to the code come from the fact that we are now using jersey's annotated injectables rather than a guice injectable at the resource class level. In short, instead of a Optional<SingularityUser> in each resource class, we now have an argument of `@Auth SingularityUser user` in each resource method.

The user is also no longer Optional. In cases where auth is disabled a default user is returned (with a flag that it is not an authenticated user. This frees us up to do more interesting things on the backend, not having to worry if the user being passed around is absent or not.

TODO:
- [ ] DOCS!!!!

SingularityService/src/main/java/com/hubspot/singularity/SingularityAuthModule.java
```diff
@@ -1,27 +1,33 @@
...
+    for (SingularityAuthenticatorClass clazz : getConfiguration().getAuthConfiguration().getAuthenticators()) {
+      multibinder.addBinding().to(clazz.getAuthenticatorClass());
+    }
+
+    binder.bind(SingularityAuthFeature.class);
+    binder.bind(SingularityMultiMethodAuthenticator.class);
+    binder.bind(SingularityAuthDatastore.class).to(getConfiguration().getAuthConfiguration().getDatastore().getAuthDatastoreClass());
     binder.bind(SingularityAuthorizationHelper.class).in(Scopes.SINGLETON);
+
+    getEnvironment().jersey().register(SingularityAuthFeature.class);
```
stevegutz: Do you need this? I think that binding it should be enough.

ssalinas: I can try it. I was under the impression i needed both

ssalinas: you are correct, not needed. Now removed

ssalinas:
Auth is working as intended here, will work on the client portion in a separate PR

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUser.java
```diff
@@ -2,33 +2,49 @@
...
+public class SingularityUser implements Principal {
   private final String id;
   private final Optional<String> name;
   private final Optional<String> email;
   private final Set<String> groups;
+  private final boolean authenticated;
+
+  public static SingularityUser defaultUser() {
+    return new SingularityUser("singularity", Optional.absent(), Optional.absent(), Collections.emptySet(), false);
+  }
```
tpetr: worth masking this a const?

SingularityService/src/main/java/com/hubspot/singularity/auth/datastore/SingularityAuthDatastore.java
```diff
@@ -1,9 +1,10 @@
...
 
-import com.google.common.base.Optional;
+import java.util.Optional;
+
 import com.hubspot.singularity.SingularityUser;
 
 public interface SingularityAuthDatastore {
   Optional<SingularityUser> getUser(String username);
-  Optional<Boolean> isHealthy();
+  com.google.common.base.Optional<Boolean> isHealthy();
```
baconmania: Do we want to mix Google and Java `Optional`s in the Singularity codebase?

ssalinas: Singularity is still on guava optionals, new dropwizard uses java util. It would be a breaking change in the client to move everything over. We need to do it at some point. ButI'm not sure this PR is the place yet

baconmania:
🚢

PR #1636, opened 2017-10-20
New dashboard in Singularity + other UI improvements

ssalinas:
@kwm4385 , this is the WIP PR we can build upon for creating user and group based views within Singularity. I have created a separate dropwizard resource with new pojos that will aggregate all data the UI needs for certain pages in a single call. We may need to do some extra bits around things like history paging and such, but we can at least get the basic details of active elements all in a single call, with some extra details like instance health included.

My goal with this would be to replace the current dashboard page with something smarter, generally eliminating the need for the requests page, as well as adding some needed details like instance health in the task detail page.

We should also try to eliminate some extraneous calls we tend to make on a number of pages like:
- fetching /requests on _every_ page
- fetching all request groups on _every_ page
- fetching the user occasionally multiple times per page

Feel free to comment on anything you might specifically want the backend to serve up all together. I'll add more detail here as I flush things out


ssalinas:
On a good suggestion from @baconmania , going to not use the specific endpoints, but instead refactor the current requests/etc ones a bit to better fit the needs of the ui (behind filters/params/etc)

ssalinas:
@kwm4385 I put together a quick 15-minute version of a new dashboard, mostly so that the logic is all written out for you to see the new data the backend is providing for you (and because I got angry at auth and wanted to take a break from it...).

Ordering by default is based on requests that have most recently had something interesting happen, and the list of requests is filtered (by the backend) to ones that are relevant to the user.

screen shot 2017-10-24 at 4 48 29 pm

`Tasks` expands/collapses when clicked so we can show more/less data without cluttering things

screen shot 2017-10-24 at 4 51 17 pm

ssalinas:
Some updates to the new UI stuff:
Dashboard has slimmed down a bit, and non-long-running tasks will now include a recently started/finished task as recent activity:
screen shot 2017-11-09 at 12 26 10 pm
With tasks expanded...
screen shot 2017-11-09 at 12 26 35 pm

Icons replace text for request types and statuses to save space:

screen shot 2017-11-09 at 12 26 56 pm

Don't show full task ID on the request detail page to save space:

screen shot 2017-11-09 at 12 26 48 pm

kwm4385:
Some thoughts on the current iteration:
 - Do people really need to see the task table on this page? If not, linking `Tasks - 1 Healthy` etc. to the request page might be enough.
 - Can we use the same icons as Orion for request types for more consistency? They can be found in `OrionUI/app/images/icons/`.
 - We probably want some type of title and subheader on the page so people know what they're looking at.
 - Consider moving the `Recent activity` column first (after the star) if that's the most important, unique information in this view. 
- Shortened task ids everywhere is 💯 
- What's the reasoning behind the external link icon next to the instance id? Typically this is used to signify the link will open in a new tab and/or leave the current app.

kwm4385:
Just one final thought here

image
Should we have icons for Pending and Cleaning as well?

cc @ssalinas

ssalinas:
Yeah, I can add two more. They are actually separate concepts, not separate states on the backend which is why i didn't at first (i.e. you'll never see a pending/cleaning in the same list as active/paused/cooldown)

ssalinas:
Extra icons added now @kwm4385

kwm4385:
I think this is in a good state to move forward to stable. We'll keep an ear to the ground for user feedback and make further tweaks based on that. 🚢

PR #1642, opened 2017-11-02
Shade protobuf for SingularityClient/SingularityBase

ssalinas:
Systems running protobuf 2.5 were having issues since mesos 1.x pulls in protobuf 2.6.1

@stevegutz

### DIFF
--- a/SingularityBase/pom.xml
+++ b/SingularityBase/pom.xml
@@ -93,8 +93,8 @@
             </filters>
             <relocations>
               <relocation>
-                <pattern>com.google.protobuf</pattern>
-                <shadedPattern>com.hubspot.singularity.shaded.com.google.protobuf.twosixone</shadedPattern>
+                <pattern>org.apache.mesos</pattern>
+                <shadedPattern>com.hubspot.singularity.shaded.org.apache.mesos.oneonetwo</shadedPattern>
               </relocation>
             </relocations>
           </configuration>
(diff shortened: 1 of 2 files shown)


stevegutz:
Looks reasonable to me. For things like protobuf its also common to add the version to the package name, eg `com.hubspot.singularity.shaded.com.google.protobuf.twosixone` or whatever.

Is shading just the client is enough? It doesn't leak into models or anything?

ssalinas:
Good point, might need it in SingularityBase as well. I can add that

ssalinas:
Ended up being easier to shade mesos completely here

baconmania:
🚢

ssalinas:
There are a few more steps we need to take to solve the protos issues. This is a step in the right direction, so going to merge this and open an additional PR rather than continuing to work off this branch

SingularityBase/pom.xml
```diff
@@ -67,6 +67,57 @@
...
+                <artifact>com.google.guava:guava</artifact>
+                <excludes>
+                  <exclude>com/google/common/base/Optional.class</exclude>
+                </excludes>
+              </filter>
+            </filters>
+            <relocations>
+              <relocation>
+                <pattern>org.apache.mesos</pattern>
+                <shadedPattern>com.hubspot.singularity.shaded.org.apache.mesos.oneonetwo</shadedPattern>
```
stevegutz: I don't think think that this will solve your problem. The issue isn't using an unshaded mesos, it's that mesos is using an unshaded protobuf that conflicts with what we want. So you want to rewrite the protobuf classes inside mesos, not mesos itself.

PR #1643, opened 2017-11-03
0.18.0 release docs and mesos 1 upgrade docs

ssalinas:


ssalinas:
Added docs about the removal of the mesos dep from SingularityBase as well as basic docs on setting up webhook auth.

@baconmania can you give this a quick read for sanity at some point?

baconmania:
🚢

PR #1644, opened 2017-11-03
Run-now with additional artifacts.

baconmania:
/cc @ssalinas

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Optional;
 import com.hubspot.mesos.Resources;
+import com.hubspot.mesos.SingularityMesosArtifact;
 import com.wordnik.swagger.annotations.ApiModelProperty;
 
 public class SingularityRunNowRequest {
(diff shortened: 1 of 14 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
```diff
@@ -67,6 +82,11 @@ public SingularityRunNowRequest(@JsonProperty("message") Optional<String> messag
     return resources;
   }
 
+  @ApiModelProperty(required=false, value="Additional artifacts to download for this run")
+  public Optional<List<SingularityMesosArtifact>> getExtraArtifacts() {
```
ssalinas: Instead of passing Optional lists around, let’s just default it to an empty list if the constructor arg is null. Same for the other objects that use this. Will save a lot of isPresent being throw around

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -449,7 +449,10 @@ private void prepareCommand(final TaskInfo.Builder bldr, final SingularityTaskId
       commandBldr.setShell(false);
     }
 
-    for (SingularityMesosArtifact artifact : task.getDeploy().getUris().or(Collections.<SingularityMesosArtifact> emptyList())) {
+    List<SingularityMesosArtifact> combinedArtifacts = task.getDeploy().getUris().or(Collections.<SingularityMesosArtifact> emptyList());
+    combinedArtifacts.addAll(task.getPendingTask().getExtraArtifacts().or(Collections.<SingularityMesosArtifact> emptyList()));
```
ssalinas: No longer need the <...> in Java 8

ssalinas:
This got folded into https://github.com/HubSpot/Singularity/pull/1650 , closing in favor of that one

PR #1645, opened 2017-11-03
Configurable taskbuilder disk allocation.

baconmania:
This is a stopgap until we make `disk` a first-class citizen in scoring our offer scoring computations.

/cc @ssalinas

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/MesosConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/MesosConfiguration.java
@@ -22,6 +22,10 @@ public class MesosConfiguration {
   private int defaultMemory = 64;
 
   private int defaultDisk = 0;
+  // ^ This refers to SingularityDeploy-scoped resources.
+
+  // v This refers to Mesos TaskInfo-scoped resources. Changing this doesn't require a redeploy of the world.
+  private int mesosTaskBuilderDiskAllocationMb = 1;
 
   private boolean checkpoint = true;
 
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -119,8 +119,9 @@ public SingularityTask buildTask(SingularityOfferHolder offerHolder, List<Resour
     } else if (MesosUtils.getDisk(offerHolder.getCurrentResources(), Optional.absent()) >= 1.0) {
```
ssalinas: Just realized, we should be passing the requiredRole here instead of an absent

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -119,8 +119,9 @@ public SingularityTask buildTask(SingularityOfferHolder offerHolder, List<Resour
...
       // If this offer contains 1MB of disk resources, claim it to enable disk usage reporting.
       // This is just a temporary hack to enable disk usage reporting where we can, not an actual way to match task disk requirements to offers.
```
ssalinas: Maybe update this just to mention simply that disk usage won’t be reported if the disk resource isn’t > 0?

SingularityService/src/main/java/com/hubspot/singularity/config/MesosConfiguration.java
```diff
@@ -22,6 +22,10 @@
   private int defaultMemory = 64;
 
   private int defaultDisk = 0;
+  // ^ This refers to SingularityDeploy-scoped resources.
+
+  // v This refers to Mesos TaskInfo-scoped resources. Changing this doesn't require a redeploy of the world.
+  private int mesosTaskBuilderDiskAllocationMb = 1;
```
ssalinas: Trying to think of a clearer name for this, but nothing coming to mind at the moment

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
baconmania: Hmm, so if the task actually requests disk, we do pass along the requiredRole (in the `if()` just above this line). For this case, though, the task hasn't requested any disk resources, but we just claim some to enable disk reporting. In this case, we don't care if the role actually matches, do we?

ssalinas:
@baconmania is this PR still needed?

baconmania:
Nope, this was just a temporary way to instrument QA for disk reporting.

PR #1647, opened 2017-11-07
S3 folders

kwm4385:
Groups S3 files in the UI into folders based on the top level directory of their S3 key (`service`, `access`, `thread_dumps`, etc.).
kapture 2017-11-07 at 14 21 11

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskS3Logs.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskS3Logs.jsx
@@ -12,7 +12,6 @@ class TaskS3Logs extends Component {
     this.state = {
       viewingGroup: null
     };
-    this.getFileType = this.getFileType.bind(this);
   }
 
   getFileType(s3File) {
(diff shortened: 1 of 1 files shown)


SingularityUI/app/components/taskDetail/TaskS3Logs.jsx
```diff
@@ -1,67 +1,118 @@
...
-          cellData={(s3File) => (
-            <a href={s3File.downloadUrl} target="_blank" title="Download">
-              <Glyphicon glyph="download-alt" />
+class TaskS3Logs extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      viewingGroup: null
+    };
+    this.getFileType = this.getFileType.bind(this);
```
andyhuang91: This doesn't need a bind since it never uses this. Sometimes I'll try to separate out pure functions / utility methods from class methods in my JS code.

kwm4385:
Talked to @andyhuang91 irl, this should be ready to ship @ssalinas

ssalinas:
🚢

PR #1649, opened 2017-11-13
Don't error when ArtifactManager copies duplicate files

pschoenfelder:


SingularityS3Base/src/main/java/com/hubspot/singularity/s3/base/ArtifactManager.java
```diff
@@ -200,10 +201,16 @@ private void downloadUri(String uri, Path path) {
...
 
+    Path destinationPath = destination.resolve(destinationFilename);
     try {
       Files.createDirectories(destination);
-      Files.copy(source, destination.resolve(destinationFilename));
-    } catch (IOException e) {
+      Files.copy(source, destinationPath);
+    } catch (FileAlreadyExistsException e) {
+      if (!calculateMd5sum(source).equals(calculateMd5sum(destinationPath))) {
+        Throwables.propagate(e);
```
szabowexler: Note that `Throwables#propagate` is deprecated (see https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate)

ssalinas: Yeah, we have an open TODO to replace a number of guava things like propagate/Optional in Singularity. If we replace propagate usage @pschoenfelder can we do it throughout the whole class at least? Would be good to start cleaning these up as we go along

ssalinas:
🚢

PR #1650, opened 2017-11-14
Overrides for run now requests

pschoenfelder:


pschoenfelder:
Full disclosure - I have no idea if this touches all the things it needs to. I added a unit test to check that environment variable overrides specified in a run-now request are propagated through to a Mesos task, but I don't know if this covers all the workflows (I doubt it does). Do you know if/what other things need to pick up on the overrides?

pschoenfelder:
Also, I added a builder for run-now requests for the unit tests since there was a lot of `Optional.absent()` boilerplate going on, but I only added it to the test package. Would it be useful elsewhere?

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilderTest.java
```diff
@@ -138,6 +141,40 @@ public void testJobUserPassedAsEnvironmentVariable() {
   }
 
   @Test
+  public void testEnvironmentVariableOverrides() {
+    Map<String, String> overrideVariables = new HashMap<>();
+    overrideVariables.put("HS_ACCESS_GROUP", "test");
```
ssalinas: generally try to keep HubSpot-specific things out of Singularity, even for tests ;)

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilderTest.java
```diff
@@ -138,6 +141,40 @@ public void testJobUserPassedAsEnvironmentVariable() {
...
+    final SingularityRequest request = new SingularityRequestBuilder("test", RequestType.WORKER)
+        .build();
+    final SingularityDeploy deploy = new SingularityDeployBuilder("test", "1")
+        .setCommand(Optional.of("/bin/echo hi"))
+        .setRunImmediately(Optional.of(runNowRequest))
+        .build();
+    final SingularityTaskRequest taskRequest = new SingularityTaskRequest(request, deploy, pendingTask);
+    final SingularityTask task = builder.buildTask(offerHolder, null, taskRequest, taskResources, executorResources);
+
+    Map<String, String> environmentVariables = task.getMesosTask()
```
ssalinas: could probably just do the for loop here. `for (Variable var : task.getMesosTask().getCommand().getEnvironment().getVariablesList())` and just do the assert equals against getName/getKey. Saves you streaming to a map, only to then iterate.

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilderTest.java
```diff
@@ -138,6 +141,40 @@ public void testJobUserPassedAsEnvironmentVariable() {
...
+
+    SingularityRunNowRequest runNowRequest = new SingularityRunNowRequestBuilder()
+        .environmentVariables(overrideVariables)
+        .build();
+
+    final SingularityRequest request = new SingularityRequestBuilder("test", RequestType.WORKER)
+        .build();
+    final SingularityDeploy deploy = new SingularityDeployBuilder("test", "1")
+        .setCommand(Optional.of("/bin/echo hi"))
+        .setRunImmediately(Optional.of(runNowRequest))
```
ssalinas: This particular field specifies to run immediately with these settings upon successful completion of a deploy. In this case it isn't actually getting transferred to your pending task

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -163,63 +166,76 @@ private boolean hasLiteralPortMapping(Optional<SingularityContainerInfo> maybeCo
...
       }
 
       if (override.getMemoryMb() != 0) {
-        setEnv(envBldr, "DEPLOY_MEM", ((long) override.getMemoryMb()));
+        envVars.put("DEPLOY_MEM", ((long) override.getMemoryMb()));
       }
+    }
 
+    if (task.getDeploy().getRunImmediately().isPresent()) {
+      SingularityRunNowRequest runNowRequest = task.getDeploy().getRunImmediately().get();
```
ssalinas: As mentioned below, this is a field that specifies config to run immediately upon successful deploy. The `SingularityPendingTask` is where the environment override needs to be trasferred to first. So, something like:

- New RunNowRequest comes in with environmentOverrides and the `SingularityRunNowRequest` is saved as part of the `SingularityPendingRequest` and put into the pending queue (triggered from the `RequestResource`
- overrides from RunNowRequest saved as part of the SingularityPendingTask when it is created based on that pending request from the queue (this is the piece that is currently missing)
- When the task builder is ready to build the task from the pending task, it can then read the environment overrides from the pending task and apply those here

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilderTest.java
```diff
@@ -138,6 +141,40 @@ public void testJobUserPassedAsEnvironmentVariable() {
...
+        .build();
+    final SingularityTaskRequest taskRequest = new SingularityTaskRequest(request, deploy, pendingTask);
```
ssalinas: The pendingTask being used here is the one created in the `Before`. Since the environment overrides will be on the pending task (see more detail in task builder comments), we'll need to construct our own pending task to test this here

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
```diff
@@ -15,16 +16,18 @@
   private final Optional<List<String>> commandLineArgs;
   private final Optional<Boolean> skipHealthchecks;
   private final Optional<Resources> resources;
+  private final Optional<Map<String, String>> environmentVariables;
```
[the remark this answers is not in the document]
pschoenfelder: Yup - definitely more clear

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -163,63 +166,76 @@ private boolean hasLiteralPortMapping(Optional<SingularityContainerInfo> maybeCo
     return maybeContainerInfo.isPresent() && maybeContainerInfo.get().getDocker().isPresent() && !maybeContainerInfo.get().getDocker().get().getLiteralHostPorts().isEmpty();
   }
 
-  private void setEnv(Environment.Builder envBldr, String key, Object value) {
+  private void setEnv(Environment.Builder envBldr, Object key, Object value) {
```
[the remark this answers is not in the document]
pschoenfelder: Hmm...there's a line missing from this diff. There were a number of calls to this that input `key.toString()` so I move the `toString` inside this method (which isn't showing up right now, I'll see if it appears next push).

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -163,63 +166,76 @@ private boolean hasLiteralPortMapping(Optional<SingularityContainerInfo> maybeCo
...
+      SingularityRunNowRequest runNowRequest = task.getDeploy().getRunImmediately().get();
+
+      if (runNowRequest.getEnvironmentVariables().isPresent()) {
+        for (Entry entry : runNowRequest.getEnvironmentVariables().get().entrySet()) {
+          envVars.put(entry.getKey().toString(), entry.getValue());
+        }
+      }
     }
 
+    Environment.Builder envBldr = Environment.newBuilder();
```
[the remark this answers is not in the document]
pschoenfelder: My unit test was catching an error scenario where an overridden variable was duplicated in the `Environment.Builder`. I didn't find anything in the docs to indicate that there's anything preventing this or a clean way to handle dupes, so I stuck everything in a map first. Do you know if there's a better way to do this without the map?

SingularityService/src/test/java/com/hubspot/singularity/SingularityRunNowRequestBuilder.java
```diff
@@ -0,0 +1,70 @@
+package com.hubspot.singularity;
+
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.base.Optional;
+import com.hubspot.mesos.Resources;
+import com.hubspot.singularity.api.SingularityRunNowRequest;
+
+public class SingularityRunNowRequestBuilder {
```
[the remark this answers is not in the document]
pschoenfelder: Awesome! Will move it

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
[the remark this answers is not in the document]
tpetr: SingularityTask uses `env` for the env var map -- should we follow that convention?

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilderTest.java
pschoenfelder: 👍 Good catch - didn't think of that

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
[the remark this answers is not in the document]
ssalinas: Good point, SingularityDeploy does as well.

SingularityService/src/test/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilderTest.java
pschoenfelder: True, but the mesos task variables are a superset of the override variables, so the loop would require some kind of checking like this: ```    Boolean checkedMyNewEnvVar = false;
    Boolean checkedStartedByUser = false;

    for (Variable var : task.getMesosTask().getCommand().getEnvironment().getVariablesList()) {
      if (overrideVariables.containsKey(var.getName())) {
        if (var.getName().equals("MY_NEW_ENV_VAR")) {
          checkedMyNewEnvVar = true;
        } else if (var.getName().equals("STARTED_BY_USER")) {
          checkedStartedByUser = true;
        }

        assertEquals(
            "Environment variable " + var + " not overridden.",
            var.getValue(),
            overrideVariables.get(var.getName()));
      }
    }
    assertTrue(checkedMyNewEnvVar);
    assertTrue(checkedStartedByUser);```
Which do you think is cleaner (or is there an even cleaner alternative?)?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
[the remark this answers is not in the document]
pschoenfelder: Think it was more useful as I was developing. Looking at it now though, no good reason. I'll revert it.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
pschoenfelder: Next diff will have changes that I believe satisfy this, though I'm not sure my implementation is as clean as it could be. Either way, let me know and I'll work through it.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityPendingTaskBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+    this.actionId = Optional.of(actionId);
+    return this;
+  }
+
+  public SingularityPendingTask build() {
+    return new SingularityPendingTask(
+        pendingTaskId, cmdLineArgsList, user, runId, skipHealthchecks, message, resources, envOverrides, actionId
+    );
+  }
+
```
ssalinas: Good to add a toString on these. Ends up being helpful for debugging later if we ever have to log it out.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRunNowRequestBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+  public SingularityRunNowRequestBuilder setRunAt(Long runAt) {
+    this.runAt = Optional.of(runAt);
+    return this;
+  }
+
+  public SingularityRunNowRequest build() {
+    return new SingularityRunNowRequest(
+        message, skipHealthchecks, runId, commandLineArgs, resources, envOverrides, runAt);
+  }
+
```
ssalinas: same toString comment for this one

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
```diff
@@ -15,34 +17,38 @@
...
-    this(message, skipHealthchecks, runId, commandLineArgs, resources, Optional.<Long>absent());
+    this(message, skipHealthchecks, runId, commandLineArgs, resources, envOverrides, Optional.<Long>absent());
   }
 
   @JsonCreator
-  public SingularityRunNowRequest(@JsonProperty("message") Optional<String> message,
-                                  @JsonProperty("skipHealthchecks") Optional<Boolean> skipHealthchecks,
-                                  @JsonProperty("runId") Optional<String> runId,
-                                  @JsonProperty("commandLineArgs") Optional<List<String>> commandLineArgs,
+  public SingularityRunNowRequest(@JsonProperty("setMessage") Optional<String> message,
```
ssalinas: the `@JsonProperty` on these needs to stay as it was before. This is one of the pieces jackson uses to determine the field name in the json. e.g. `@JsonProperty("message")` becomes `{"message": {}}`

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
```diff
@@ -62,11 +68,16 @@ public SingularityRunNowRequest(@JsonProperty("message") Optional<String> messag
     return skipHealthchecks;
   }
 
-  @ApiModelProperty(required=false, value="Override the resources from the active deploy for this run")
+  @ApiModelProperty(required=false, value="Override the setResources from the active deploy for this run")
```
ssalinas: similar comment here, don't want to rename these to `set...`

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRunNowRequestBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+  private Optional<Long> runAt;
+
+  public SingularityRunNowRequestBuilder()
+  {
+    this.message = Optional.absent();
+    this.runId = Optional.absent();
+    this.commandLineArgs = Optional.absent();
+    this.skipHealthchecks = Optional.absent();
+    this.resources = Optional.absent();
+    this.envOverrides = null;
```
ssalinas: empty map instead of null here?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -787,9 +787,19 @@ private int numInstancesExpected(SingularityRequest request, SingularityPendingR
...
+              new SingularityPendingTaskId(
+                  request.getId(), deployId, nextRunAt.get(), nextInstanceNumber,
+                  pendingRequest.getPendingType(), pendingRequest.getTimestamp()),
+              pendingRequest.getCmdLineArgsList(),
+              pendingRequest.getUser(),
+              pendingRequest.getRunId(),
+              pendingRequest.getSkipHealthchecks(),
+              pendingRequest.getMessage(),
+              pendingRequest.getResources(),
+              null,
```
ssalinas: this should be the last missing piece of the puzzle to connect it all the way through. The `SingularityPendingRequest` is the object that will carry the envOverrides from the RunNowRequest to the pending task.

In `RequestResource.scheduleImmediately` we create a `SingularityPendingRequest` from the provided `SingularityRunNowRequest`. That pending request will need to take in the environment overrides, so that we can later use them on this line to put them on the pending task.

The `RequestResource` calls `checkRunNowRequest` on the `SingularityValidator`, which is what actually constructs the object actually running some checks. That is where the envOverrides will need to be put on the `SingularityPendingRequest`

SingularityBase/src/main/java/com/hubspot/singularity/SingularityPendingTaskBuilder.java
pschoenfelder: Good to know :)

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
pschoenfelder: Hmmm...think that was my improper use of Intellij's refactoring tools. Going to need to be more careful of that in the future.

ssalinas:
From what I can see this looks good. Thanks for those builder simplifications as well 🎉 

I think we should be good to give this one a go in `hs_staging` if it was working for you locally

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -787,9 +787,19 @@ private int numInstancesExpected(SingularityRequest request, SingularityPendingR
...
 
-      newTasks
-        .add(new SingularityPendingTask(new SingularityPendingTaskId(request.getId(), deployId, nextRunAt.get(), nextInstanceNumber, pendingRequest.getPendingType(), pendingRequest.getTimestamp()),
-          pendingRequest.getCmdLineArgsList(), pendingRequest.getUser(), pendingRequest.getRunId(), pendingRequest.getSkipHealthchecks(), pendingRequest.getMessage(), pendingRequest.getResources(), pendingRequest.getActionId()));
+      newTasks.add(
+          new SingularityPendingTask(
+              new SingularityPendingTaskId(
+                  request.getId(), deployId, nextRunAt.get(), nextInstanceNumber,
+                  pendingRequest.getPendingType(), pendingRequest.getTimestamp()),
+              pendingRequest.getCmdLineArgsList(),
```
ssalinas: indent ended up a bit funny here

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -274,39 +274,46 @@ private void finishDeploy(SingularityRequestWithState requestWithState, Optional
...
-          requestManager.addToPendingQueue(new SingularityPendingRequest(requestId, deployId, timestamp, user, pendingType, commandLineArgs, runId, skipHealthChecks, message, Optional.absent(), resources, runNowRequest.getRunAt()));
+          pendingType = PendingType.ONEOFF;
         } else {
           // Don't run one-off / on-demand task when already at instance count cap
-          PendingType pendingType = canceledOr(deployResult.getDeployState(), PendingType.NEW_DEPLOY);
-          requestManager.addToPendingQueue(new SingularityPendingRequest(requestId, deployId, timestamp, user, pendingType, commandLineArgs, runId, skipHealthChecks, message, Optional.absent(), resources, runNowRequest.getRunAt()));
+          pendingType = PendingType.NEW_DEPLOY;
         }
       }
+      if (pendingType != null) {
```
baconmania: Should we be handling the case where pendingType == null here?

pschoenfelder: I don't think so. This is just a small refactor of the previous code, which would do nothing in the case where the new code has pendingType == null.

ssalinas:
🚢

ssalinas:
🚢

pschoenfelder:
🚢

PR #1651, opened 2017-11-15
Option to provide credentials supplier in client

ssalinas:
With the new webhook auth, it is more likely that a long running system would have to re-authenticate, or provide a means of grabbing an auth token to send. This adds changes to the `SingulairtyClient` to optionally provide a `Supplier<SingularityClientCredentials>`. By default, if the supplier is present, the function will be called on the first request made by the client in order to grab credentials. And, if an unauthorized response is later received the client will try to re-authenticate as part of its retry strategy. This should be useful for longer running systems where a session may expire after some time


ssalinas:
Not needed after all for now. Closing but I'll keep the branch

PR #1654, opened 2017-11-17
Ability to specify quick links for requests

ssalinas:
This adds a way that we can set custom links to other places in the infrastructure from the request detail page. The links can be templated using formats like: `http://example.com?person={{request.owners.0}}`. The links can be specified per request type as well. I also put a top level key of "request" in the config in case we later decide to add some for tasks or something else more specific.

Links, if present, will appear in a small dropdown button in the request action buttons section:

screen shot 2017-11-17 at 7 44 22 am


ssalinas:
cc @kwm4385

SingularityUI/app/utils.es6
```diff
@@ -486,6 +487,32 @@ const Utils = {
...
+          if (value === undefined){
+            throw tokenPath[i] + "' not found in " + tag;
+          }
+          if (i === tokenPath.length - 1){
+            return value;
+          }
+        }
+      });
+    } catch (err) {
+      return undefined;
```
kwm4385: Is there a reason you're returning `undefined` rather than `null`?

ssalinas: no particular reason. Is there a preference/reasoning for one over the other?

kwm4385:
🚢

baconmania:
🚢

PR #1655, opened 2017-11-17
Easier setup for local dev with SingularityService

ssalinas:
This sets the mesos master ip and hostname in our compose-dev setup to the docker vm ip. That way we can connect a local running intellij/eclipse/etc SingularityService to the docker mesos cluster for faster development

cc @pschoenfelder


pschoenfelder:
🚢

PR #1657, opened 2017-11-20
Score with disk resources.

baconmania:
This PR rolls disk usage into Singularity's offer scoring calcs.

It also changes the default relative offer weights to `0.25 cpu; 0.65 mem; 0.10 disk` based on the fact that typical apps which Singularity is intended to host are stateless, primarily I/O bound, then CPU bound, with minimal, bounded disk usage.

Still need to update tests to exercise disk resources in tests of `SingularityMesosOfferScheduler#score()`.

/cc @ssalinas


SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -202,13 +202,17 @@
 
   private int maxTasksPerOfferPerRequest = 0;
 
-  private double longRunningUsedCpuWeightForOffer = 0.30;
+  private double longRunningUsedCpuWeightForOffer = 0.25;
 
-  private double longRunningUsedMemWeightForOffer = 0.70;
+  private double longRunningUsedMemWeightForOffer = 0.65;
```
ssalinas: This actually changes the relative weights of cpu vs memory from memory being 2.333 times the weight of cpu to 2.666 times the weight of cpu. Would we instead want .27 and .63 to keep the relative weights of those two the same?

ssalinas: Followed up in-person, will keep as is for now. mem > cpu > disk is preserved here which is the most important piece

ssalinas:
FYI that cluster coordinator change should already be in master. Maybe merge master into this before getting this in staging to avoid additional merge conflicts later on?

Looks good though 👍

tpetr:
> typical apps which Singularity is intended to host are stateless, primarily I/O bound, then CPU bound, with minimal, bounded disk usage.

This is likely outside the scope of the PR, but it could be interesting if users were able to set the weights on a per-Request basis.

baconmania:
Aye, that's a cool thought.

PR #1658, opened 2017-11-27
Corrected path for executor download fallback

ssalinas:
Previously this was using the task app directory when copying (not unzipping) which does not match what the S3Downloader service does, and does not match what untar does. This updates it to match

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorArtifactFetcher.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorArtifactFetcher.java
@@ -194,7 +194,7 @@ public class SingularityExecutorArtifactFetcher {
       if (Objects.toString(fetched.getFileName()).endsWith(".tar.gz")) {
         artifactManager.untar(fetched, task.getArtifactPath(remoteArtifact, task.getTaskDefinition().getTaskDirectoryPath()));
       } else {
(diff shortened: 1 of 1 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorArtifactFetcher.java
```diff
@@ -194,7 +194,7 @@ private void downloadRemoteArtifact(RemoteArtifact remoteArtifact, ArtifactManag
       if (Objects.toString(fetched.getFileName()).endsWith(".tar.gz")) {
         artifactManager.untar(fetched, task.getArtifactPath(remoteArtifact, task.getTaskDefinition().getTaskDirectoryPath()));
       } else {
-        artifactManager.copy(fetched, task.getArtifactPath(remoteArtifact, task.getTaskDefinition().getTaskAppDirectoryPath()), remoteArtifact.getFilename());
+        artifactManager.copy(fetched, task.getArtifactPath(remoteArtifact, task.getTaskDefinition().getTaskDirectoryPath()), remoteArtifact.getFilename());
```
tpetr: is this correct in every circumstance? I think we need to combine getTaskDirectoryPath() and targetFolderRelativeToTask() (if present)

tpetr: nevermind, I forgot that's already happening in `getArtifactPath()`

jonathanwgoodwin:
:Ship:

jonathanwgoodwin:
🚢

PR #1661, opened 2017-11-30
Enable auth headers in old tailer

ssalinas:
The old tailer has a call that was missed in the original auth updates


baconmania:
🚢

PR #1663, opened 2017-12-01
Reimplement getPortByIndex in SingularityTask

ssalinas:
This was removed in the removal of the mesos dep from SingularityBase. Adding it back in for convenience.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/mesos/protos/MesosRangesObject.java
+++ b/SingularityBase/src/main/java/com/hubspot/mesos/protos/MesosRangesObject.java
@@ -1,0 +1,57 @@
+package com.hubspot.mesos.protos;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Optional;
+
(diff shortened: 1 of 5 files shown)


SingularityBase/src/main/java/com/hubspot/mesos/protos/MesosRangesObject.java
```diff
@@ -0,0 +1,57 @@
...
+    this.range = range;
+  }
+
+  public List<MesosRangeObject> getRange() {
+    return range.or(Collections.emptyList());
+  }
+
+  @JsonIgnore // to mimic mesos
+  public List<MesosRangeObject> getRangesList() {
+    return getRange();
```
darcatron: not sure I understand the reason for this method. doesn't seem to do anything different than `getRange()`

baconmania:
🚢

SingularityBase/src/main/java/com/hubspot/mesos/protos/MesosRangesObject.java
baconmania: It's to mimic the convention used by the Mesos API, with stuff like [`getPortsList()`](http://mesos.apache.org/api/latest/java/org/apache/mesos/Protos.PortsOrBuilder.html#getPortsList--).

ssalinas: correct, added with the aim of not breaking binary compatibility for those who were using the old mesos protos objects

PR #1668, opened 2017-12-01
Support user overrides in Run-Now requests.

baconmania:


baconmania:
Just wanted to mention that there's a subtlety in `SingularityMesosTaskBuilder` that might not be entirely obvious. When using our custom executor, the `SingularityDeploy#user` field controls which user the Mesos containerizer switches to before running the `singularity-executor` script, while the `SingularityDeploy#executorData#user` field controls which user our executor switches to before actually running the Task's configured command. Is this something we want to make clearer?

/cc @ssalinas @tpetr

baconmania:
Tested & working for both default and custom executors.

PR #1671, opened 2017-12-04
Remove SingularityServiceBase module, no longer needed

### LINKED ISSUE #1670: UI broken in 0.18.1
After upgrading to 0.18.1 the singularity UI is no longer working, returning 404's for static web assets. It seems to be just urls like `/singularity/static/*`; the api urls seem to be functional.

I had previously been running on a fork version: https://github.com/intellifylearning/Singularity/compare/master...HubSpot:master

ssalinas:
`SingularityServiceBase` was originally split out when we were toying with the coordinator and had the possibility of two modules which could serve up the UI. Due to the fact that the way we package and access the UI internally slightly differs from our release process, we didn't catch the fact that moving this to another module messes with the static resources for the SingularityService shaded jar.

This PR puts all of the SingularityServiceBase bits back into SingularityService now that the separate module is no longer needed

/fixes #1670


Docs/getting-started/install.md
```diff
@@ -63,7 +63,7 @@ database:
   url: jdbc:mysql://[database host]:[database port]/[database name]
 
 mesos:
-  master: http://[mesos master hostname]/api/v1/scheduler
+  master: [comma separated list of mesos master host:post]
```
baconmania: `host:port`?

baconmania:
Good to go other than the small typo there.

baconmania:
:ship:

PR #1674, opened 2017-12-04
Different order for ui build profile

ssalinas:
Finally found the root cause of https://github.com/HubSpot/Singularity/issues/1670 , also masked by our build/deploy process for the UI being slightly different internally. We were running add-resources before we actually built the ui :facepalm:

/cc @baconmania


baconmania:
🚢

PR #1681, opened 2017-12-06
Provide option to prevent emails for scale events

darcatron:
This updates the scaleRequest object to include another optional boolean `skipEmailNotification`. By default, we'll send emails for scale events, but users can toggle their request to prevent email notifications.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityScaleRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityScaleRequest.java
@@ -11,16 +11,23 @@ public class SingularityScaleRequest extends SingularityExpiringRequestParent {
   private final Optional<Boolean> skipHealthchecks;
   private final Optional<Boolean> bounce;
   private final Optional<Boolean> incremental;
+  private final Optional<Boolean> skipEmailNotification;
 
   @JsonCreator
-  public SingularityScaleRequest(@JsonProperty("instances") Optional<Integer> instances, @JsonProperty("durationMillis") Optional<Long> durationMillis,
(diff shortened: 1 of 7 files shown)


ssalinas:
🚢

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityScaleRequest.java
```diff
@@ -11,16 +11,23 @@
...
-      @JsonProperty("skipHealthchecks") Optional<Boolean> skipHealthchecks, @JsonProperty("actionId") Optional<String> actionId, @JsonProperty("message") Optional<String> message,
-      @JsonProperty("bounce") Optional<Boolean> bounce, @JsonProperty("incremental") Optional<Boolean> incremental) {
+  public SingularityScaleRequest(@JsonProperty("instances") Optional<Integer> instances,
+                                 @JsonProperty("durationMillis") Optional<Long> durationMillis,
+                                 @JsonProperty("skipHealthchecks") Optional<Boolean> skipHealthchecks,
+                                 @JsonProperty("actionId") Optional<String> actionId,
+                                 @JsonProperty("message") Optional<String> message,
+                                 @JsonProperty("bounce") Optional<Boolean> bounce,
+                                 @JsonProperty("incremental") Optional<Boolean> incremental,
+                                 @JsonProperty("skipEmailNotification") Optional<Boolean> skipEmailNotification) {
```
ssalinas: documenting in-person conversation here as well, we should keep the old constructor around as well so we don't break anyone's builds 👍

ssalinas:
🚢

PR #1682, opened 2017-12-07
Ability to update authorized groups separately from full request

ssalinas:
This PR does a few things:
- Allows a readWrite user to update readWriteGroups or group as long as they are authorized for both the new _and_ old groups
- Adds an endpoint to only update group, readWriteGroups, readOnlyGroups for a request, without needing to post the full request body
- Adds an endpoint to verify a user is authorized to perform a group update on a request


darcatron:
🚢 ☘️

PR #1683, opened 2017-12-08
Support usage of DefaultServerFactory to find port

ssalinas:
Needed for Singularity to handle its own ssl.

For the moment requires that an http port is present either in simple or connector formats in dropwizard yaml. A future improvement will be to also parse an https port and default to that when available


baconmania:
🚢

PR #1684, opened 2017-12-08
Point Singularity at the "external" interface where mesos-master is listening.

baconmania:


ssalinas:
Let's add this to docker-compose.yml as well as should be gtg 👍

ssalinas:
🚢

PR #1685, opened 2017-12-13
WIP - Allow a request in FINISHED state to be redeployed

ssalinas:
Should probably add some messaging in the ui about what FINISHED actually means (that a provided schedule has no more iterations).

This _should_ cover the case of allowing a finished request to be redeployed, since there is currently no other way to update it other than to pause + deploy to unpause with a new schedule. Previously it was hitting the case in the deploy checker for `isActive` and failing every deploy at that point.

This sets the state of the request back to active when the new deploy is created, which will allow it to pass the deploy checked.

Still need to add tests and determine if a deploy failure will allow it to move back to finished state correctly


ssalinas:
Changed this up and added a test for it. Instead of setting to active before the deploy, I am setting it to active after a successful deploy, and allowing FINISHED states to pass through the deploy checker.

pschoenfelder:
🚢

PR #1688, opened 2017-12-20
Also check pending requests on the track task endpoint

ssalinas:
If a RunNowRequest was enqueued, but still in the pending request queue waiting to be processed by the scheduler, the track endpoint would return an absent. We should also be able to mark the task as pending when it is in the pending request queue. This adds an additional check of the pending requests


darcatron:
:ship:

PR #1689, opened 2017-12-20
Pull in the shaded-protobuf version of mesos

jhaber:
This bumps to basepom 18.1 and switches the mesos dep to the shaded version (excluding the non-shaded version where necessary)

In terms of dep changes this brings:
dropwizard 1.0.6 -> 1.0.9
guava-retrying 1.0.7 -> 2.0.0
guice 4.0 -> 4.1
jackson-databind 2.7.9 -> 2.7.9.1
jade4j 0.4.0 -> 0.4.2
javax.mail 1.5.2 -> 1.5.6
jersey 2.22.2 -> 2.23.2
jetty 9.3.9.v20160517 -> 9.3.20.v20170531
metrics 3.1.2 -> 3.1.4
mysql 5.1.39 -> 5.1.43
netty3 3.9.2.Final -> 3.9.4.Final
netty4 4.1.4 -> 4.1.8

These upgrades seem pretty benign but if you're concerned about any of them we can force the previous version

### DIFF
--- a/SingularityExecutor/pom.xml
+++ b/SingularityExecutor/pom.xml
@@ -58,6 +58,7 @@
     <dependency>
       <groupId>org.apache.mesos</groupId>
       <artifactId>mesos</artifactId>
+      <classifier>shaded-protobuf</classifier>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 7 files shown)


jhaber:
Closing as jackson-datatype-protobuf wouldn't be able to handle these shaded protobufs. If we ever need/want to do this in the future we could get around this by adding explicit Jackson serializers/deserializers for the Mesos protobufs we care about

PR #1692, opened 2018-01-08
Wire up overrides for the S3 uploader path.

baconmania:


SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
```diff
@@ -40,6 +42,7 @@ public SingularityRunNowRequest(@JsonProperty("message") Optional<String> messag
                                   @JsonProperty("runId") Optional<String> runId,
                                   @JsonProperty("commandLineArgs") Optional<List<String>> commandLineArgs,
                                   @JsonProperty("resources") Optional<Resources> resources,
+                                  @JsonProperty("s3UploaderAdditionalFiles") List<SingularityS3UploaderFile> s3UploaderAdditionalFiles,
```
ssalinas: remember this can be used from the client, so adding args will break the build for folks

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosTaskBuilder.java
```diff
@@ -417,7 +418,12 @@ private void prepareCustomExecutor(final TaskInfo.Builder bldr, final Singularit
         executorDataBldr.setExtraCmdLineArgs(extraCmdLineArgsBuilder.build());
       }
 
-      List<SingularityS3UploaderFile> uploaderAdditionalFiles = configuration.getS3ConfigurationOptional().isPresent() ? configuration.getS3ConfigurationOptional().get().getS3UploaderAdditionalFiles() : Collections.<SingularityS3UploaderFile>emptyList();
+      List<SingularityS3UploaderFile> uploaderAdditionalFiles = new ArrayList<>();
+      if (configuration.getS3ConfigurationOptional().isPresent()) {
+        uploaderAdditionalFiles.addAll(configuration.getS3ConfigurationOptional().get().getS3UploaderAdditionalFiles());
+      }
+      uploaderAdditionalFiles.addAll(task.getPendingTask().getS3UploaderAdditionalFiles());
```
ssalinas: is it possible for two of these to overlap? The original use for this was to override where a specific file was being sent. Maybe we need to dedupe on the directory/glob in the additional file settings? If the uploader tries to upload it twice, one instance of that will fail when the first one done tries to delete it

baconmania: Hmm, that's a tricky one. Since we don't have a priori information about the actual set of files in the sandbox, it's nontrivial to determine if the glob patterns overlap. We could just compare the pattern strings directly, but that would of course not cover all potential overlaps.

baconmania: After digging on this, it looks like a [ConcurrentMap](https://github.com/HubSpot/Singularity/blob/master/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java#L347-L350) is used to [ensure](https://github.com/HubSpot/Singularity/blob/master/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3Uploader.java#L175-L179) that two uploaders don't simultaneously try to handle the same file. 

This means we won't have uploader errors due to conflicts, although it does leave a race condition in place. The way to solve this in the future would be to ensure that upload overrides for a file use a matching `directory` and `glob` strings as the corresponding global config, and then just compare those fields directly instead of trying to compute glob overlaps.

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityRunNowRequest.java
baconmania: 👍  good catch. Restored the constructor from the most recent release.

ssalinas:
🚢

PR #1693, opened 2018-01-09
Properly send task destroy message to executor

ssalinas:
I was mistakenly calling `message` here instead of `mesosSchedulerClient.frameworkMessage`, so the scheduler was just talking to itself :facepalm:

/cc @baconmania


baconmania:
🚢

PR #1696, opened 2018-01-18
Fix nav bar for mobile view

benheng:


benheng:
### Hidden navigation with hamburger button:
> screen shot 2018-01-18 at 15 41 49

### Hamburger button clicked opens a vertical nav list:
> screen shot 2018-01-18 at 15 42 11

andyhuang91:
LGTM. Since we are using react-bootstrap, you could also simplify the rest of the navigation file by using those components.

PR #1697, opened 2018-01-19
Support uploads to GCS

ssalinas:
WIP, still need to figure out passing credentials around in the metadata

cc @baconmania


SingularityS3Uploader/pom.xml
```diff
@@ -14,6 +14,16 @@
     <basepom.shaded.main-class>com.hubspot.singularity.s3uploader.SingularityS3UploaderRunner</basepom.shaded.main-class>
   </properties>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.protobuf</groupId>
+        <artifactId>protobuf-java</artifactId>
+        <version>3.5.1</version>
```
ssalinas: This is gross, but still figuring out a clean way to do these dependencies. The latest version of `google-cloud-storage` also pulls in an incompatible version of `google-http-client` as compared to the latest version of `google-auth-library-oauth2-http` which is an additional pain

ssalinas: @baconmania or @stevegutz any thoughts on the best way to handle these weird dep issues here? Works to override protobuf here because the S3Uploader doesn't actually have to pull in org.apache.mesos, but not sure that it's the cleanest thing to do

ssalinas:
tested and working in hs_staging

SingularityS3Uploader/pom.xml
stevegutz: What's pulling in the protobuf dep?

ssalinas: org.apache.mesos pulls in 2.6, google-cloud-storage pulls in 3.5.1

ssalinas: I guess I'm more wondering if it's too much of a code smell to override the version for just this module when the rest of Singularity is using 2.6

baconmania:
🚢

PR #1698, opened 2018-01-23
Don't show s3 logs error message as a pop up

ssalinas:
In the specific case of the S3 logs, folks have found the pop up error message to be more annoying than helpful. This updates the code to catch the 500 and avoid that pop up, replacing it with a message on the S3 Logs section itself

cc @kwm4385


kwm4385:
🚢

PR #1700, opened 2018-01-29
Fixed task webhook docs

pschoenfelder:


Docs/reference/apidocs/models.md
```diff
@@ -1194,6 +1195,14 @@ Models:
 | shellRequestId | [SingularityTaskShellCommandRequestId](models.md#model-SingularityTaskShellCommandRequestId) | optional |  |
 
 
+## <a name="model-SingularityTaskWebhook"></a> SingularityTaskWebhook
```
ssalinas: A quick note that everything in the apidocs folder here is auto generated by swagger for us. What you'll want to do is add swagger annotations to the object, similar to what you see [here](https://github.com/HubSpot/Singularity/blob/master/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java#L193). That should get swagger to include it in the models docs for us

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskWebhook.java
```diff
@@ -14,10 +15,12 @@ public SingularityTaskWebhook(@JsonProperty("task") SingularityTask task, @JsonP
     this.taskUpdate = taskUpdate;
   }
 
+  @ApiModelProperty(required=false, value="The task this webhook refers to.")
```
ssalinas: Due to the fact that this is not a model returned by a specific endpoint, I don't think swagger will pick it up without an `@ApiModel` annotation on the class as a whole. You can test this out running something like `mvn -Pbuild-swagger-documentation -DskipTests=true -B -q -fae install` to generate the documentation and check if the model appears

pschoenfelder: Hmm good catch. Even with the `@ApiModel` annotation it doesn't appear though. Google's not telling me much about this. I've compared against the other classes annotated with this in Singularity too and it seems okay. Any ideas?

pschoenfelder: Update - added the link to github instead of the docs link since swagger still isn't building this even when marked with the annotation.

ssalinas:
🚢

PR #1701, opened 2018-01-30
Permalinks for bounce and scale modals

kwm4385:
Appending `?bounce=true` or `?scale=true` on any request page will have the respective modal automatically open.


ssalinas:
🚢

PR #1703, opened 2018-02-02
Better webhook auth timeouts and exception messages

ssalinas:
Previously, when using multiple auth methods, if all failed singularity would only return the message from the final one attempted. Also, the AsyncHttpClient used for webhook auth calls had all of the default timeouts and retries set, which are too high to auth calls.

This PR updates the multiple auth method caller to gather exception messages from each attempted auth method and return a combined message. It also updates the webhook auth http client to user much shorter timeouts (now defaults to 2s connect, 2s request, 2 retries) which are configurable in the auth configuration

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/auth/authenticator/SingularityWebhookAuthenticator.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/auth/authenticator/SingularityWebhookAuthenticator.java
@@ -42,8 +42,8 @@ public class SingularityWebhookAuthenticator implements SingularityAuthenticator
         .refreshAfterWrite(webhookAuthConfiguration.getCacheValidationMs(), TimeUnit.MILLISECONDS)
         .build(new CacheLoader<String, SingularityUserPermissionsResponse>() {
           @Override
-          public SingularityUserPermissionsResponse load(String authHeaderVaule) throws Exception {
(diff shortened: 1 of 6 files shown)


SingularityService/src/main/java/com/hubspot/singularity/auth/authenticator/SingularityWebhookAuthenticator.java
```diff
@@ -26,18 +29,34 @@
...
     this.asyncHttpClient = asyncHttpClient;
     this.webhookAuthConfiguration = configuration.getWebhookAuthConfiguration();
     this.objectMapper = objectMapper;
     this.permissionsCache = CacheBuilder.<String, SingularityUserPermissionsResponse>newBuilder()
-        .expireAfterWrite(webhookAuthConfiguration.getCacheValidationMs(), TimeUnit.MILLISECONDS)
-        .build();
+        .refreshAfterWrite(webhookAuthConfiguration.getCacheValidationMs(), TimeUnit.MILLISECONDS)
+        .build(new CacheLoader<String, SingularityUserPermissionsResponse>() {
+          @Override
+          public SingularityUserPermissionsResponse load(String authHeaderVaule) throws Exception {
```
aaylward: s/authHeaderVaule/authHeaderValue

pschoenfelder:
🚢

PR #1704, opened 2018-02-02
Add support for configurable navbar title links.

baconmania:
Looks like 
image

This setup defaults to having the nav header title link back to the Dashboard (as is the case today) if no links are configured.

Might have to tweak the CSS class setup slightly, since the `Admin` dropdown has squared corners, but this one's got rounded corners.

### DIFF
--- a/SingularityUI/app/components/common/Navigation.jsx
+++ b/SingularityUI/app/components/common/Navigation.jsx
@@ -35,6 +35,26 @@ function isActive(navbarPath, fragment) {
 // put into page wrapper, render children
 const Navigation = (props) => {
   const fragment = props.location.pathname.split('/')[1];
+  let renderedNavTitle;
+
+  if (Object.keys(config.navTitleLinks).length > 0) {
+    const renderedNavTitleLinks = Object.keys(config.navTitleLinks).map(linkTitle => {
+      return <li><a href={config.navTitleLinks[linkTitle]}>{linkTitle}</a></li>;
+    });
+
+    renderedNavTitle =
+      <div className="dropdown nav">
+        <a href="#" className="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
+          {config.title} <span className="caret"></span>
+        </a>
+        <ul className="dropdown-menu">
+          {renderedNavTitleLinks}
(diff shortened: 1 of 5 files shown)


tpetr:
yaaaas

ssalinas:
Will let @kwm4385 and the more-UI-inclined take a look at the react bits, but I agree on making this and the admin dropdown match

SingularityUI/app/components/common/Navigation.jsx
```diff
@@ -35,6 +35,26 @@ function isActive(navbarPath, fragment) {
 // put into page wrapper, render children
 const Navigation = (props) => {
   const fragment = props.location.pathname.split('/')[1];
+  const navTitle = config.navTitleLinks
+    ? (
+      <ul className="nav navbar-nav">
```
kwm4385: Wrapping this in the same `ul` structure will give the consistent menu look you seek.

SingularityUI/app/components/common/Navigation.jsx
```diff
@@ -35,6 +35,26 @@ function isActive(navbarPath, fragment) {
 // put into page wrapper, render children
 const Navigation = (props) => {
   const fragment = props.location.pathname.split('/')[1];
+  const navTitle = config.navTitleLinks
```
kwm4385: Needed a null check for the config field itself somewhere, otherwise it would break if it wasn't set.

SingularityUI/app/components/common/Navigation.jsx
```diff
@@ -35,6 +35,26 @@ function isActive(navbarPath, fragment) {
...
+    ? (
+      <ul className="nav navbar-nav">
+        <li className="dropdown nav">
+          <a href="#" className="dropdown-toggle navbar-brand" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
+            {config.title} <span className="caret" />
+          </a>
+          <ul className="dropdown-menu">
+            {Object.keys(config.navTitleLinks).map((linkTitle, index) =>
```
kwm4385: These links can be inlined

kwm4385:
image

ssalinas:
🚢

PR #1705, opened 2018-02-06
Capital search

benheng:
Request and task filtering with all caps
For example, filtering with "HWB" would return something like this:
```
HelloWorldBen
LookHeyWheresBillyGoneTo
```


SingularityUI/app/selectors/requests.es6
```diff
@@ -1,165 +0,0 @@
...
-const getSearchFilter = (state) => state.ui.requestsPage;
-const getCurrentGroup = (state) => state.ui.dashboard.currentGroup;
-
-function findRequestIds(requests) {
-  return _.map(requests, (request) => {
-    return _.extend({}, request, {id: request.request ? request.request.id : request.requestId});
-  });
-}
-
-export const getStarredRequests = createSelector(
```
ssalinas: were the rest of these methods unused I take it?

benheng: Yeah none of these were used anywhere and I was wondering if it was duplicated by `selectors/requests/filterSelector` but now I'm not sure. It looks like it has selectors for some incomplete features, like `getStarredRequests`. I preferring removing unused code, but I can see how this could be helpful for completing some feature requests.

kwm4385:
This doesn't seem to be working on staging, did you deploy after merging there?

SingularityUI/app/selectors/requests/filterSelector.es6
```diff
@@ -74,9 +74,16 @@ export default createSelector([getRequests, getFilter, getUtilizations], (reques
       });
       filteredRequests = _.union(res1, res2).reverse();
     } else {
-      _.each(filteredRequests, (requestParent) => {requestParent.id = id.extract(requestParent);});
+      _.each(filteredRequests, (requestParent) => (requestParent.id = id.extract(requestParent)));
+      // Allow searching by the first letter of each word by applying same
+      // search heuristics to just the upper case characters of each option
+      const options = {
+        extract: Utils.isAllUpperCase(filter.searchFilter)
+          ? (requestParent) => Utils.getUpperCaseCharacters(requestParent.id)
```
andyhuang91: Doesn't the exact method have to be called in the all uppercase case too?

SingularityUI/app/selectors/requests/filterSelector.es6
```diff
@@ -74,9 +74,16 @@ export default createSelector([getRequests, getFilter, getUtilizations], (reques
       });
       filteredRequests = _.union(res1, res2).reverse();
     } else {
-      _.each(filteredRequests, (requestParent) => {requestParent.id = id.extract(requestParent);});
+      _.each(filteredRequests, (requestParent) => (requestParent.id = id.extract(requestParent)));
```
andyhuang91: Returning a value in `_.each` shouldn't do anything so I'm not sure why this change was necessary. The existing code is pretty unclear about what this loop actually does. There appears to be some sort of side effect with the `extract` function. Would it be easy to clean up the existing code?

benheng:
@kwm4385 I pushed it to hs_staging, but forgot to deploy it. I'll do it again once I've addressed the other comments.

SingularityUI/app/selectors/requests/filterSelector.es6
benheng: This _is_ the extract method. In the uppercase case, it's returning only the capitals of each request.

SingularityUI/app/selectors/requests/filterSelector.es6
benheng: Hmm I did seem to make it less clear since now it's returning something and we don't use it. it was unclear to begin so I'll see if I can clean it up.

benheng: `selectors/tasks` has similar code and it looks like it makes more sense in that file and less sense here.

kwm4385:
🚢

SingularityUI/app/selectors/requests/filterSelector.es6
```diff
@@ -62,22 +62,29 @@ export default createSelector([getRequests, getFilter, getUtilizations], (reques
...
+      const users = _.filter(filteredRequests, (requestParent) => (
+        micromatch.isMatch(getUser(requestParent), `${filter.searchFilter}*`)
+      ));
+      const ids = _.filter(filteredRequests, (requestParent) => (
+        micromatch.isMatch(getId(requestParent), `${filter.searchFilter}*`)
+      ));
+      filteredRequests = _.union(users, ids);
+    } else {
+      // Allow searching by the first letter of each word by applying same
+      // search heuristics to just the upper case characters of each option
```
andyhuang91: Comment is in wrong place

SingularityUI/app/selectors/requests/filterSelector.es6
```diff
@@ -62,22 +62,29 @@ export default createSelector([getRequests, getFilter, getUtilizations], (reques
...
   if (filter.searchFilter) {
-    const id = {extract: (requestParent) => requestParent.id || ''};
-    const user = {extract: (requestParent) => `${requestParent.hasActiveDeploy ? requestParent.requestDeployState.activeDeploy.user : ''}`};
+    const getId = (requestParent) => requestParent.id || '';
+    const getUser = (requestParent) => requestParent.hasActiveDeploy && requestParent.requestDeployState.activeDeploy.user || '';
 
     if (Utils.isGlobFilter(filter.searchFilter)) {
-      const res1 = _.filter(filteredRequests, (requestParent) => {
-        return micromatch.any(user.extract(requestParent), `${filter.searchFilter}*`);
+      const users = _.filter(filteredRequests, (requestParent) => (
```
andyhuang91: maybe `userMatches` and `idMatches` for clarity instead of just `users`

SingularityUI/app/selectors/tasks.es6
```diff
@@ -49,29 +49,36 @@ export const getFilteredTasks = createSelector(
...
+        const ids = _.filter(tasks, (task) => (
+          micromatch.isMatch(getId(task), `${filter.filterText}*`)
+        ));
+        const racks = _.filter(tasks, (task) => (
+          micromatch.isMatch(getRack(task), `${filter.filterText}*`)
+        ));
+        tasks = _.union(hosts, ids, racks);
+      } else {
+        // Allow searching by the first letter of each word by applying same
+        // search heuristics to just the upper case characters of each option
```
andyhuang91: comment here is also in the wrong place

benheng:
Okay I've actually correctly deployed it to `hs_staging` this time.

PR #1706, opened 2018-02-07
Smarter table pagination

benheng:
Make table pagination more obvious by fetching ahead to see if there is more data. If not, hide the "next" button.

kapture 2018-02-07 at 15 30 45

### DIFF
--- a/SingularityUI/app/components/common/table/UITable.jsx
+++ b/SingularityUI/app/components/common/table/UITable.jsx
@@ -10,44 +10,40 @@ class UITable extends Component {
 
   constructor(props) {
     super(props);
-
-    let { data } = props;
-    const { defaultSortBy, defaultSortDirection, rowChunkSize } = props;
-    if (defaultSortBy) {
-      data = this.doSort(data, defaultSortBy, defaultSortDirection);
-    }
+    const {
+      data,
+      defaultSortBy,
+      defaultSortDirection,
+      rowChunkSize
+    } = props;
 
     this.state = {
+      chunkNum: 1,
+      data: defaultSortBy
+        ? this.doSort(data, defaultSortBy, defaultSortDirection)
+        : data,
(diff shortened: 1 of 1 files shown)


andyhuang91:
Do we want to hide the button or disable it? Also, should we continue to display the current page number?

SingularityUI/app/components/common/table/UITable.jsx
```diff
@@ -79,23 +75,24 @@ class UITable extends Component {
...
-    }
-    if (!updateStateAfterFetching) {
-      this.setState({chunkNum, rowChunkSize, sortBy, lastPage});
-    }
-    return this.props.fetchDataFromApi(chunkNum, rowChunkSize, sortBy).then(() => {
-      if (updateStateAfterFetching) {
-        this.setState({chunkNum, rowChunkSize, sortBy, lastPage});
+    const lookaheadNum = (chunkNum * rowChunkSize) + 1;
+    this.setState({ lookahead: true });
+    return this.props.fetchDataFromApi(lookaheadNum, 1).then((lookaheadResponse) => {
```
andyhuang91: Is this something that we always want to do for UITable or should this be configurable. You might also want to consider doing both of these requests in parallel and using `Promise.all`. That might make it unnecessary to separately keep track of the lookahead request state.

benheng: I think it makes sense in the current UITable design since it only applies to tables that have the `fetchDataFromApi` prop set. Do you foresee tables that have this prop set but won't have a lookahead call?

I had considered running the requests in parallel, but I thought it might be more complicated because the order of completion is undefined. Since the `data` prop of the UI table is supplied by the parent component and _that_ is supplied by the `requestApi` object.

benheng:
We're going to opt for better back-end support instead of making two API calls from the front-end.

PR #1707, opened 2018-02-12
Enable server side encryption params for uploads

ssalinas:
Enable usage of `SSEAwsKeyManagementParams` for S3 and `BlobWriteOption.encryptionKey` for GCS


baconmania:
🚢

PR #1708, opened 2018-02-13
Better catch for statusUpdate exceptions

ssalinas:
When enqueueing a status update to be processed asynchronously, if we hit the `maxStatusUpdateQueueSize` limit, a `RejectedExecutionException` is thrown _before_ `processStatusUpdateAsync` returns. The exception is not in the CompletableFuture.

This adds an additional catch and some better logging for this case.


baconmania:
🚢

PR #1709, opened 2018-02-13
Wire up slack notifications

kahliloppenheimer:
This is an attempt at implementing the ability for Singularity to notify via Slack, the same way it does via email.

This PR introduces the notion of the `SingularityIntercom`, which serves as a mechanism to broadcast notifications to all enabled `SingularityNotifier`s (i.e. `SmtpMailerNotifier` and `SlackNotifier`).

This implementation uses the same zk-backed email queue for all notifications now, selectively deciding which to execute when pulling them off of the queue.

The current way it's set up, email notifications _must_ be setup for any other form of notification to work, because currently the email logic dictates whether or not notifications get pushed into the zk queue.

Additionally, slack must be configured at the `SingularityConfiguration` level with a Slack token, but the actual per-request channel-notification overrides come in at the `SingularityRequest` level.


kahliloppenheimer:
cc @ssalinas @darcatron

ssalinas:
We are taking a bit of a different approach to notifications rework. Closing this for now

PR #1713, opened 2018-02-15
Fix mobile menu responsiveness

benheng:
The `nav` class on the `li` element was causing its height to fill the nav bar, preventing users from clicking the hamburger button. The hamburger button is only visible in views where the width is less than 768px.


andyhuang91:
🚢

PR #1714, opened 2018-02-15
removed unused path set

darcatron:
Pulled out the smaller update from https://github.com/HubSpot/Singularity/pull/1639. This will be easier to verify in our environments


ssalinas:
closing this, now folded into #1724

PR #1715, opened 2018-02-15
Use total system cpus, not totalCpus in system calculation

ssalinas:
When oversubscribing the resource, these values will be different. the load metrics are not as relevant without a true base value (systemCpus)


baconmania:
🚢

PR #1717, opened 2018-02-15
Collect cpu usage for tasks that have just started

ssalinas:
Right now we won't update the overall slave usage for a task that has just started. This lets us calculate the cpus from the task starting timestamp to include those as well in the slave usage metrics


baconmania:
🚢

PR #1718, opened 2018-02-15
Account for task level overrides in usage collection

ssalinas:
If a run now task overrides resources, the deploy level ones are no longer accurate. Account for that in the usage poller


baconmania:
🚢

PR #1719, opened 2018-02-16
Remove unneeded call to unsafeProcessStatusUpdate, fix tasksPerOfferHost check

ssalinas:
There was an extra line here processing the status update in duplicate outside of the lock. A few layers down, this can call `scheduleTasks` in the scheduler, resulting in a few possible duplicate taskIds


baconmania:
🚢

PR #1720, opened 2018-02-16
Account for expected usage when scoring offers

ssalinas:
Right now it's possible for us to easily overload a fairly empty machine if we have lots of tasks to schedule. The way the score is calculated, we don't account for other tasks we may have _just_ scheduled. This adds the reserved and last recorded max usage to factoring into scoring, updating those after each task is added to an offer host.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveUsageWithCalculatedScores.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveUsageWithCalculatedScores.java
@@ -2,40 +2,52 @@ package com.hubspot.singularity.mesos;
 
 import com.hubspot.singularity.SingularitySlaveUsage;
 import com.hubspot.singularity.SingularitySlaveUsage.ResourceUsageType;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveUsageWithCalculatedScores.java
```diff
@@ -77,8 +112,40 @@ SingularitySlaveUsage getSlaveUsage() {
...
+  private double getSystemLoadMetric() {
+    switch (systemLoadMetric) {
+      case LOAD_1:
+        return slaveUsage.getSystemLoad1Min();
+      case LOAD_15:
+        return slaveUsage.getSystemLoad5Min();
+      case LOAD_5:
+      default:
+        return slaveUsage.getSystemLoad15Min();
+    }
```
baconmania: Fat-finger switcheroo here

ssalinas: Thanks, fixed

baconmania:
🚢

PR #1721, opened 2018-02-17
Fix cached offer checkin in resourceOffers

ssalinas:
Need to add this in a traditional for loop so we are using the original object in the cache. Otherwise, when we call checkIn, we don't actually check in the original object in the cache and it can no longer be checked out again


baconmania:
🚢

PR #1724, opened 2018-02-20
 Uploader refactoring and additional attempts for immediate uploaders

ssalinas:
@darcatron I'm folding #1714 into this as well to avoid too many merge conflicts. Will close the other when they are both in hs_qa

There were a few cases where an immediate uploader could miss files. Particularly, if an uploader was immediate, but then something attempted to recreate it (wrote the file again), before the original one expired and was removed. This updates the uploader driver to:

- Trigger additional immediately upload attempts when the metadata is written to
- Check for any remaining files matching the glob on immediate uploader expiration, uploading them if some are found
- Refactor a number of things to java8-ify them and store immediate uploaders in a way that better matches regular uploaders

### DIFF
--- a/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
+++ b/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
@@ -274,7 +274,7 @@ public class SingularityS3UploaderDriver extends WatchServiceHelper implements S
         LOG.debug("Retrying immediate uploader {}", uploaderMetadata);
         performImmediateUpload(uploader);
       } else {
-        LOG.debug("Uploader for metadata {} not found to retry, recreating", uploaderMetadata);
+        LOG.debug("Uploader for metadata {} not found to retry upload", uploaderMetadata);
       }
     }
 
--- a/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
+++ b/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
@@ -74,7 +74,7 @@ public class SingularityS3UploaderDriver extends WatchServiceHelper implements S
   private final String hostname;
   private final SingularityRunnerExceptionNotifier exceptionNotifier;
 
-  private final Map<S3UploadMetadata, SingularityUploader> metadataToimmediateUploader;
+  private final Map<S3UploadMetadata, SingularityUploader> metadataToImmediateUploader;
   private final Map<S3UploadMetadata, CompletableFuture<Integer>> immediateUploadersFutures;
 
   private ScheduledFuture<?> future;
@@ -109,7 +109,7 @@ public class SingularityS3UploaderDriver extends WatchServiceHelper implements S
     this.exceptionNotifier = exceptionNotifier;
 
     this.immediateUploadersFutures = new ConcurrentHashMap<>();
-    this.metadataToimmediateUploader = new ConcurrentHashMap<>();
+    this.metadataToImmediateUploader = new ConcurrentHashMap<>();
   }
 
   private void readInitialFiles() throws IOException {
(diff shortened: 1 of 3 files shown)


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
```diff
@@ -76,9 +74,8 @@
   private final String hostname;
   private final SingularityRunnerExceptionNotifier exceptionNotifier;
 
-  private final List<S3UploadMetadata> immediateUploadMetadata;
-  private final ReentrantLock lock;
-  private final ConcurrentMap<SingularityUploader, Future<Integer>> immediateUploaders;
+  private final Map<S3UploadMetadata, SingularityUploader> metadataToimmediateUploader;
```
darcatron: nit `metadataToimmediateUploader` -> `metadataToImmediateUploader` (capitalize immediate)

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
```diff
@@ -256,31 +268,35 @@ private int checkUploads() {
...
-    for (SingularityUploader uploader : toRetry) {
-      LOG.debug("Retrying immediate uploader {}", uploader);
-      performImmediateUpload(uploader);
+    for (S3UploadMetadata uploaderMetadata : toRetry) {
+      SingularityUploader uploader = metadataToimmediateUploader.remove(uploaderMetadata);
+      if (uploader != null) {
+        LOG.debug("Retrying immediate uploader {}", uploaderMetadata);
+        performImmediateUpload(uploader);
+      } else {
+        LOG.debug("Uploader for metadata {} not found to retry, recreating", uploaderMetadata);
```
darcatron: not sure I understand this debug line. What is being re-created?

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
```diff
@@ -256,31 +268,35 @@ private int checkUploads() {
...
     }
 
-    for (SingularityUploader uploader : toRetry) {
-      LOG.debug("Retrying immediate uploader {}", uploader);
-      performImmediateUpload(uploader);
+    for (S3UploadMetadata uploaderMetadata : toRetry) {
+      SingularityUploader uploader = metadataToimmediateUploader.remove(uploaderMetadata);
+      if (uploader != null) {
+        LOG.debug("Retrying immediate uploader {}", uploaderMetadata);
+        performImmediateUpload(uploader);
```
darcatron: For this retry loop, the uploader is removed from `metadataToImmeditateUploader` and when it's called to upload immediately, it'll be added to the `immediteUploadersFutures` map. During the next run of `checkUploads`, won't the uploader be missing from `metadataToImmediateUploader` and cause the check to skip adding the uploadedFiles to the total count? Instead it'll be moved to the `toRemove` list and hit the 30 second check

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
ssalinas: fixed

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
ssalinas: Ah, copy-pasta error. the ,recreating should be removed, if we don't find the uploader in the map it means it somehow got added to both toRemove and toRetry, which theoretically should never happen, but better than throwing a NPE, updated the message

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
ssalinas: good catch. It ends up working out that it still retries, but that should be a .get so we can do the full retry -> toRemove loop . Fixed

darcatron:
lgtm 🚢

PR #1726, opened 2018-02-20
Don't operate directly on pending tasks during statusUpdate

ssalinas:
This was causing a race condition with the new locking setup where a pending task could be selected for offer scheduling, then deleted by statusUpdate while it was still being processed for offers. This PR updates the code to:
- Save the pending request to the queue instead of directly calling scheduleTasks
- Update tests that check pending tasks to call `scheduler.drainPendingQueue()` after calling `statusUpdate`

/cc @baconmania


baconmania:
🚢

PR #1727, opened 2018-02-20
Allow user to configure which load metric is used for offer scoring

ssalinas:
We collect the 1, 5, and 15 minute load. Right now it just uses 15 by default. This allows the user to configure which is used and changes the default to 5 minute


ssalinas:
folded into #1720 , closing

PR #1735, opened 2018-02-22
 Add the ability to shuffle long running tasks when a host is overloaded

ssalinas:
It is very common to oversubscribe on resources to use fewer large hosts more efficiently. Unfortunately that can have unintended consequences like overloading a host's cpu or oom-ing if memory is oversubscribed too much. This PR tackles the first of those cases (cpu). During our usage polling, we will check the system load on the machine (as reported by the mesos slave api). If it is > 1.0, this is indicative of the machine being overloaded and we will attempt to create cleanups on high cpu usage long running tasks on the host in order to disperse the load to hosts with more free resources. We will only allow up to 3 cleaning tasks per host by default, and will stop creating cleanups if we hit that limit or believe the usage of the cleaned tasks accounts for enough of the cpu to get the host back to acceptable levels

This is akin to mesos' revocable resources, however fully integrating those into Singularity would be a much larger lift for similar functionality

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
@@ -221,7 +221,7 @@ public class SingularityUsagePoller extends SingularityLeaderOnlyPoller {
           if (slaveOverloaded && configuration.isShuffleTasksForOverloadedSlaves() && currentUsage != null && currentUsage.getCpusUsed() > 0) {
             if (isLongRunning(task)) {
               Optional<SingularityTaskHistoryUpdate> maybeCleanupUpdate = taskManager.getTaskHistoryUpdate(task, ExtendedTaskState.TASK_CLEANING);
-              if (maybeCleanupUpdate.isPresent() && isTaskShuffleCleanup(maybeCleanupUpdate.get())) {
+              if (maybeCleanupUpdate.isPresent() && isTaskAlreadyCleanedUpForShuffle(maybeCleanupUpdate.get())) {
                 LOG.trace("Task {} already being cleaned up to spread cpu usage, skipping", taskId);
                 shuffledTasks++;
               } else {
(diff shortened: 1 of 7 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -183,12 +211,48 @@ public void runActionOnPoll() {
...
+                possibleTasksToShuffle.add(new TaskIdWithUsage(task, currentUsage));
+              }
+            }
+          }
+        }
+
+        if (slaveOverloaded && configuration.isShuffleTasksForOverloadedSlaves()) {
+          possibleTasksToShuffle.sort((u1, u2) -> Double.compare(u2.getUsage().getCpusUsed(), u1.getUsage().getCpusUsed()));
+          for (TaskIdWithUsage taskIdWithUsage : possibleTasksToShuffle) {
+            if (cpuOverage <= 0 || shuffledTasks > configuration.getMaxTasksToShuffleForCpuOverage()) {
```
darcatron: if this line is reached then, `slaveOverloaded` must be `true`. Since `slaveOverloaded` is only true if `used > total`, then it should be an error if `cpuOverage <= 0`. If you want to keep the check, it might be worth putting in an error log

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -183,12 +211,48 @@ public void runActionOnPoll() {
...
             usageManager.saveCurrentTaskUsage(taskId, currentUsage);
 
             cpusUsedOnSlave += taskCpusUsed;
           }
+
+          if (slaveOverloaded && configuration.isShuffleTasksForOverloadedSlaves() && currentUsage != null && currentUsage.getCpusUsed() > 0) {
+            if (isLongRunning(task)) {
+              Optional<SingularityTaskHistoryUpdate> maybeCleanupUpdate = taskManager.getTaskHistoryUpdate(task, ExtendedTaskState.TASK_CLEANING);
+              if (maybeCleanupUpdate.isPresent() && isTaskShuffleCleanup(maybeCleanupUpdate.get())) {
+                LOG.trace("Task {} already being cleaned up to spread cpu usage, skipping", taskId);
```
darcatron: nit: I think `isTaskShuffleCleanup` is a little confusing. Maybe `isTaskAlreadyShuffled` or something similar?The trace also implies it's currently being cleaned up, but the method checks the historical statuses so it's possible it was already shuffled. For future debugging it might be good to have the log line note that. E.g. `Task {} is being or was already cleaned up to spread cpu usage, skipping"

darcatron:
:ship:

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
ssalinas: cpuOverage is mutable here inside the for loop. We decrement it for each task that we move off of the slave and this check breaks us out if we either removed enough cpu, or hit our max number of tasks to shuffle at once

darcatron: ah, i see the update now. my mistake

darcatron:
🚢

PR #1736, opened 2018-02-23
Upgrade swagger and use open api 3.0

ssalinas:
This PR is step 1 in reorganizing SingularityBase and making the whole project more usable

- Adds a large pile of missing documentation
- Upgrades swagger to 2.x versions which use openapi 3.0 spec
- serves up the openapi.json via an endpoint rather than compiling markdown into the jar

TODO:
- [x] finish off annotation and docs updates
- [x] set up swagger-ui-react to read the openapi.json within the ui


ssalinas:
Filed https://github.com/swagger-api/swagger-core/issues/2740 for the circular reference issue we ran into. Will try to make a PR for that later if I have time

SingularityUI/app/components/apidocs/ApiDocs.jsx
```diff
@@ -0,0 +1,42 @@
...
+import rootComponent from '../../rootComponent';
+import { refresh } from '../../actions/ui/apidocs';
+
+import SwaggerUi, {presets} from 'swagger-ui';
+import 'swagger-ui/dist/swagger-ui.css';
+
+class ApiDocs extends Component {
+
+  componentDidMount() {
+    SwaggerUi({
```
ssalinas: @kwm4385 no rush, but if you get a chance I'm a bit stumped. This works locally when running with gulp serve, pointed against a local singularity or against our staging environment. But, the deployed version doesn't work. Comes back with errors of `Expected the reducer to be a function`.

pschoenfelder:
🚢

PR #1740, opened 2018-02-27
Only send healthcheck object from ui when setting all fields

### LINKED ISSUE #1739: Health Checks issue
If I want to deploy new task using UI, for request type WORKER, ui only allows setting timeout after task is considered healthy.  
image

However, my deploy fails to pass healthcheck:  

image

What does healthcheck check after all? That's just a worker, not listening to any port.

ssalinas:
also fix the validator to catch this case. This will need to be back-ported to 0.17 and 0.18

/fixes #1739


baconmania:
🚢

PR #1743, opened 2018-02-28
Add 0.17.1/019.1 release docs

ssalinas:
Backporting the fix to 0.17 for those who have not done the mesos 1 upgrade. Skipping 0.18 since it is the same mesos version as 0.19


baconmania:
🚢

PR #1748, opened 2018-02-28
Permalink request within group

benheng:
Previously, request groups didn't support permalinks for requests within the group. It would simply select the first request and display that.

You'd have link that looked like this:
```
localhost/singularity/ui/group/RequestGroup1
```

Which would contain some requests: RequestGroup1-foo1, RequestGroup1-foo2, RequestGroup1-foo3

Now you can link them directly:
```
localhost/singularity/ui/group/RequestGroup1/RequestGroup1-foo1
localhost/singularity/ui/group/RequestGroup1/RequestGroup1-foo2
localhost/singularity/ui/group/RequestGroup1/RequestGroup1-foo3
```

The default behavior is maintained, no request ID means display the first request. Incorrect request ID will still load the selected group page, but the request data will not be displayed since there is none.

### DIFF
--- a/SingularityUI/app/components/groupDetail/GroupDetail.jsx
+++ b/SingularityUI/app/components/groupDetail/GroupDetail.jsx
@@ -1,6 +1,6 @@
-import React, {PropTypes, Component} from 'react';
+import React, { PropTypes } from 'react';
 import { connect } from 'react-redux';
-import { LinkContainer } from 'react-router-bootstrap';
+import { withRouter, routerShape } from 'react-router';
 import rootComponent from '../../rootComponent';
 
 import { Row, Col, Nav, NavItem, Label } from 'react-bootstrap';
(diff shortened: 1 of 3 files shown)


SingularityUI/app/components/groupDetail/GroupDetail.jsx
```diff
@@ -1,5 +1,6 @@
 import React, {PropTypes, Component} from 'react';
 import { connect } from 'react-redux';
+import { LinkContainer } from 'react-router-bootstrap';
```
kwm4385: What's the reasoning behind using `LinkContainer` over the standard react-router `Link` or an onclick handler?

benheng: `Link` doesn't play nice with bootstrap components. I also wanted the URL to change when a user clicked on a request ID which is why I went with `LinkContainer`. Would it have been better to have an onClick handler and wrap the component using `withRouter`?

benheng: `Link` doesn't play nice with bootstrap components. I also wanted the URL to change when a user clicked on a request ID which is why I went with `LinkContainer`. Would it have been better to have an onClick handler and wrap the component using `withRouter`?

SingularityUI/app/components/groupDetail/GroupDetail.jsx
```diff
@@ -50,9 +51,15 @@ class GroupDetail extends Component {
 
             <Nav bsStyle="pills" stacked={true} activeKey={this.state.showRequestId} onSelect={this.handleRequestSelect}>
               {group.requestIds.map((requestId, index) =>
-                <NavItem className="request-group-navitem" key={index} eventKey={requestId}>
-                  {requestId} {requestsNotFound.includes(requestId) && <Label bsStyle="danger">Deleted</Label>}
-                </NavItem>
+                <LinkContainer
+                  eventKey={requestId}
+                  key={index}
```
andyhuang91: `requestId` is a logical key here

SingularityUI/app/components/groupDetail/GroupDetail.jsx
```diff
@@ -13,7 +14,7 @@ class GroupDetail extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      showRequestId: _.first(props.group.requestIds)
+      showRequestId: props.params.requestId || _.first(props.group.requestIds),
```
andyhuang91: component state isn't needed anymore, and you can get rid of `onSelect={this.handleRequestSelect}` on the Nav component

SingularityUI/app/components/groupDetail/GroupDetail.jsx
kwm4385: If we were already using `react-router-bootstrap` I'd say this is the best solution, however, I'm hesitant to add another dependency for this single use case since we already have so many. For that reason I'd say go with handling it via `onClick` and `withRouter`.

kwm4385:
🚢

PR #1750, opened 2018-02-28
WIP -  Move to java.util.Optional and use immutables

ssalinas:
This is a WIP built on top of #1736 . It includes a number of updates around getting us to the latest basepom version, swagger updates, cleanup of dozens of deprecation warnings, a move to java.util.Optional, and (once I finish it) a move of many SingularityBase models to use [immutables](https://immutables.github.io/).

We are hoping that this update to docs and to the SingularityBase objects and organization of the packages will be a one time breaking change (for java client users), that will allow us to move more quickly and lower the barrier to entry for other contributors in the future.

The aim of all of the updates to the models is to promote ease of use, without changing the current json. There should be no migrations or deprecations needed in terms of the structure of the json returned and consumed by Singularity.

cc @darcatron @pschoenfelder @baconmania


ssalinas:
Step 2, restructure SingularityBase so it isn't one giant package. Full list of changes at https://gist.github.com/ssalinas/4ac6d6dfe939251b9e7b8e9a352c37ab . I'll write a little script in there to do easy find/replace in directories

ssalinas:
Putting this on hold for the moment in favor of just the swagger upgrade in the previous PR. I'm think to minimize risk in using immutables, it may end up being wiser to only convert those that give us more immediate benefit (i.e. ones with giant constructors and many fields like SingularityState), then slowly redo the rest

PR #1754, opened 2018-03-06
Add a new scoring mode accounting for max historical usage of all tasks

ssalinas:
This adds a new strategy for offer scoring that will additionally take into account:
- the max historical recorded usage for all active tasks currently on a slave
- the system load if it is higher than the predicted maximum
- the estimated max usage of tasks about to be added to the host

It also makes the free/used resource weights for long/non-long running tasks configurable in the yaml


pschoenfelder:
🚢

PR #1755, opened 2018-03-13
Catch exceptions in offer scoring

ssalinas:
We are seeing a few instances where we are getting stuck on the `allOf(...).join()`. Best guess is that async semaphore isn't completing properly for this case. This adds exception handling within each future, setting a reference to the throwable that will then be thrown back up, causing singularity to abort (as it would normally do for uncaught exceptions in scoring). This isn't a fix for the underlying bug yet, but I'm hoping this can tease out the root cause of the exception, since we are currently swallowing these exceptions

cc @baconmania


pschoenfelder:
🚢

PR #1757, opened 2018-03-19
Don't count pending requests towards underprovisioning

pschoenfelder:


SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
```diff
@@ -195,6 +198,7 @@ public SingularityState generateState(boolean includeRequestIds) {
       updatePossiblyUnderProvisionedAndOverProvisionedIds(requestWithState, numInstances, overProvisionedRequestIds, possiblyUnderProvisionedRequestIds);
     }
 
+    removePendingRequestIds(possiblyUnderProvisionedRequestIds);
```
ssalinas: can we give this a better name? Like `filterForPendingRequests` or something more descriptive like that?

SingularityService/src/test/java/com/hubspot/singularity/data/StateManagerTest.java
```diff
@@ -53,4 +57,21 @@ public void itDoesntCountCleaningTasks() {
...
   }
+
+  @Test
+  public void itDoesntFlagPendingRequestsForUnderOrOverProvisioning() {
+    initRequest();
+    initFirstDeploy();
+
+    SingularityRequest request = requestResource.getRequest(requestId, singularityUser).getRequest();
+
+    requestManager.activate(request.toBuilder().setInstances(Optional.of(0)).build(), RequestHistoryType.UPDATED, System.currentTimeMillis(), Optional.<String> absent(), Optional.<String> absent());
```
ssalinas: did you confirm that this test actually fails without the filter you added? Seems odd to have it at instances of `0` (tbh I thought our validation would catch that...)

pschoenfelder:
Yeah, I commented out my original code, wrote the failing test, then
uncommented the code and it passed.

ssalinas:
Cool, let's move this one through the environments and get it merged. 🚢

PR #1758, opened 2018-03-21
Enable framework auth over http api

ssalinas:
We were missing a few extra things to authenticate over the http api as compared with the older scheduler driver. mesos rxjava looks like it only supports [basic auth](https://github.com/mesosphere/mesos-rxjava/blob/master/mesos-rxjava-client/src/main/java/com/mesosphere/mesos/rx/java/MesosClient.java#L422-L423) at the moment, so that is now wired up to the `mesosUsername` and `mesosPassword` parameters. `credentialPrincipal` is wired up properly to FrameworkInfo for those that want to use it for authorization as well.


pschoenfelder:
🚢

PR #1759, opened 2018-03-21
Fixed a logging statement

dalazx:
```
DEBUG [2018-03-21 14:56:31,242] com.hubspot.singularity.scheduler.SingularityDeployHealthHelper: Task task-1-1521644025943-1-10.38.0.0-DEFAULT has been running for 00:27.903, has not yet reached running threshold of 00:00.030
```
`00:00.030` looks wrong. should've been `00:30.000` instead.


ssalinas:
🚢  thanks for the fix! I'll merge it through our environments and get it in the next release

PR #1760, opened 2018-03-22
Sort tasks to shuffle by overusage, not usage

ssalinas:
This updates the choice of which task to shuffle to focus on those that are using far more than their requested resources. For example, a worker that requested 1 resource and is using 2 will be chosen over another worker that requested 10 and is using 11.


baconmania:
🚢

Issue #1737, opened 2018-02-23
Paused request should remain paused after new deploy


yuriy-filatov:
Hey there.
I'll try to explain the cause of this request.
We're running Singularity 0.18.2 to orchestrate our mesos jobs in Production.
We have 200+ requests, several of them are paused.
Sometimes we have to roll out changes to all existing jobs - it can be either new link to some new endpoint or any other dynamic data. 
We're experiencing quite massive pita in current setup - all paused requests become active immediately after new deploy.
I do understand that this request might broke internal state machine, but is it possible to add override key in the settings somewhere? This will definitely save us some time and hair.
Thanks.

ssalinas:
The reason we don't have this currently is more around the health checks for a request. If a request is paused and we accept the deploy as a success, we have no verification that the new deploy is viable/will work when unpaused. However, we have had a few requests for similar functionality internally, so it's definitely something we could look into supporting in future releases

yuriy-filatov:
Well, that's why we have deploy checks, eh. If request is non-deployable - it should not be sent to Singularity at all.

Please let me know if this will be included into some incoming release.

Thank you.

yuriy-filatov:
Hey @ssalinas 
Hope my message finds you well.
I've tried to use 'allowDeployOfPausedRequests' -- but it doesn't seem to work.
I've added 'allowDeployOfPausedRequests=true' into startup script (along with working 'hostname' option for example) -- nothing changed in s9y behavior.
Can you please check/explain how to use that option?

ssalinas:
Hmm, I can get it to work on our staging env. When you POST the deploy, are you getting the conflict message still like before? First guess is that the configuration option isn't being picked up for some reason. You guys use the docker image to run Singularity correct? Can try and confirm with that as well to see if it's maybe the jar in the docker image is incorrect or something

yuriy-filatov:
I've verified s9y shaded jar md5 inside running container - it's definitely 0.19.2.
I've tried to pass options through:
- startup shell script insode docker which we use for numerous dynamic options which differ from env to env alike 
``#!/bin/bash
exec /usr/bin/java -Ddw.ui.title="${UI_TITLE}" -Ddw.ui.navColor="${UI_NAVCOLOR}" -Ddw.hostname=${HOSTNAME} -Ddw.allowDeployOfPausedRequests=true -jar SingularityService.jar server singularity-config.yaml``
- plain option in singularity-config.yaml
None of these work for us.

[...]

Issue #1640, opened 2017-10-26
s3 logs missing from ui when no group overrides set in config


jaredstehler:
`s3. groupOverrides` isn't listed as required for singularity service but when not specified, the endpoint `/api/logs/task/{taskid}` returns 

```
{
    "code": 404,
    "message": "S3 configuration was absent"
}
```

this is because the provider bails early checking for it:

https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/SingularityMainModule.java#L255

https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/resources/S3LogResource.java#L535

workaround is to set a dummy group override in config..

Issue #1512, opened 2017-04-20
Allow the log level for the `com.hubspot.singularity` logger to be overridden from the container start script?


demobox:
The default service configuration sets the log level of the `com.hubspot.singularity` logger [to `TRACE`](https://github.com/HubSpot/Singularity/blob/master/SingularityService/singularity.yml#L32), which is quite verbose and can produce very large log files if the service is running for a long time.

While the [container start script](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/docker/start.sh) allows many of the configuration parameters to be overridden, the log level for this logger is not one of the values that can be changed.

Assuming that it's even possible to override a DropWizard key/value pair setting via system properties (especially one whose key contains `.` characters - the DropWizard docs don't seem clear on that), would it be possible to make this configuration property overridable too?

Happy to put together a PR if this sounds like a useful idea.

ssalinas:
Yeah, I think we need a better setup in general for the docker conatiner, such that we can configure anything in the yaml file, and not have to keep updating it when we add new configuration fields. Internally, we only use it for development, so it hasn't gotten as much attention for extensive configurability.

I could see hooking up something like http://dropwizard.readthedocs.io/en/latest/manual/core.html#man-core-environment-variables so that we can override any property via environment variables instead.

and PRs always welcome ;)

PtrTeixeira:
An alternative is just to mount a new config file to `/etc/singularity/singularity.yaml` when you run the container. That'll just overwrite the default config file

demobox:
> when you run the container. That'll just overwrite the default config file

That's the option we've considered as a short-term fix for this, but in general we'd also like to contribute to making this easier "out of the box", if possible.

Thanks for the suggestion!

PR #1762, opened 2018-03-23
Various bug fixes and updates

ssalinas:
/fixes https://github.com/HubSpot/Singularity/issues/1737 (can be configured to allow paused request deploys, false by default)
/fixes https://github.com/HubSpot/Singularity/issues/1640
/fixes https://github.com/HubSpot/Singularity/issues/1512


pschoenfelder:
🚢

PR #1763, opened 2018-03-23
Support SSE S3 in SingularityUploader

leoxlin:
- This enables a flag to turn on the SSE-S3 which transparently encrypts the data on the server side and is managed by S3.
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html

- We've tested the object metadata flag to make it indeeds sets SSE-S3 when we inspect the uploaded object on S3. This follows the S3 java-sdk documentation.
https://docs.aws.amazon.com/AmazonS3/latest/dev/SSEUsingJavaSDK.html

- I am unfamiliar with the naming convention for the JSON Values so open to change it to something more suitable for our convention so far.

@baconmania @ssalinas

### DIFF
--- a/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/S3UploadMetadata.java
+++ b/SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/S3UploadMetadata.java
@@ -60,6 +60,7 @@ public class S3UploadMetadata {
   private final Map<String, Object> gcsCredentials;
   private final Optional<String> gcsStorageClass;
   private final Optional<String> encryptionKey;
+  private final Optional<Boolean> encryptionS3;
 
   @JsonCreator
   public S3UploadMetadata(@JsonProperty("directory") String directory,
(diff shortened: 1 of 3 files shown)


SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/S3UploadMetadata.java
```diff
@@ -79,7 +80,8 @@ public S3UploadMetadata(@JsonProperty("directory") String directory,
                           @JsonProperty("uploaderType") Optional<SingularityUploaderType> uploaderType,
                           @JsonProperty("gcsCredentials") Map<String, Object> gcsCredentials,
                           @JsonProperty("gcsStorageClass") Optional<String> gcsStorageClass,
-                          @JsonProperty("encryptionKey") Optional<String> encryptionKey) {
+                          @JsonProperty("encryptionKey") Optional<String> encryptionKey,
+                          @JsonProperty("s3ServerSideEncryption") Optional<Boolean> s3ServerSideEncryption) {
```
ssalinas: Can we name this something more like `useS3ServerSideEncryption`? Makes it a bit more obvious it's a boolean and not a string to specify a type of encryption. Other than that PR looks good 👍

ssalinas:
🚢 I'll move this through to staging/qa/etc

SingularityRunnerBase/src/main/java/com/hubspot/singularity/runner/base/shared/S3UploadMetadata.java
```diff
@@ -217,6 +220,10 @@ public SingularityUploaderType getUploaderType() {
     return encryptionKey;
   }
 
+  public boolean getUseS3ServerSideEncryption() {
```
ssalinas: sry, missed this earlier. Can this be `isUseS3ServerSideEncryption`

pschoenfelder:
🚢

PR #1766, opened 2018-03-27
More usage collection in parallel, less webhooks in parallel

### LINKED ISSUE #1765: Rate limiting and parallelism limits for webhook delivery
We just had an incident where a misconfigured webhook (pointed at a service that no longer existed) trivially takes down a Singularity in a way that neither crashes nor is recoverable.  The file descriptor table (left at a small-ish default 4096 entries) filled entirely with webhook dispatches, and the rate that they were created exceeded the rate that they failed.   This led to many IOExceptions and SocketExceptions saying, 'out of file descriptors'.

The same webhook should have a limited number of parallel requests to it, and possibly some kind of circuit breaker to protect against failed webhooks.

Different webhooks should be reasonably separated so that one failure cannot affect others

The total number of webhook dispatches should be limited so an administrator can ensure it does not bring down the rest of Singularity

ssalinas:
/fixes #1765

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/hooks/SingularityWebhookSender.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/hooks/SingularityWebhookSender.java
@@ -1,7 +1,11 @@
 package com.hubspot.singularity.hooks;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import javax.inject.Singleton;
@@ -14,6 +18,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.inject.Inject;
 import com.hubspot.mesos.JavaUtils;
 import com.hubspot.singularity.SingularityDeployUpdate;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/hooks/SingularityWebhookSender.java
```diff
@@ -60,23 +74,30 @@ public void checkWebhooks() {
     int requestUpdates = 0;
     int deployUpdates = 0;
 
-    for (SingularityWebhook webhook : webhooks) {
+    List<CompletableFuture<Void>> webhookFutures = new ArrayList<>();
```
baconmania: Is this list ever written to?

SingularityService/src/main/java/com/hubspot/singularity/hooks/SingularityWebhookSender.java
```diff
@@ -39,13 +47,19 @@
...
   @Inject
   public SingularityWebhookSender(SingularityConfiguration configuration, AsyncHttpClient http, ObjectMapper objectMapper, TaskHistoryHelper taskHistoryHelper, WebhookManager webhookManager) {
     this.configuration = configuration;
     this.http = http;
     this.webhookManager = webhookManager;
     this.taskHistoryHelper = taskHistoryHelper;
     this.objectMapper = objectMapper;
+
+    this.webhookSemaphore = AsyncSemaphore.newBuilder(configuration::getMaxConcurrentWebhooks).build();
+    this.webhookExecutorService = Executors.newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat("webhooks-%d").build());
```
baconmania: Doesn't look like this guy is used anywhere after being created.

baconmania:
🚢

pschoenfelder:
🚢

stevenschlansker:
Thanks!

PR #1767, opened 2018-03-27
Periodically flush the queue to make sure batch work does not get stuck

ssalinas:
Also adds better polling of the queue when reusing permits.

@baconmania @pschoenfelder


baconmania:
🚢

pschoenfelder:
🚢

PR #1768, opened 2018-03-27
Keep track of offers not accounted for in SingularityOfferHolders returned by checkOffers

ssalinas:
We've occasionally seen cases where an offer remains in CHECKED_OUT state after checkOffers is run. There is a possibility that checkOffers does not return an offer holder for every offer. In this case, the cached offers still need to be returned to the cache


pschoenfelder:
🚢

PR #1769, opened 2018-03-29
Support placeholders in webhook URIs.

baconmania:


ssalinas:
FYI I merged #1766 so you only have to fix the conflicts once

ssalinas:
Code looks fine, we should add a short blurb to the docs in `Docs/reference/webhooks.md` that this is possible though. Other than that , 👍

ssalinas:
🚢

PR #1770, opened 2018-03-29
Support a configurable delay for task shuffles.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -280,6 +280,14 @@ public void runActionOnPoll() {
     }
   }
 
+  private boolean isEligibleForShuffle(SingularityTaskId task) {
+    return (
+        isLongRunning(task)
+            && !configuration.getDoNotShuffleRequests().contains(task.getRequestId())
+            && (TimeUnit.MILLISECONDS.toMinutes(System.currentTimeMillis() - task.getStartedAt()) > configuration.getMinutesBeforeNewTaskEligibleForShuffle())
```
ssalinas: I wonder if we should go by the task running timestamp instead? Take out any delay that may have been caused by mesos/singularity that way

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -15,6 +15,8 @@
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
+import javax.ws.rs.HEAD;
```
ssalinas: leftover from merge conflict? (intellij auto import stuff?)

baconmania: Good catch.

ssalinas:
🚢

PR #1773, opened 2018-04-03
Retry lost tasks

pschoenfelder:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -175,6 +176,10 @@ private void saveNewTaskStatusHolder(SingularityTaskId taskIdObj, SingularityTas
     return Optional.absent();
   }
 
+  private void relaunchTask(SingularityTask task) {
+    taskManager.savePendingTask(task.getTaskRequest().getPendingTask());
```
pschoenfelder: Is this actually how to put something on the pending queue? Not convinced this is correct

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -200,15 +205,35 @@ private void unsafeProcessStatusUpdate(Protos.TaskStatus status, SingularityTask
...
+          || status.getReason() == Reason.REASON_AGENT_REMOVED
+          || status.getReason() == Reason.REASON_AGENT_RESTARTED
+          || status.getReason() == Reason.REASON_AGENT_UNKNOWN
+          || status.getReason() == Reason.REASON_MASTER_DISCONNECTED
+          || status.getReason() == Reason.REASON_AGENT_DISCONNECTED;
+
+      RequestType requestType = task.isPresent() ? task.get().getTaskRequest().getRequest().getRequestType() : null;
+      boolean isRelaunchable =
+          requestType != null
+              && (requestType == RequestType.ON_DEMAND
```
ssalinas: You can just use !isLongRunning here instead of having to specify all 3

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -200,15 +205,35 @@ private void unsafeProcessStatusUpdate(Protos.TaskStatus status, SingularityTask
...
+                  || requestType == RequestType.SCHEDULED
+                  || requestType == RequestType.RUN_ONCE);
+
+      if (isMesosFailure && isRelaunchable) {
+        LOG.info("Relaunching lost task {}", task);
+        relaunchTask(task.get());
+      } else {
+        lostTasksMeter.mark();
```
ssalinas: we'll still want the lost task meter to fire regardless of failure type. The lost task metric allows us to do things like alert on a large wave of these (indicating that something is wrong with singularity. Similar with the disaster detection piece below

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
ssalinas: I'm iffy on having this here. This is partially because we let the status update handler write to pending tasks, and partially because it means we are reusing a pending task id. Generally we let the SingularityScheduler do all of the work of creating a pending task to keep responsibility for those types of operations separate. We actually removed bits from the status update handler a little while back so that it would avoid mutating the pending task queue.

Instead I'd suggest using `requestManager` to add to the pending request queue. This will let the scheduler do it's normal thing and rebuild a full new pending task with new unique ID from that pending request

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
pschoenfelder: Won't this still cause the alerts that the original ticket wanted to resolve then?

ssalinas: There weren't alerts in the original issue. The original issue was that a failure that was mesos'/singularity's fault was counting towards the retry count for a scheduled task. i.e. it was given 2 attempts, both failed due to the invalid offers reason. So, even though the task wasn't the thing that failed, it wasn't retried again

pschoenfelder:
Where is that set of unit tests?

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityRetryLostJobsTest.java
```diff
@@ -0,0 +1,77 @@
...
+      Assert.assertTrue(false);
+    }
+
+    if (shouldRetry) {
+      Assert.assertEquals(requestManager.getPendingRequests().size(), 1);
+      Assert.assertEquals(requestManager.getPendingRequests().get(0).getPendingType(), PendingType.RETRY);
+    } else {
+      System.out.println(requestManager.getPendingRequests());
+      Assert.assertEquals(requestManager.getPendingRequests().size(), 0);
+//      Assert.assertEquals(requestManager.getPendingRequests().get(0).getPendingType(), PendingType.TASK_DONE);
```
pschoenfelder: I think I'm misunderstanding something because one of the two negative test cases fails here. The two negative test cases have different behaviors. One, `itDoesNotRetryLostRequestsDueToNonAgentFailures`, results in an empty pending queue, which is what I'd expect. The other, `itDoesNotRetryLostLongRunningRequests`, winds up with a request in the pending queue, even though I've manually verified through the debugger that the new "relaunch task" code path is never hit. This request also winds up with a `PendingType` of `TASK_DONE` which is definitely suspicious. Before I spend more time debugging, is there some obvious behavior I'm missing?

ssalinas: So, that's normal that TASK_DONE is enqueued. What's different is that a TASK_DONE for a scheduled job won't necessarily result in a new task being started right away. We want to make sure we have our own pending request of a different type, that will trigger the desired behavior vs TASK_DONE which would do things like just schedule the next interval for a scheduled task instead of scheduling right away (and for an ON_DEMAND I believe TASK_DONE would be a no-op)

PR #1775, opened 2018-04-04
Add an optional cpu hard limit

ssalinas:
If a hard limit on cpu is not enabled on the mesos slave/agent, cpu usage is only controlled by cpu shares. It can be useful to also add an additional hard limit, such that a single task cannot overwhelm a slave. This adds a `cpuHardLimit` and `cpuHardLimitScaleFactor` configuration parameter in the base SingularityConfiguration, which will:

- propagate a hard limit to ExecutorData for the task. If the hard limit is below the requested task cpus, that task's new hard limit is requested cpus * cpuHardLimitScaleFactor
- triggers the executor to echo a calculated cfs quota and cfs period (defaults to the recommended 100000) to the cgroup settings for the task in runner.sh. For docker tasks, this adds the --cpu-quota/--cpu-period flags to the docker run command

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
@@ -327,7 +327,9 @@ public class SingularityExecutorMonitor {
             SingularityTaskExecutorData taskExecutorData = (SingularityTaskExecutorData) task.getExecutorData();
             if (taskExecutorData.getCpuHardLimit().isPresent()) {
               try {
-                cgroupCheckers.put(task.getTaskId(), new SingularityExecutorCgroupCfsChecker(task, taskExecutorData.getCpuHardLimit().get(), configuration.getDefaultCfsPeriod()));
+                SingularityExecutorCgroupCfsChecker cfsChecker = new SingularityExecutorCgroupCfsChecker(task, taskExecutorData.getCpuHardLimit().get(), configuration.getDefaultCfsPeriod());
+                cfsChecker.watch();
+                cgroupCheckers.put(task.getTaskId(), cfsChecker);
               } catch (Throwable t) {
                 LOG.error("Could not start cgorup checker for task {}", task.getTaskId(), t);
               }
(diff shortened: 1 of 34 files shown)


ssalinas:
Additionally started tracking cpu throttled time in the metrics and reporting it in the UI. Also fixed our UI bug where the cpu resource usage wasn't updating frequently enough

ssalinas:
Interesting tidbit, the nr_throttled and nr_periods do not actually get collected by the mesos slave when `cgroups_enable_cfs` is not set

SingularityExecutor/src/main/resources/docker.sh.hbs
```diff
@@ -137,6 +137,19 @@ PARENT_CGROUP=$(get_deepest_cgroup)
 DOCKER_OPTIONS+=(--name=$CONTAINER_NAME)
 DOCKER_OPTIONS+=(--cgroup-parent=$PARENT_CGROUP)
 DOCKER_OPTIONS+=(--env-file=docker.env)
+{{#if runContext.cfsQuota }}
+function get_base_cgroup_directory {
+  if [ -d "/cgroup" ]; then
+    echo "/cgroup"
```
PaulFurtado: These are just mountpoints and people/distros are free to mount the cgroup controllers anywhere. You can get this robustly with findmnt from util-linux (which is installed by default on pretty much every distro).

```
findmnt --kernel --first-only --types cgroup --options cpu --noheadings --output TARGET
```
will output /cgroup/cpu on a HubSpot machine.

ssalinas: 👍 thanks, will update

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
```diff
@@ -358,6 +369,21 @@ public void onFailure(Throwable t) {
...
+  private void startCgroupWatcher(final SingularityExecutorTask task) {
+    SingularityTaskExecutorData taskExecutorData = (SingularityTaskExecutorData) task.getExecutorData();
+    if (taskExecutorData.getCpuHardLimit().isPresent()) {
+      cgroupCfsWatcherService.submit(() -> {
+        try {
+          SingularityExecutorCgroupCfsChecker cfsChecker = new SingularityExecutorCgroupCfsChecker(task, taskExecutorData.getCpuHardLimit().get(), configuration.getDefaultCfsPeriod());
+          cfsChecker.watch();
+          cgroupCheckers.put(task.getTaskId(), cfsChecker);
+        } catch (Throwable t) {
+          LOG.error("Could not start cgorup checker for task {}", task.getTaskId(), t);
```
baconmania: minor typo

baconmania:
🚢

baconmania:
🚢

PR #1776, opened 2018-04-04
Configurably omit offers from hosts that are overloaded

ssalinas:
If, according to the metrics collected from the slave metrics endpoint, a slave is overloaded, omit these offers from scoring and do not schedule tasks there. This feature is configurable via `omitOverloadedHosts` under the `mesos` config, and defaults to `false`/off

Overloaded is currently defined as a load5 > 1.0 to catch ongoing load, or a load1 > 1.5 to catch shorter spikes large enough to be disruptive to tasks on the host.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularitySlaveUsage.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularitySlaveUsage.java
@@ -2,10 +2,9 @@ package com.hubspot.singularity;
 
 import java.util.Map;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.google.common.base.Optional;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Optional;
 
 public class SingularitySlaveUsage {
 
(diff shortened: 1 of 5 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularitySlaveUsage.java
```diff
@@ -78,6 +79,12 @@ public SingularitySlaveUsage(@JsonProperty("cpusUsed") double cpusUsed,
     this.slaveDiskTotal = slaveDiskTotal;
   }
 
+  @JsonIgnore
+  public boolean isOverloaded() {
+    // Any host where load5 is > 1 is overloaded. Also consider a higher threshold for load1 to take into account spikes large enough to be disruptive
+    return  systemLoad5Min > 1.0 || systemLoad1Min > 1.5;
```
baconmania: Do we want to make these values tunable knobs as well?

ssalinas: yeah was going back and forth on that. Might as well for flexibility. I'll update 👍

baconmania:
🚢

baconmania:
🚢

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -353,6 +355,11 @@ private double score(SingularityOfferHolder offerHolder, Map<String, Integer> ta
       return 0;
     }
 
+    if (mesosConfiguration.isOmitOverloadedHosts() && maybeSlaveUsage.isPresent() && maybeSlaveUsage.get().isOverloaded()) {
+      LOG.debug("Slave {} is overloaded (), ignoring offer");
```
baconmania: Looks like we missed the interp'd parameters here.

PR #1777, opened 2018-04-04
Do not upload files outside task sandbox

pschoenfelder:


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskLogManager.java
```diff
@@ -69,6 +69,12 @@ private boolean writeS3MetadataFileForRotatedFiles(boolean finished) {
 
     for (SingularityS3UploaderFile additionalFile : taskDefinition.getExecutorData().getS3UploaderAdditionalFiles()) {
       Path directory = additionalFile.getDirectory().isPresent() ? taskDefinition.getTaskDirectoryPath().resolve(additionalFile.getDirectory().get()) : taskDefinition.getTaskDirectoryPath();
+
+      if (!directory.toAbsolutePath().startsWith(taskDefinition.getTaskDirectoryPath())) {
```
ssalinas: Do we want to `&&` this with `isCheckSubdirectories()`. The main thing we're trying to protect against is the uploader going crazy checking all subdirectories of `/` or something like that. Debating if we should allow a way for folks to use this for things outside of the sandbox, within reason. @baconmania thoughts?

pschoenfelder: As far as I can tell the `startsWith` method of checking this should not start recursively checking directories so I don't think we have to worry about going crazy checking all subdirs of `/`.

ssalinas: It's not the `startsWith`, that's just a string check, I'm worried about. It's the `isCheckSubdirectories()` value of the `additionalFile`. If it isn't checking subdirectories/not recursive, the most files the uploader will eventually scan is whatever is in the specified folder. If it _is_ recursive, that's the problem we are trying to prevent. (e.g. `isCheckSubdirectories()` -> `true` with a folder of `/` essentially causes it to check every folder on the host

That said I could go either way on limiting this. It's basically a question of whether or not we expect a task to write to and want to upload something outside it's sandbox, and if it does we certainly need to impose a limit that the folder isn't checked recursively

baconmania: I could see potentially only doing this limitation based on `isCheckSubdirectories() == true`, but my vote here right now would be to default to limiting everything to the sandbox unless we see an ask where people want to be able to upload things outside the sandbox.

ssalinas: 👍 fair enough, let's leave the check as-is for now then

ssalinas:
🚢

PR #1781, opened 2018-04-10
Log the full list of healthy task ids when killing a task

ssalinas:
This is to help figure out a possible race condition in pending tasks / bug with pending task creation


pschoenfelder:
🚢

PR #1784, opened 2018-04-19
Simplify offer scoring

ssalinas:
- Consolidate on a single strategy for offer scoring that will account for usage if it is available, using allocated resource values if not
- Stop treating long running, non-long running separately in offer scoring
- Consolidate configurable values
```
mesos:
  allocatedResourceWeight: 0.5
  inUseResourceWeight: 0.5
  cpuWeight: 0.4
  memWeight: 0.4
  diskWeight: 0.2
```
- rename of used/free to allocated/inUse

cc @baconmania @pschoenfelder


pschoenfelder:
🚢

PR #1785, opened 2018-04-19
Send email on failing replacement tasks.

baconmania:
/cc @ssalinas


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityNewTaskChecker.java
```diff
@@ -65,31 +68,35 @@
...
-                                   SingularityDeployHealthHelper deployHealthHelper, DisasterManager disasterManager) {
+                                   SingularityDeployHealthHelper deployHealthHelper, DisasterManager disasterManager, SingularityMailer mailer) {
     this.configuration = configuration;
     this.requestManager = requestManager;
     this.taskManager = taskManager;
     this.lbClient = lbClient;
     this.abort = abort;
 
     this.taskIdToCheck = Maps.newConcurrentMap();
+    this.requestIdToUnhealthyCount = Maps.newConcurrentMap();
```
ssalinas: documenting from in-person conversation. Right now this would reset when singularity deploys. It would be good to persist these counts somewhere

ssalinas:
🚢

ssalinas:
🚢

PR #1786, opened 2018-04-20
Add default expiration times for scale/pause modals

benheng:


benheng:
Closing for now since there's no consensus on what the default behavior should be.

PR #1788, opened 2018-04-24
Add ability to set attributes that mark slave for only preemptible tasks

ssalinas:
For example, set:
```yaml
preemptibleTasksOnlyMachineAttributes:
  preemptibleOnly: "true"
```

And only long-running (worker/service) or shorter non-long-running (`avgRunTime` < `preemptibleTaskMaxExpectedRuntimeMs`) will be placed on those slaves. As a result they are easier to shut down and replace if a portion of the mesos cluster is rotated out frequently


ssalinas:
tested and working as intended on staging 😃

PR #1792, opened 2018-04-26
Support HTTP2 healthchecks.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/SingularityMainModule.java
```diff
@@ -157,6 +158,7 @@ public void configure(Binder binder) {
     binder.bind(MetricRegistry.class).toProvider(DropwizardMetricRegistryProvider.class).in(Scopes.SINGLETON);
 
     binder.bind(AsyncHttpClient.class).to(SingularityAsyncHttpClient.class).in(Scopes.SINGLETON);
+    binder.bind(OkHttpClient.class).in(Scopes.SINGLETON);
```
ssalinas: Will want to also make sure to close this if it is Closeable. The `SingularityAsyncHttpClient` is bound for that reason, since it also implements dropwizard's `Managed` and will be closed on shutdown

ssalinas:
folding this into #1922 to avoid a ton of merge conflicts on staging

PR #1795, opened 2018-05-08
Add clear all buttons for dead slaves and inactive hosts

ssalinas:
These lists can be cumbersome to clean up, making a button for it


pschoenfelder:
🚢

PR #1798, opened 2018-05-15
Allow retrying run now requests

pschoenfelder:


SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -1491,10 +1491,10 @@ public void testOnDemandTasksPersist() {
   }
 
   @Test
-  public void testRunNowScheduledJobDoesNotRetry() {
+  public void testRunNowScheduledJobDoesNotRetryWithZeroLimit() {
     initScheduledRequest();
     SingularityRequest request = requestResource.getRequest(requestId, singularityUser).getRequest();
-    SingularityRequest newRequest = request.toBuilder().setNumRetriesOnFailure(Optional.of(2)).build();
+    SingularityRequest newRequest = request.toBuilder().setNumRetriesOnFailure(Optional.of(0)).build();
```
ssalinas: I'm wondering if we want to change this behavior or not. The previous behavior is that a UI-initiated scheduled job (i.e. running manually before the scheduled time) will not have retries, which seems like it would generally be the desired case. I'm thinking we should make a different case specific for ON_DEMAND types. Maybe the check you removed in the Scheduler could instead be specific to only SCHEDULED request types?

ssalinas:
looks gtg, let's give this a try in staging

pschoenfelder:
This worked in staging (including the UI changes). Used this to do a quick test - https://private.hubteamqa.com/singularity-test-iad/request/pschoenfelder-test-request

@ssalinas Cool to move to QA?

ssalinas:
🚢

PR #1799, opened 2018-05-15
Improve order of search results

Tamba32:
**Original Issue**
> I feel like my search should have caused `janus-web` to be the first entry in this search:
image

---
**Fix**
Before:
image

After:
image


andyhuang91:
@Tamba32 Can you update the screenshots to provide an example with the same search string and also comment on what issue this change addresses?

andyhuang91:
🚢

PR #1800, opened 2018-05-18
Ability to maintain the same path for custom nav bar links

ssalinas:
This will replace a placeholder with the current path. For task specific paths, it will instead route to the request page, since a specific task id is unlikely to exist in a different cluster


baconmania:
🚢

PR #1802, opened 2018-05-25
Auto expand Healthcheck section when healthcheck link is clicked

benheng:
This will only auto expand it the first time it's clicked because the `CollapsableSection` component handles it's own internal state. We're overriding it once, but since we're not controlling it from a parent component, I didn't want to disable the expand/collapse feature of the component.

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskDetail.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskDetail.jsx
@@ -130,7 +130,8 @@ class TaskDetail extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      previousUsage: null
+      previousUsage: null,
+      expandedHealthchecks: false,
     };
   }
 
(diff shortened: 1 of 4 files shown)


SingularityUI/app/components/taskDetail/TaskDetail.jsx
```diff
@@ -477,7 +478,12 @@ class TaskDetail extends Component {
     return (
       <div className="task-detail detail-view">
         {this.renderHeader(cleanup)}
-        <TaskAlerts task={this.props.task} deploy={this.props.deploy} pendingDeploys={this.props.pendingDeploys} />
+        <TaskAlerts
+          task={this.props.task}
+          deploy={this.props.deploy}
+          pendingDeploys={this.props.pendingDeploys}
+          onViewHealthchecks={() => this.setState({ expandedHealthchecks: true })}
```
andyhuang91: It feels a little weird to have a `onViewHealthchecks` prop. That prop is really specific. This implementation also doesn't seem to auto-expand the section when using the url directly or refreshing the page.

PR #1804, opened 2018-06-01
Faster refresh for request and task detail pages

ssalinas:
The request refresher should also refresh active tasks and scheduled tasks so things like watching a bounce don't require a refresh of the whole page. Similar for task detail, watching a launching task shouldn't require a page refresh. Shortened the refresh time on these as well

cc @kwm4385


kwm4385:
@ssalinas Was this related sentry resolved on QA?  https://sentry.io/hubspot-dev/singularityui-hs_qa/issues/575490579/

baconmania:
🚢

PR #1810, opened 2018-06-14
Task cleanup fixes for decommission and delete

ssalinas:
- Requests in deleting can get stuck if a task was already bouncing, fix that by allowing the deleting cleanup to overwrite a previous one
- If a slave is decommissioned at the same time as a deploy is happening, a new task from the new deploy can be left in a cleaning state. The pending request was always being made against the active deploy. It will now be made against the deploy for the particular task being cleaned


baconmania:
🚢

PR #1811, opened 2018-06-19
Create task cleanups for scale down at request update time

ssalinas:
We can run into a certain sate where consistent continuous instance numbers are not kept. If a cleanup is created between a request scaling down and the scheduler processing the pending request update, this can occur. This PR creates the task cleanups for the specific instances that need to be cleaned up at request update time to avoid that possible window.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
@@ -4,6 +4,7 @@ import static com.hubspot.singularity.WebExceptions.checkBadRequest;
 import static com.hubspot.singularity.WebExceptions.checkConflict;
 import static com.hubspot.singularity.WebExceptions.checkNotNullBadRequest;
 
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -149,6 +151,24 @@ private void submitRequest(SingularityRequest request, Optional<SingularityReque
       requestState = oldRequestWithState.get().getState();
     }
 
+    if (oldRequest.isPresent() && oldRequest.get().getInstancesSafe() < request.getInstancesSafe()) {
```
baconmania: I think we want to swap the operands on this comparison.

ssalinas: We do indeed, hadn’t actually gotten a chance to finish unit test and testing in staging in this one yet

pschoenfelder:
🚢

PR #1814, opened 2018-06-27
Updated logrotate frequency to allow for an override 

kdorosh:
To use this feature add additional file/file groups (via wildcard) and times to AdditionalLogrotateFile.

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
@@ -13,17 +13,17 @@ public class SingularityExecutorLogrotateAdditionalFile {
 
     @JsonCreator
     public static SingularityExecutorLogrotateAdditionalFile fromString(String value) {
-        return new SingularityExecutorLogrotateAdditionalFile(value, null, null, null);
+        return new SingularityExecutorLogrotateAdditionalFile(value, Optional.absent(), Optional.absent(), null);
     }
 
     @JsonCreator
     public SingularityExecutorLogrotateAdditionalFile(@JsonProperty("filename") String filename,
-        @JsonProperty("extension") String extension,
-        @JsonProperty("dateformat") String dateformat,
+                                                      @JsonProperty("extension") Optional<String> extension,
+                                                      @JsonProperty("dateformat") Optional<String> dateformat,
         @JsonProperty("logrotateFrequencyOverride") SingularityExecutorLogrotateFrequency logrotateFrequencyOverride) {
         this.filename = filename;
-        this.extension = Optional.fromNullable(extension);
-        this.dateformat = Optional.fromNullable(dateformat);
+        this.extension = extension;
+        this.dateformat = dateformat;
         this.logrotateFrequencyOverride = Optional.fromNullable(logrotateFrequencyOverride);
     }
 
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
@@ -3,24 +3,28 @@ package com.hubspot.singularity.executor.config;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Optional;
+import com.hubspot.singularity.executor.SingularityExecutorLogrotateFrequency;
 
 public class SingularityExecutorLogrotateAdditionalFile {
     private final String filename;
     private final Optional<String> extension;
     private final Optional<String> dateformat;
+    private final Optional<SingularityExecutorLogrotateFrequency> frequencyOverride;
 
     @JsonCreator
     public static SingularityExecutorLogrotateAdditionalFile fromString(String value) {
-        return new SingularityExecutorLogrotateAdditionalFile(value, Optional.<String>absent(), Optional.<String>absent());
+        return new SingularityExecutorLogrotateAdditionalFile(value, Optional.absent(), Optional.absent(), Optional.absent());
     }
 
     @JsonCreator
     public SingularityExecutorLogrotateAdditionalFile(@JsonProperty("filename") String filename,
         @JsonProperty("extension") Optional<String> extension,
-        @JsonProperty("dateformat") Optional<String> dateformat) {
+        @JsonProperty("dateformat") Optional<String> dateformat,
+        @JsonProperty("logrotateFrequencyOverride") Optional<SingularityExecutorLogrotateFrequency> frequencyOverride) {
         this.filename = filename;
         this.extension = extension;
         this.dateformat = dateformat;
+        this.frequencyOverride = frequencyOverride;
     }
 
     public String getFilename() {
@@ -33,5 +37,9 @@ public class SingularityExecutorLogrotateAdditionalFile {
 
     public Optional<String> getDateformat() {
         return dateformat;
+    }
+
+    public Optional<SingularityExecutorLogrotateFrequency> getFrequencyOverride() {
+        return frequencyOverride;
     }
 }
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
@@ -7,20 +7,20 @@ import com.hubspot.singularity.executor.SingularityExecutorLogrotateFrequency;
 
 public class SingularityExecutorLogrotateAdditionalFile {
     private final String filename;
-    private final Optional<String> extension;
-    private final Optional<String> dateformat;
-    private final Optional<SingularityExecutorLogrotateFrequency> frequencyOverride;
+    private final String extension;
+    private final String dateformat;
+    private final SingularityExecutorLogrotateFrequency frequencyOverride;
 
     @JsonCreator
     public static SingularityExecutorLogrotateAdditionalFile fromString(String value) {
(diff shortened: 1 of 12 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
```diff
@@ -3,35 +3,43 @@
...
         this.extension = extension;
         this.dateformat = dateformat;
+        this.frequencyOverride = frequencyOverride;
     }
 
     public String getFilename() {
         return filename;
     }
 
     public Optional<String> getExtension() {
```
baconmania: Each of the getters in a POJO like this should have the same return type as the private member variables they expose. In this case, I'd recommend having `extension`, `dateformat`, and `frequencyOverride` be `Optional<>`s (in both their member variable types and their getter types).

Our Jackson YAML deserialization should automatically take care of setting a missing field to an empty `Optional<>` in the resulting Java object.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/LogrotateTemplateContext.java
```diff
@@ -88,7 +88,8 @@ public String getCompressExt() {
         new LogrotateAdditionalFile(
           taskDefinition.getTaskDirectoryPath().resolve(additionalFile.getFilename()).toString(),
           additionalFile.getExtension().or(Strings.emptyToNull(Files.getFileExtension(additionalFile.getFilename()))),
-          dateformat
+          dateformat,
+            additionalFile.getLogrotateFrequencyOverride()
```
baconmania: We'll want to be consistent on the indent here before we merge up to master.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorLogrotateAdditionalFile.java
```diff
@@ -3,24 +3,28 @@
...
     public static SingularityExecutorLogrotateAdditionalFile fromString(String value) {
-        return new SingularityExecutorLogrotateAdditionalFile(value, Optional.<String>absent(), Optional.<String>absent());
+        return new SingularityExecutorLogrotateAdditionalFile(value, null, null, null);
     }
 
     @JsonCreator
     public SingularityExecutorLogrotateAdditionalFile(@JsonProperty("filename") String filename,
-        @JsonProperty("extension") Optional<String> extension,
-        @JsonProperty("dateformat") Optional<String> dateformat) {
+        @JsonProperty("extension") String extension,
```
ssalinas: Curious why you removed the optionals from the arguments. Both cases should work for that particular usage but there are a few things to keep in mind:
- Jackson can handle the optionals just fine when reading from json and a missing key will be an absent optional
- Not relevant for this case, but always be on the lookout for other uses of a constructor that would cause something like this to be a breaking change. For this particular case, these POJOs are just used to model something read from json. But for other cases, releasing a new version with a changed constructor would break other users

SingularityExecutor/src/main/resources/logrotate.conf.hbs
```diff
@@ -36,6 +36,7 @@ notifempty
 
 {{#if extrasFiles}}
 {{#each extrasFiles}}{{{filename}}} {
+  {{#if logrotateFrequencyOverride }}{{{logrotateFrequencyOverride}}}{{/if}}
```
ssalinas: We should double check if this will 'just work'. Not all versions of logrotate support `hourly` by default (Many are normally run on a daily cron, not hourly). If you look in `SingularityExecutorTaskLogManager.writeLogrotateFile` you'll notice that we additionally write an hourly cron for the specific logrotate conf file that needs it. We will likely need an additional check in that method to see if an additional file has an hourly rotate and possibly a separate cron file if so (the current cron uses the `-f` option to support hourly and force a logrotate run for those versions that do not support it)

ssalinas: Additional note, enough OSs likely support hourly logrotate now that we could do away with this extra check if we verify that it still works as intended

kdorosh: We have discussed this in depth, and more recent OSs do support it (and we have been supporting it at HubSpot for a while). We have [already updated] (https://git.hubteam.com/HubSpotProtected/puppet-deploy/pull/3869) the import statement to include https://github.com/rodjek/puppet-logrotate/blob/master/manifests/hourly.pp in our Mesos slave Puppet deploy. I have updated the code to put logrotate config files in /etc/logrotate.d/hourly. We need to monitor this in QA and make sure there are no averse performance penalties.

kdorosh: We have discussed this at length and determined that it's ok to put the logrotate config files at `/etc/logrotate.d/hourly`. The Mesos slave running our `SingularityExecutor` needs to support hourly cron jobs at this directory. We have already internally updated our puppet deploy to enforce this config on mesos slaves.

kdorosh: We have decided to support older logrotate versions and added code to Singularity to force a rotation hourly using the `-f` flag and a cron schedule. This implementation writes config files to two directories:
* `/etc/logrotate.d`
* `/etc/logrotate.d/hourly`

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/LogrotateTemplateContext.java
```diff
@@ -69,10 +71,29 @@ public String getCompressExt() {
...
+        .filter(p -> ! p.getLogrotateFrequencyOverride().equals(SingularityExecutorLogrotateFrequency.HOURLY)).collect(Collectors.toList());
+  }
+
+  /**
+   * Extra files for logrotate to rotate hourly. If these do not exist logrotate will continue without error.
+   * @return filenames to rotate.
+   */
+  public List<LogrotateAdditionalFile> getExtrasFilesHourly() {
+    return getAllExtraFiles().stream()
+        .filter(p -> p.getLogrotateFrequencyOverride().equals(SingularityExecutorLogrotateFrequency.HOURLY)).collect(Collectors.toList());
```
ssalinas: nit on formatting, put the collect call on a new line to make it more readable

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/LogrotateTemplateContext.java
```diff
@@ -84,17 +105,23 @@ public String getCompressExt() {
...
+                dateformat,
+                additionalFile.getLogrotateFrequencyOverride()
+            ));
+      }
     }
 
     return transformed;
   }
 
+
```
ssalinas: nit, extra new line

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/LogrotateTemplateContext.java
```diff
@@ -84,17 +105,23 @@ public String getCompressExt() {
         dateformat = configuration.getLogrotateExtrasDateformat().startsWith("-") ? configuration.getLogrotateExtrasDateformat().substring(1) : configuration.getLogrotateExtrasDateformat();
       }
 
-      transformed.add(
-        new LogrotateAdditionalFile(
-          taskDefinition.getTaskDirectoryPath().resolve(additionalFile.getFilename()).toString(),
-          additionalFile.getExtension().or(Strings.emptyToNull(Files.getFileExtension(additionalFile.getFilename()))),
-          dateformat
-      ));
+      if (additionalFile.getLogrotateFrequencyOverride().isPresent() &&
```
ssalinas: method name is getAll, did you meant to filter to hourly here still?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskLogManager.java
```diff
@@ -93,15 +94,42 @@ private boolean writeS3MetadataFileForRotatedFiles(boolean finished) {
...
+      String cronScheduleString = freq.getCronSchedule().isPresent() ? freq.getCronSchedule().get() : "Error in cron schedule"; // This should never evaluate to false
+
+      log.info("Writing logrotate cron entry with schedule '{}' to {}", cronScheduleString, getLogrotateCronPath());
+      templateManager.writeCronEntryForLogrotate(getLogrotateCronPath(), new LogrotateCronTemplateContext(configuration, taskDefinition, freq));
     }
   }
 
+  private Optional<SingularityExecutorLogrotateFrequency> getAdditionalHourlyFile() {
+    // Get the cron schedule for an additional file with an HOURLY schedule, if there is any
+    java.util.Optional<SingularityExecutorLogrotateAdditionalFile> additionalHourlyFile = configuration.getLogrotateAdditionalFiles()
```
ssalinas: nit, while util.Optional is the newer/nicer. It's probably easier for us to stick with guava optional until we do a consolidated swap across the whole repo.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/LogrotateTemplateContext.java
```diff
@@ -69,10 +71,30 @@ public String getCompressExt() {
...
+   * Extra files for logrotate to rotate hourly. If these do not exist logrotate will continue without error.
+   * @return filenames to rotate.
+   */
+  public List<LogrotateAdditionalFile> getExtrasFilesHourly() {
+    return getAllExtraFiles().stream()
+        .filter(p -> p.getLogrotateFrequencyOverride().equals(SingularityExecutorLogrotateFrequency.HOURLY.getLogrotateValue()))
+        .collect(Collectors.toList());
+  }
+
+  public boolean getIsGlobalLogrotateHourly() {
```
ssalinas: For booleans, the standard is to use the `is` prefix by itself (`isGlobalLogrotateHourly`). A number of libraries like reflections (and the the handlebars templater in this case) go by that standard.

ssalinas:
🚢

pschoenfelder:
🚢

baconmania:
🚢

PR #1817, opened 2018-07-11
Proxy download from Mesos slave over Singularity. 

kdorosh:
Solves problem of being on different network than Mesos slave. Directly proxying through a configurable load balancer can still be looked into.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
@@ -6,6 +6,7 @@ import static com.hubspot.singularity.WebExceptions.checkNotFound;
 import static com.hubspot.singularity.WebExceptions.notFound;
 
 import java.io.InputStream;
+import java.net.ConnectException;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
@@ -20,8 +21,6 @@ import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.client.Client;
-import javax.ws.rs.client.ClientBuilder;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -73,9 +72,11 @@ import com.hubspot.singularity.api.SingularityKillTaskRequest;
 import com.hubspot.singularity.api.SingularityTaskMetadataRequest;
 import com.hubspot.singularity.auth.SingularityAuthorizationHelper;
 import com.hubspot.singularity.config.ApiPaths;
+import com.hubspot.singularity.config.MesosConfiguration;
 import com.hubspot.singularity.config.SingularityTaskMetadataConfiguration;
 import com.hubspot.singularity.data.DisasterManager;
 import com.hubspot.singularity.data.RequestManager;
+import com.hubspot.singularity.data.SandboxManager.SlaveNotFoundException;
 import com.hubspot.singularity.data.SingularityValidator;
 import com.hubspot.singularity.data.SlaveManager;
 import com.hubspot.singularity.data.TaskManager;
@@ -99,6 +100,8 @@ import io.swagger.v3.oas.annotations.tags.Tags;
 public class TaskResource extends AbstractLeaderAwareResource {
   private static final Logger LOG = LoggerFactory.getLogger(TaskResource.class);
 
+  private final AsyncHttpClient httpClient;
+  private final MesosConfiguration configuration;
   private final TaskManager taskManager;
   private final RequestManager requestManager;
   private final SlaveManager slaveManager;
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -609,4 +613,30 @@ private SingularityTaskShellCommandRequest startShellCommand(SingularityTaskId t
     SingularityTaskId taskIdObj = getTaskIdFromStr(taskId);
     return taskManager.getTaskShellCommandUpdates(new SingularityTaskShellCommandRequestId(taskIdObj, commandName, commandTimestamp));
   }
+
+  @GET
+  @Path("/download/{httpPrefix}/{slaveHostname}/port/{port}/path/{path}")
+  @Produces(MediaType.APPLICATION_OCTET_STREAM)
+  @Operation(summary = "Proxy a file download from a Mesos Slave through Singularity")
+  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
```
baconmania: Since this is a `GET`, it shouldn't consume anything.

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -609,4 +613,30 @@ private SingularityTaskShellCommandRequest startShellCommand(SingularityTaskId t
...
+  public Response downloadFileOverProxy(
+      @Parameter(required = true, description = "Https (if available) or http") @PathParam("httpPrefix") String httpPrefix,
+      @Parameter(required = true, description = "Mesos slave hostname") @PathParam("slaveHostname") String slaveHostname,
+      @Parameter(required = true, description = "Http port for Mesos slave") @PathParam("port") String httpPort,
+      @Parameter(required = true, description = "Full file path to file on Mesos slave to be downloaded") @PathParam("path") String fileFullPath
+  ) {
+
+    Client client = ClientBuilder.newClient();
```
baconmania: I believe we want to inject & use the `AsyncHttpClient` here instead of using a JAX-RS client, similarly to https://github.com/HubSpot/Singularity/blob/c2d88711176b436a0e1683a83953222b7be0443d/SingularityService/src/main/java/com/hubspot/singularity/data/SandboxManager.java#L58-L63.

SingularityUI/app/components/taskDetail/TaskDetail.jsx
```diff
@@ -178,9 +178,9 @@ class TaskDetail extends Component {
           file.uiPath = file.name;
         }
 
-        file.fullPath = `${files.fullPathToRoot}/${files.currentDirectory}/${file.name}`;
-        file.downloadLink = `${httpPrefix}://${files.slaveHostname}:${httpPort}/files/download.json?path=${file.fullPath}`;
+        file.fullPath = encodeURIComponent(`${files.fullPathToRoot}/${files.currentDirectory}/${file.name}`);
 
+        file.downloadLink = `${config.apiRoot}/tasks/download/${httpPrefix}/${files.slaveHostname}/port/${httpPort}/path/${file.fullPath}`;
```
baconmania: The value of `httpPrefix` is [based on the Singularity configuration](https://github.com/HubSpot/Singularity/blob/c2d88711176b436a0e1683a83953222b7be0443d/SingularityUI/app/components/taskDetail/TaskDetail.jsx#L168-L173), so we shouldn't need to send this as a parameter to the backend (since the backend also has access to that configuration and knows whether the slave communicates over HTTP or HTTPS and which port it's listening on).

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -609,4 +613,30 @@ private SingularityTaskShellCommandRequest startShellCommand(SingularityTaskId t
     SingularityTaskId taskIdObj = getTaskIdFromStr(taskId);
     return taskManager.getTaskShellCommandUpdates(new SingularityTaskShellCommandRequestId(taskIdObj, commandName, commandTimestamp));
   }
+
+  @GET
+  @Path("/download/{httpPrefix}/{slaveHostname}/port/{port}/path/{path}")
```
baconmania: Sort of a nit, but a better API design here could be something like `GET /download?slaveHostname=x&path=y` (since the actual resource being retrieved is a `download`). The rest of our API doesn't stick to these guidelines religiously, though, so probably not a huge deal. I'll let @ssalinas weigh in as well.

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
kdorosh: -- whoops -- left this there from when I had it as a `POST`

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
kdorosh: Ah neat, will do. We already have one being injected too which is awesome since creating these clients are expensive operations.

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
kdorosh: Wouldn't it be better to have them both as `PathParam`s instead of `QueryParam`s since both the slave hostname and the path are both required fields? I did change the URL a bit to better match entity class hierarchy standards. See top answer and comments on [this](https://stackoverflow.com/questions/11552248/when-to-use-queryparam-vs-pathparam) StackOverflow post.

kdorosh:
This was merged into `hs_staging` and `hs_qa` yesterday, 7/12/18.

ssalinas:
🚢

PR #1820, opened 2018-07-14
Update branch version to match pom version

TheGreatGooo:
Update the version so that blazar will build branches with the correct version that matches the pom version


ssalinas:
Replaced by #1821

PR #1822, opened 2018-07-17
WIP: Prevent new host overloading

pschoenfelder:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -240,6 +238,59 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
     return offerHolders.values();
   }
 
+  private Void buildScoringFuture(
```
ssalinas: nit on naming, this method isn't actually building the future, it's a synchronous method for doing the actual scoring. I'd either rename this to calculateScore (or something like that), or move the supplyAsync inside this method and have it return the actual `CompletableFuture<Void>`

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -240,6 +238,59 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
...
+                    mesosConfiguration.getLoad5OverloadedThreshold(),
+                    mesosConfiguration.getLoad1OverloadedThreshold(),
+                    usage.getTimestamp()
+                ));
+              } else {
+                throw new RuntimeException(throwable);
+              }
+          });
+      }
+      return null;
```
ssalinas: So, I want to mention the choice here, could go either way on this. Currently this looks like the newly updated slave metrics will not be taken into account because we are returning here. Wouldn't we want to continue on to the scoring since we've gathered new metrics and put them in the map that is fed to `calculateScore`?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveUsageWithCalculatedScores.java
```diff
@@ -121,6 +129,10 @@ SingularitySlaveUsage getSlaveUsage() {
     return diskInUseScore;
   }
 
+  long getTimestamp() {
```
ssalinas: you should be able to just do getSlaveUsage().getTimestamp() instead of having to store it in two places

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -126,15 +126,35 @@ public void runActionOnPoll() {
...
+            utilizationPerRequestId, totalMemBytesUsed.get(), totalMemBytesAvailable.get(),
+            totalCpuUsed.get(), totalCpuAvailable.get(), totalDiskBytesUsed.get(), totalDiskBytesAvailable.get(), now));
     utilizationPerRequestId.values().forEach(usageManager::saveRequestUtilization);
 
     if (configuration.isShuffleTasksForOverloadedSlaves()) {
       shuffleTasksOnOverloadedHosts(overLoadedHosts);
     }
   }
 
+  public CompletableFuture<SingularitySlaveUsage> getSlaveUsage(SingularitySlave slave) {
```
ssalinas: The only thing I find as a code smell here is that now the offer scoring flow will rely on the usage semaphore and executor having enough permits/threads. Since within the offer scoring we are already in a block that is executed async, it may be worth calling `collectSlaveUsage` directly when used from the offer context, to avoid an additional layer of async work

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -314,11 +334,13 @@ private void collectSlaveUage(SingularitySlave slave,
 
       LOG.debug("Saving slave {} usage {}", slave.getHost(), slaveUsage);
       usageManager.saveSpecificSlaveUsageAndSetCurrent(slave.getId(), slaveUsage);
+      return slaveUsage;
     } catch (Throwable t) {
       String message = String.format("Could not get slave usage for host %s", slave.getHost());
       LOG.error(message, t);
       exceptionNotifier.notify(message, t);
     }
+    return null; // TODO: is this really okay?
```
ssalinas: I don't think this method is called anywhere else that expects a return value. Could always wrap in an optional to make it more explicit that the result might not be there.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
pschoenfelder: Yes, I'll go with the latter option. For some reason I was thinking only the slave usage is necessary but probably safer to update the score too.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -240,6 +240,71 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
...
+                currentSlaveUsagesBySlaveId.put(slaveId, new SingularitySlaveUsageWithCalculatedScores(
+                    usage.get(),
+                    mesosConfiguration.getScoreUsingSystemLoad(),
+                    getMaxProbableUsageForSlave(activeTaskIds, requestUtilizations, offerHolders.get(slaveId).getSanitizedHost()),
+                    mesosConfiguration.getLoad5OverloadedThreshold(),
+                    mesosConfiguration.getLoad1OverloadedThreshold(),
+                    usage.get().getTimestamp()
+                ));
+              } else {
+                throw new RuntimeException(throwable);
```
ssalinas: Where is the handling for this runtime exception? We currently aren't calling a get or join for the future created here which causes two issues for us:
- calculateScore below may end up being called before the metrics recollection has a chance to run
- If a RuntimeException is thrown here it is lost to us since there it will not propagate out of the future and no catch block currently logs it

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -240,6 +240,71 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
...
+      Optional<SingularitySlave> maybeSlave = slaveManager.getSlave(slaveId);
+      if (maybeSlave.isPresent()) {
+        CompletableFuture.supplyAsync(() ->
+            usagePoller.collectSlaveUsage(
+              maybeSlave.get(),
+              System.currentTimeMillis(),
+              new ConcurrentHashMap<>(),
+              usageManager.getRequestUtilizations(),
+              new ConcurrentHashMap<>(),
+              new AtomicLong(),
```
ssalinas: given these arguments will be the same each time, does it make sense to create an overloaded method to handle those bits in UsagePoller instead?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -180,7 +191,8 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
                 mesosConfiguration.getScoreUsingSystemLoad(),
                 getMaxProbableUsageForSlave(activeTaskIds, requestUtilizations, offerHolders.get(usageWithId.getSlaveId()).getSanitizedHost()),
                 mesosConfiguration.getLoad5OverloadedThreshold(),
-                mesosConfiguration.getLoad1OverloadedThreshold()
+                mesosConfiguration.getLoad1OverloadedThreshold(),
+                usageWithId.getTimestamp()
```
ssalinas: Second thoughts about the setup. Would it make sense to instead collect additional usages in this block instead? I'm realizing that the loop below would be called for each pending task. If we hit a case where collecting a particular slave usage is throwing exceptions or timing out, we will continue to recheck it for each pending task. Whereas, if we check in this block instead, we can just omit that up front and leave the block below as it was previously.

If we move the usage collection here, we'll likely want to convert this from parallelStream to a list of CompletableFutures like below to have better control over the concurrency

ssalinas:
🚢

pschoenfelder:
🚢

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsageHelper.java
```diff
@@ -45,4 +103,472 @@ public SingularityUsageHelper(SlaveManager slaveManager) {
...
+        SingularityTaskId task;
+        try {
+          task = SingularityTaskId.valueOf(taskId);
+        } catch (InvalidSingularityTaskIdException e) {
+          LOG.error("Couldn't get SingularityTaskId for {}", taskUsage);
+          continue;
+        }
+
+        SingularityTaskUsage latestUsage = getUsage(taskUsage);
+        memoryBytesUsedOnSlave += latestUsage.getMemoryTotalBytes();
```
ssalinas: Will these end up being any different than `systemMemTotalBytes - systemMemFreeBytes` ? I'm wondering if there isn't actually a need to total everything up for these cases. `getUsage` is still a zk call and it'd be nice to eliminate if we can. The totaled up by task values and the slave-reported totals seem fairly similar. I'm not certain why we have both in the POJO TBH, would have to look through commit history. (Or maybe @darcatron knows since he wrote the original versions of usage collection?)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsageHelper.java
```diff
@@ -45,4 +103,446 @@ public SingularityUsageHelper(SlaveManager slaveManager) {
...
+      } else {
+        memoryMbTotal = Optional.of(slave.getResources().get().getMemoryMegaBytes().get().longValue());
+        cpusTotal = Optional.of(slave.getResources().get().getNumCpus().get().doubleValue());
+        diskMbTotal = Optional.of(slave.getResources().get().getDiskSpace().get());
+      }
+
+      SingularitySlaveUsage slaveUsage = new SingularitySlaveUsage(
+          cpuReservedOnSlave, cpuReservedOnSlave, cpusTotal,
+          memoryMbReservedOnSlave, memoryMbReservedOnSlave, memoryMbTotal,
+          diskMbReservedOnSlave, diskMbReservedOnSlave, diskMbTotal,
```
pschoenfelder: Seems pretty funky to send in these same variables, but that's what reducing everything to use the metric snapshot resulted in (see the plain `collectSlaveUsage` below to compare).

ssalinas: I guess the 'clean' way to do it could be to have a separate POJO with a more minimal set of fields, which gets used in the scheduling. The fuller class, which would extend that, has all of the fields is used in the poller + api

e.g. SingularitySlaveUsage extends SingularitySimpleSlaveUsage {}

baconmania:
🚢

PR #1825, opened 2018-07-30
Deprioritize STARTUP-type pending requests.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -239,6 +239,16 @@ public SingularityCreateResult finish(SingularityRequest request, long timestamp
   public SingularityCreateResult addToPendingQueue(SingularityPendingRequest pendingRequest) {
     SingularityCreateResult result = create(getPendingPath(pendingRequest), pendingRequest, pendingRequestTranscoder);
 
+    if (result == SingularityCreateResult.EXISTED) {
+      SingularityPendingRequest existingPendingRequest = getPendingRequest(pendingRequest.getRequestId(), pendingRequest.getDeployId()).get();
```
ssalinas: Since the scheduler poller could run between the create and this get, we can't assume that this optional will be present. Should probably add the isPresent check

ssalinas:
🚢

PR #1829, opened 2018-08-08
Prepare a StreamingOutput response when serving file downloads.

baconmania:
This prepares a 4k buffer and pipes the HTTP response InputStream to a jax-rs compatible StreamingOutput object.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
@@ -697,13 +697,13 @@ public class TaskResource extends AbstractLeaderAwareResource {
 
     @Override
     public STATE onHeadersReceived(HttpResponseHeaders headers) {
-      LOG.trace("Proxying download of {} from Mesos: Headers={}", requestBuilder.build().getHeaders(), headers.getHeaders());
+      LOG.trace("Proxying download of {} from Mesos: Headers={}", requestBuilder.build().getUrl(), headers.getHeaders());
       return STATE.CONTINUE;
     }
 
     @Override
     public Void onCompleted() {
-      LOG.info("Proxying download of {} from Mesos: Completed.");
+      LOG.info("Proxying download of {} from Mesos: Completed.", requestBuilder.build().getUrl());
       return null;
     }
 


SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -634,19 +643,22 @@ public Response downloadFileOverProxy(
             httpPrefix, slaveHostname, httpPort);
 
     try {
-      final InputStream responseStream = httpClient.prepareGet(url)
-          .addQueryParameter("path", fileFullPath)
-          .execute()
-          .get()
-          .getResponseBodyAsStream();
+      PerRequestConfig unlimitedTimeout = new PerRequestConfig();
+      unlimitedTimeout.setRequestTimeoutInMs(-1);
```
baconmania: The default 60 second timeout on our AsyncHttpClient is insufficient for large file downloads.

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -656,4 +668,56 @@ public Response downloadFileOverProxy(
     }
 
   }
+
+  private static class NingOutputToJaxRsStreamingOutputWrapper implements AsyncHandler<Void>, StreamingOutput {
```
baconmania: Fully open to suggestions on a better name for this, heh.

SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -656,4 +668,56 @@ public Response downloadFileOverProxy(
...
+      return null;
+    }
+
+    // StreamingOutput implementation: just make the OutputStream available to the AsyncHandler<> implementation.
+    @Override
+    public void write(OutputStream output) throws WebApplicationException, IOException {
+      if (wrappedOutputStream == null) {
+        wrappedOutputStream = output;
+        try {
+          requestBuilder.execute(this).get();
```
baconmania: We need to do it this way because the `OutputStream` doesn't become available to us until Jersey actually calls `write()` on this implementation. The alternative would be to begin the proxied request _before_ waiting for Jersey, in which case we'd be back to buffering a potentially large response in memory.

ssalinas:
🚢

PR #1831, opened 2018-08-08
Only run a single persister at once

ssalinas:
Don't overwhelm zk with too many writes at once with all of these running in parallel.

@baconmania


baconmania:
🚢

PR #1832, opened 2018-08-09
Nix the old dashboard and default to requests page

ssalinas:
Requests page is more often used and loads faster. Also adds an autofocus on the end of the filter input


SingularityUI/app/components/common/Navigation.jsx
```diff
@@ -84,9 +84,6 @@ const Navigation = (props) => {
         </div>
         <div className="collapse navbar-collapse" id="navbar-collapse">
           <ul className="nav navbar-nav">
-            <li className={classnames({active: isActive('/', fragment)})}>
```
andyhuang91: If `Requests` is the default tab now, we should reorder the nav items so that `Status` isn't first.

ssalinas: Yeah we can do that as well

baconmania:
🚢

PR #1833, opened 2018-08-16
Move directory logging to trace level

ssalinas:
Easy to fill up disk for lots of directories. Move this to trace level


baconmania:
🚢

PR #1834, opened 2018-08-16
Option to skip scheduling on slaves with missing usage data

ssalinas:
This was causing us to not hit some of the overloaded new host logic, since it relies on the previous usage data. Putting this behind a flag, defaulted to false


baconmania:
🚢

PR #1836, opened 2018-08-16
Remove second declaration of defaultProps in UITable in SingaularityUI.

Tamba32:
Resolves issue that caused the requests table not to show up at `singularity/ui/requests/`.


ssalinas:
🚢

Issue #1803, opened 2018-05-28
Unnecessary LOG.error message when running singularity on a cluster that contains another framework


gtinjr:
When installing SingularityService 0.19.2 on a Mesos cluster with Marathon installed,  Singularity logs error messages because it can not find singularityTaskId on tasks that got deployed by marathon.

Could you please remove the LOG.error("Couldn't get SingularityTaskId for {}", taskUsage); from SingularityUsagePoller.java around line 199?

Thx

PR #1840, opened 2018-08-27
Don't try to parse task ids from other frameworks

ssalinas:
/fixes https://github.com/HubSpot/Singularity/issues/1803


pschoenfelder:
🚢

PR #1841, opened 2018-08-28
Fix required groups auth check on requests endpoints

ssalinas:
Also adds a more descriptive empty table message to the requests page

cc @jonathanwgoodwin


pschoenfelder:
🚢

PR #1843, opened 2018-08-29
Add email unsubscriptions

pschoenfelder:


ssalinas:
storage/api portions look fine here. Only potential thing I see is that, since Singularity sends _a lot_ of emails, we will make quite a few getChildren calls fetching that list. Can we maybe cache it for some amount of time? Even caching for 5 minutes or so would save a ton of calls

SingularityService/src/main/java/com/hubspot/singularity/data/NotificationsManager.java
```diff
@@ -0,0 +1,55 @@
...
+              @Override public List<String> load(String key) throws Exception {
+                return getChildren(key);
+              }
+            }
+        );
+  }
+
+  public void addToBlacklist(String email) {
+    create(getEmailPath(email));
+    cache.invalidate(BLACKLIST_ROOT);
```
ssalinas: last note here... depending what singularity service instance is hit this may or may not do any good. If the leader is hit with the api call, then it will be effective immediately. If a non-leader instance is hit, the leader will still be waiting until its cache expires. If we want this to be 100% consistent on the leader, we'd want to add the `maybeProxyToLeader` pattern we use in other resources to the NotificationsResource.

pschoenfelder: Good point. But how does hitting the leader ensure synchronization across all instances if each has it's own cache? Wouldn't we need to invalidate the cache on every single instance?

ssalinas: It doesn't ensure synchronization everywhere, but it ensures that the one actually doing the email sending is up to date

ssalinas:
🚢

PR #1846, opened 2018-09-06
Logic fix for rebalance racks cleanup

ssalinas:
We previously had some issues where rebalance racks would be triggered when scaling down to {num racks}. This fixes the logic to not create unnecessary cleanups


pschoenfelder:
🚢

PR #1848, opened 2018-09-10
Run-now attribute overrides.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -245,11 +245,16 @@ private boolean isSlaveAttributesMatch(SingularityOfferHolder offer, Singularity
...
 
-      if ((taskRequest.getRequest().getRequiredSlaveAttributes().isPresent() && !taskRequest.getRequest().getRequiredSlaveAttributes().get().isEmpty())
+      if (!taskRequest.getPendingTask().getRequiredSlaveAttributeOverrides().isEmpty()
+          || !taskRequest.getPendingTask().getAllowedSlaveAttributeOverrides().isEmpty()
+          || (taskRequest.getRequest().getRequiredSlaveAttributes().isPresent() && !taskRequest.getRequest().getRequiredSlaveAttributes().get().isEmpty())
           || (taskRequest.getRequest().getAllowedSlaveAttributes().isPresent() && !taskRequest.getRequest().getAllowedSlaveAttributes().get().isEmpty())) {
         Map<String, String> mergedAttributes = new HashMap<>();
         mergedAttributes.putAll(taskRequest.getRequest().getRequiredSlaveAttributes().or(new HashMap<>()));
         mergedAttributes.putAll(taskRequest.getRequest().getAllowedSlaveAttributes().or(new HashMap<>()));
+        mergedAttributes.putAll(taskRequest.getPendingTask().getAllowedSlaveAttributeOverrides());
```
ssalinas: Opinion question, do we want these to be additive or a replacement? override makes me think it replaces the ones at the request level, not adds to them

baconmania: Yeah, it's a good question. I was initially going to go with "replace instead of merge", but then I felt like the least surprising thing would be to follow what we do for env vars (which is merge the run-now overrides into the `SingularityDeploy`-level env vars).

I could also see the argument that the env var example isn't really comparable to the attributes case, though.

ssalinas: I think I'd lean toward overwrite for this case. If we merge there isn't a way to 'unset' a requirement for an attribute

ssalinas:
🚢

ssalinas:
🚢

PR #1849, opened 2018-09-12
Slave attribute minimums

pschoenfelder:
Allows a minimum number of tasks on slaves with a certain attribute


pschoenfelder:
@baconmania Thanks for the help understanding the scheduler better - I found and fixed the issue we were discussing earlier. Definitely still need to add in the rebalancing part for when hosts are decommed, but this is a start.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -115,6 +117,8 @@ SlaveMatchState doesOfferMatch(SingularityOfferHolder offerHolder, SingularityTa
 
     if (!isSlaveAttributesMatch(offerHolder, taskRequest, isPreemptibleTask)) {
       return SlaveMatchState.SLAVE_ATTRIBUTES_DO_NOT_MATCH;
+    } else if (!areSlaveAttributeMinimumsFeasible(offerHolder, taskRequest, activeTaskIdsForRequest)) {
+      return SlaveMatchState.SLAVE_ATTRIBUTES_DO_NOT_MATCH; // TODO: other enum?
```
ssalinas: SLAVE_ATTRIBUTES_DO_NOT_MATCH enum is probably fine here, since it is still another case of this that we are checking for

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackManager.java
```diff
@@ -277,6 +281,47 @@ private boolean isSlaveAttributesMatch(SingularityOfferHolder offer, Singularity
...
+          return false;
+        }
+      }
+    }
+    return true;
+  }
+
+  private long getNumInstancesWithAttribute(List<SingularityTaskId> taskIds, String attrKey, String attrValue) {
+    return taskIds.stream()
+        .map(id -> slaveManager.getSlave(taskManager.getTask(id).get().getMesosTask().getSlaveId().getValue()).get().getAttributes().get(attrKey))
```
ssalinas: I think we already fetch a full list of slaves somewhere before this. Each of these would be a zk call currently. So, we should either:
- Add slaveManager stuff to leader cache so this is all in memory
or 
- Pass through a list of slave data we have already read from zk for use here

pschoenfelder: I'll add to leader cache, think that's the safer bet since I don't see a prefetch anywhere close before.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -488,29 +489,95 @@ private int scheduleTasks(SingularityRequest request, RequestState state,
...
+        rebalanceRacks(request, state, deployStatistics, pendingRequest, maybePendingDeploy, now, remainingActiveTasks);
+      }
+      if (request.getSlaveAttributeMinimums().isPresent()) {
+        rebalanceAttributeDistribution(request, state, deployStatistics, pendingRequest, maybePendingDeploy, now, remainingActiveTasks);
+      }
+    }
+    return numMissingInstances;
+  }
+
+  private void rebalanceAttributeDistribution(
```
pschoenfelder: @ssalinas @baconmania this is a grossly inefficient implementation. Any ideas or suggestions greatly appreciated.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -488,29 +489,95 @@ private int scheduleTasks(SingularityRequest request, RequestState state,
...
+      RequestState state,
+      SingularityDeployStatistics deployStatistics,
+      SingularityPendingRequest pendingRequest,
+      Optional<SingularityPendingDeploy> maybePendingDeploy,
+      long now,
+      List<SingularityTaskId> remainingActiveTasks) {
+    Map<String, Map<String, Integer>> instanceAttrCounts = new HashMap<>();
+
+    for (SingularityTaskId taskId: remainingActiveTasks) {
+      SingularitySlave slave = leaderCache.getSlave(taskManager.getTask(taskId).get().getMesosTask().getSlaveId().getValue()).get();
```
ssalinas: Totally missed that slave manager already has a `if (leaderCache.active()) {return leaderCache.getSlave(slaveId);}`, my bad on that one. Guess we can use slaveManager after all in these spots

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
ssalinas: My first thought on this is that we'll want to maintain one instance of `extraTasksToClean`, not recreate it each loop. For a single attribute minimum specified, this works just fine. It breaks down more for multiple since we'd be creating multiple cleanups possibly for the same task. Also, you are looping through the tasks twice here which you can likely just do it once.

Current:
 - loop through tasks to find instance attr counts
- loop through minimums and loop through tasks again to determine which to clean

Proposed:
- Loop through tasks to build lists of task ids with/without the attributes you care about (e.g. something like Map<Attr, Map<Boolean, List<TaskId>>> as opposed to just counting.
- Loop through minimums to find what might need a count of cleanups and pop those out of the lists if they aren't already being clean up for a different attribute minimum

There is probably a way we can create less overall cleanups by figuring out which cleanups satisfy the most requirements, but I think that may be over optimizing considering we are designing this currently for a rather specific use case of a single minimum attribute

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -185,20 +188,13 @@ private void submitRequest(SingularityRequest request, Optional<SingularityReque
...
-            LOG.info("{} - {} - {} - {}", countPerRack, perRack, extraCleanedTasks.size(), taskId);
-            if (countPerRack.count(taskId.getSanitizedRackId()) > perRack && extraCleanedTasks.size() < numActiveRacks / 2) {
-              extraCleanedTasks.add(taskId);
-              LOG.info("Cleaning up task {} to evenly distribute tasks among racks", taskId);
-              taskManager.createTaskCleanup(new SingularityTaskCleanup(user.getEmail(), TaskCleanupType.REBALANCE_RACKS, System.currentTimeMillis(), taskId, Optional.<String>absent(), Optional.<String>absent(), Optional.absent()));
-            }
-          }
+          rebalancingHelper.rebalanceRacks(request, remainingActiveTasks, user.getEmail());
+          // Why not remove cleaned tasks from remainingActiveTasks here like in SingualrityScheduler?
+          // Why does scheduler do straight up user, and this does user.getemail?
```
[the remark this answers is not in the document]
ssalinas: I believe the username is often the email anyways. It can be the username instead there though, that just powers what shows up in the UI (e.g. user that triggered the cleanup).

`remainingActiveTasks` here already excludes those that were cleaned up. In the loop above this we are only adding to that list if we don't create a cleanup

ssalinas:
@pschoenfelder I know there are some efficiency improvements to be made, but I think this should be good to try in staging if you want. The unit tests have proved out it functions and it's probably worth a 'real world' test to see how much time it's worth sinking into those improvements on that loop mentioned above

PR #1850, opened 2018-09-13
Make requestUtilization map thread safe.

baconmania:
The `requestUtilizations` map is iterated over when it gets loaded into a new `HashMap` here:
https://github.com/HubSpot/Singularity/blob/2c77825acf1cd5929029cb678ecb71172ac8e566/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityLeaderCache.java#L460

Because `requestUtilizations` wasn't previously threadsafe, we'd get occasional `ConcurrentModificationException`s here.


ssalinas:
🚢

PR #1851, opened 2018-09-13
lsof alternative for open files check

ssalinas:
We have found that with high process counts this lsof call can be very inefficient and cause spurts of high load. This adds an option of using fuser which, while not a perfect replacement, is a faster alternative for this specific use case


baconmania:
🚢

PR #1852, opened 2018-09-19
Use a lock & a timeout when checking whether an upload target file is open.

baconmania:


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityUploader.java
```diff
@@ -68,6 +72,8 @@
     this.logIdentifier = String.format("[%s]", metadataPath.getFileName());
     this.configuration = configuration;
     this.exceptionNotifier = exceptionNotifier;
+
+    this.checkFileOpenLock = new ReentrantLock();
```
ssalinas: Just checking, is this class always a singleton? Want to make sure we aren't just creating multiple locks. I had thought the singleton was the uploader driver, not the uploader

baconmania: Good catch. Fixed.

ssalinas:
🚢

ssalinas:
🚢

PR #1853, opened 2018-09-25
Failover when we miss Mesos Master heartbeats.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerImpl.java
```diff
@@ -329,7 +339,8 @@ public void error(String message) {
 
   @Override
   public void heartbeat(Event event) {
-    LOG.debug("Heartbeat from mesos");
+    long now = System.currentTimeMillis();
+    LOG.debug("Heartbeat from mesos. Delta since last heartbeat is {}ms", now - lastHeartbeatTime.getAndSet(now));
```
ssalinas: seems like a weird thing to have the getAndSet inside the debug line. I know it gets evaluated the same, but I had to read this a few times to convince myself it would always be hit

baconmania: Aye, it'll get eval'd before (and regardless of whether) the logger actually tries to write the line. Agreed that it's unnecessarily dense though, I can split this out for clarity.

PR #1854, opened 2018-09-26
Open sandbox files from SingularityUI

### LINKED ISSUE #1357: Singularity file browser / downloader should provide correct Content-Type headers
Right now everything, including text files like `stdout`, seems to be delivered as `application/octet-stream`.  It would be nice if Singularity made basic efforts to guess a mime type, maybe via `Files.probeContentType`, and deliver that.  Probably will fix #390 as well.

ssalinas:
Adds an open endpoint that guesses at content type, allowing certain extensions to be opened from the UI

/fixes #1357

### DIFF
--- a/SingularityUI/app/components/taskDetail/TaskDetail.jsx
+++ b/SingularityUI/app/components/taskDetail/TaskDetail.jsx
@@ -180,6 +180,10 @@ class TaskDetail extends Component {
 
         file.fullPath = `${files.fullPathToRoot}/${files.currentDirectory}/${file.name}`;
         file.downloadLink = `${config.apiRoot}/tasks/download?slaveHostname=${files.slaveHostname}&path=${file.fullPath}`;
+        const extensionMatcher = /(?:\.([^.]+))?$/;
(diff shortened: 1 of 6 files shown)


SingularityUI/app/components/taskDetail/TaskDetail.jsx
```diff
@@ -180,6 +180,10 @@ class TaskDetail extends Component {
 
         file.fullPath = `${files.fullPathToRoot}/${files.currentDirectory}/${file.name}`;
         file.downloadLink = `${config.apiRoot}/tasks/download?slaveHostname=${files.slaveHostname}&path=${file.fullPath}`;
+        const extensionMatcher = /(?:\.([^.]+))?$/;
```
pschoenfelder: What about using something like `FilenameUtils.getExtension(fileName);` instead of a regex for clarity?

ssalinas: java(script) :lol:

pschoenfelder: Never mind, javascript :facepalm:

pschoenfelder:
🚢

PR #1860, opened 2018-09-28
Make SingularityUI snappier

ssalinas:
and update node/npm...
and remove the vulnerable dependencies...
and...
and...

WIP

cc @kwm4385


ssalinas:
Summary so far:
- Address most of the vulnerable dep changes. The bootstrap one requires an upgrade to bootstrap 4
- Uses the request ids list (from global search) for the requests table until the regular request json finishes loading, will allow the user to start searching/filtering earlier.
- Global search just uses request ids rather than requesting the full json (which was previously done on every page because of this)
- remove a duplicate ravenjs dep, we are including sentry in other ways
- remove select2 dep, unused that I could find
- update node sass so we can run on newer node/npm

before:
screen shot 2018-10-01 at 11 16 36 am
after:
screen shot 2018-10-01 at 11 16 15 am

kwm4385:
Code looks fine besides a few style nits, but I'm not worried about that since we don't have up to date linting on here (A project for another time perhaps).

Are we still sticking with the requests page as the index?

ssalinas:
Yeah, seemed to be the consensus that requests was the most useful page to land on. The dashboard attempted to be more personalized, but most folks just went right to requests to filter for the thing they were looking for anyways

kwm4385:
Gotcha, and now that its fast enough we don't have the need to make the global search the landing.

🚢

PR #1865, opened 2018-10-10
Prevent the optimizer from selecting an index based on `ORDER BY startedAt`.

baconmania:


ssalinas:
🚢

mikebell90:
I did add some utility methods for checking if running Postgres or MySQL in one of the classss. A hack would be to do a conditional there

baconmania:
It should be compatible; this change is equivalent to sorting by a function, which both MySQL and Postgres support. I was originally going to go with an index hint, but that definitely wouldn't have been compatible on the Postgres side.

For what it's worth, we're not intending to merge this change into master.

ssalinas:
@baconmania where are we going with this PR, leaving it in this branch for now or do we have a version we could merge to master?

ssalinas:
@baconmania Since we've found the issue internally, should we update this PR to instead be a poller that occasionally runs a table stats update? Thinking update rather than just closing since it'd be a pain to unmerge these commits from the branches

baconmania:
We're handling this internally in a different way now, so I'm closing this.

PR #1866, opened 2018-10-16
Clean up logrotate files sooner

ssalinas:
Currently when a task fails we don't clean up certain pieces for a while. We should clean up logs/logrotate right away so there aren't piles of stale logrotate configs pointing to directories that don't exist


pschoenfelder:
🚢

PR #1867, opened 2018-10-17
Upgrade to mesos 1.6.1 and fork of mesos-rxjava

ssalinas:
We ran into a few network partition bugs with the mesos master that are fixed in later versions. This will allow us to update to mesos 1.6.1. Mesos 1.6 uses protobuf 3, which means we also had to upgrade mesos-rxjava. Since the project is seemingly abandoned, we now have our own fork at https://github.com/HubSpot/mesos-rxjava, which is now updated with the newer mesos and protobuf versions. PR still needs a good bit of testing


ssalinas:
built and pushed `hubspot/singularitybase:1.6.1-1` docker image with mesos 1.6.1 installed

ssalinas:
In local testing, Singularity running 1.6.1 could still effectively communicate with a 1.1.0 mesos master. Will also check the reverse to determine best upgrade order. As normal, slaves/executors should likely be last

ssalinas:
Upgrading master/scheduler in either order works fine. new executor (1.6.1) seems to be able to run fine on old mesos (1.1.0) hosts as well

ssalinas:
Note for posterity, not going all the way to 1.7 because it requires a higher minimum docker version than we are running

baconmania:
:ship:

PR #1868, opened 2018-10-19
Email unsubscription bugfix

pschoenfelder:
cc @ssalinas

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/smtp/SmtpMailer.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/smtp/SmtpMailer.java
@@ -737,12 +737,8 @@ public class SmtpMailer implements SingularityMailer, Managed {
       }
     }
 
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/smtp/SmtpMailer.java
```diff
@@ -742,7 +742,7 @@ private void queueMail(final Collection<SingularityEmailDestination> destination
     ccList.removeAll(emailBlacklist);
 
     if (actionTaker.isPresent() && !Strings.isNullOrEmpty(actionTaker.get())) {
-      if (destination.contains(SingularityEmailDestination.ACTION_TAKER)) {
+      if (destination.contains(SingularityEmailDestination.ACTION_TAKER) && !emailBlacklist.contains(actionTaker.get())) {
```
ssalinas: why not just move the removeAll after the action take block?

ssalinas:
🚢

PR #1869, opened 2018-10-22
Faster run now enqueue

ssalinas:
There are a number of zk calls associated with hydrating the object returned by a run now request. For a client that does not need the information, offer the option to skip it


pschoenfelder:
🚢

PR #1871, opened 2018-10-24
Add ability to ignore logrotate hourly output

ssalinas:
We often see the output of this flooding cron emails as tasks shut down or oom, since any cron output gets emailed. This will allow us to ignore that output since we already have executor cleanup to eventually come around and clean these up

cc @pranav


pschoenfelder:
🚢

PR #1873, opened 2018-10-31
Non-web health checks

pschoenfelder:


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
```diff
@@ -83,6 +102,10 @@ public SingularityExecutorMonitor(@Named(SingularityExecutorModule.ALREADY_SHUT_
     this.threadChecker = threadChecker;
     this.threadChecker.start(this);
 
+    this.requestManager = requestManager;
+    this.deployManager = deployManager;
+    this.taskManager = taskManager;
```
ssalinas: SingularityData stuff would be a very heavy dep to add into every executor. I do not think we want to add these here.
- `requestManager` data your fetching here isn't actually being used
- `HealthcheckOptions` we should feed in through the `SingularityTaskExecutorData` like you are doing with the path
- There is no valuable information in saving health check results for the task since all we are doing is checking a file. There will essentially be a single pass/fail health check result, and we can even leave that as encoded in the task status update to keep things simpler if we want. The basics would be that the status update of running serves as our healthcheck success, or a status update of killing/failed/etc serves as our failure.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorModule.java
```diff
@@ -38,6 +39,7 @@
 
   @Override
   protected void configure() {
+    install(new SingularityServiceModule());
```
ssalinas: side note to bringing in this dep, installing this module would also likely start up a number of pollers/db connections/zk connections/etc which we don't want

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorMonitor.java
```diff
@@ -328,6 +351,52 @@ private void onSuccessThrows(ProcessBuilder processBuilder) {
...
+                    healthCheckEnd,
+                    Optional.absent(),
+                    Optional.of("Task timed out on health checks."),
+                    taskId,
+                    Optional.of(true)));
+                return;
+              }
+            }
+
             processRunningTasks.put(task.getTaskId(), submitProcessMonitor(task, processBuilder));
```
ssalinas: The task process doesn't actually start running until after this line. We'd want to make sure the process + monitors are up and running before we start looking at health checks. Ideally it is after we start and right before the TASK_RUNNING. Something like [here](https://github.com/HubSpot/Singularity/blob/master/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java#L27) I think would make it much simpler.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityDeploy.java
```diff
@@ -89,6 +89,7 @@
   private final Optional<Long> healthcheckMaxTotalTimeoutSeconds;
 
   private final Optional<HealthcheckOptions> healthcheck;
+  private final Optional<String> healthCheckResultFilePath;
```
ssalinas: I think we can make this part of the `HealthcheckOptions` object. There are healthcheck-related fields outside that object, but in general they are deprecated, in favor of the ones consolidated in `HealthcheckOptions`

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -226,6 +226,9 @@
   @JsonProperty
   private boolean useFileAttributes = false;
 
+  @JsonProperty
+  private int healthcheckMaxRetries = 30;
```
ssalinas: Maybe name this `default...` instead? Makes it clearer as to why there is a setting here

SingularityExecutorCleanup/src/main/java/com/hubspot/singularity/executor/cleanup/config/SingularityExecutorCleanupConfiguration.java
```diff
@@ -71,6 +72,12 @@
   @NotNull
   private String defaultS3Bucket = "";
 
+  @NotNull
+  private Optional<String> healthCheckResultFilePath = Optional.absent();
```
ssalinas: why are these needed in executor cleanup? It should have no interaction with health checks

SingularityExecutorCleanup/src/main/java/com/hubspot/singularity/executor/cleanup/SingularityExecutorCleanup.java
```diff
@@ -215,7 +215,9 @@ private SingularityExecutorTaskDefinition withDefaults(SingularityExecutorTaskDe
             oldDefinition.getExecutorData().getRequestGroup(),
             oldDefinition.getExecutorData().getS3StorageClass(),
             oldDefinition.getExecutorData().getApplyS3StorageClassAfterBytes(),
-            oldDefinition.getExecutorData().getCpuHardLimit()
+            oldDefinition.getExecutorData().getCpuHardLimit(),
+            oldDefinition.getExecutorData().getHealthCheckResultFilePath().isPresent() ? oldDefinition.getExecutorData().getHealthCheckResultFilePath() : cleanupConfiguration.getHealthCheckResultFilePath(),
```
ssalinas: cleanup shouldn't need to interact with the check at all, can just set it to either `oldDefinition.getExecutorData().get....` or just to absent

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -39,4 +58,27 @@ public String toString() {
...
+      try {
+        Integer healthcheckMaxRetries = maybeOptions.get().getMaxRetries().or(configuration.getHealthcheckMaxRetries());
+
+        Retryer<String> retryer = RetryerBuilder.<String>newBuilder()
+            .retryIfResult(path -> !new File(path).exists())
+            .withWaitStrategy(WaitStrategies.fixedWait(1L, TimeUnit.SECONDS))
+            .withStopStrategy(StopStrategies.stopAfterAttempt(healthcheckMaxRetries))
+            .build();
+
+        retryer.call(() -> expectedHealthCheckResultFilePath.get());
```
ssalinas: would it make more sense for the exists to be here, and the retryer to a have a retryIf predicate instead (i.e. retry if false)?

SingularityExecutorCleanup/src/main/java/com/hubspot/singularity/executor/cleanup/SingularityExecutorCleanup.java
pschoenfelder: This constructs a new `SingularityExecutorTaskDefinition` and I was following the pattern here. I would think just sending `Optional.absent()` would be okay too though. Thougths?

ssalinas: absent would be fine here since it should be unused in this context

pschoenfelder:
@ssalinas What do you think the best way to test this is?

ssalinas:
Will likely just need to create a sample worker in our staging cluster. Can make the command something easy like `sleep 30 && touch ./healthy && sleep 1000` to give you time to look at what's going on

pschoenfelder:
Finally got this working (thanks for the extra 👀@ssalinas).

One last bit I want to resolve is the difference between directories. The directory where the command is run is different from where the health check takes place. What directory do users typically expect to use? If I were a user I wouldn't want to worry about the directory structure for this, and would assume these run from a common directory.
```
    "executorData":{
        "cmd":"bash -c 'sleep 5 && touch ../../healthy.out'",
    },
    "healthcheck": {
        "healthcheckResultFilePath": "healthy.out"
    }
```

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -39,4 +61,32 @@ public String toString() {
     return "SingularityExecutorTaskProcessCallable [task=" + task + "]";
   }
 
+  private void runHealthcheck() {
+    Optional<HealthcheckOptions> maybeOptions = task.getTaskDefinition().getHealthcheckOptions();
+    Optional<String> expectedHealthcheckResultFilePath = task.getTaskDefinition().getHealthcheckResultFilePath();
+
+    String taskAppDirectory = task.getTaskDefinition().getTaskAppDirectory();
```
ssalinas: nit, doesn't seem like this needs to be outside the if statement

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -39,4 +61,31 @@ public String toString() {
...
+      LOG.debug("Checking for healthcheck file {}", expectedHealthcheckResultFilePath.get());
+      String taskAppDirectory = task.getTaskDefinition().getTaskAppDirectory();
+      File fullHealthcheckPath = Paths.get(taskAppDirectory, expectedHealthcheckResultFilePath.get()).toFile();
+
+      try {
+        Integer healthcheckMaxRetries = maybeOptions.get().getMaxRetries().or(configuration.getDefaultHealthcheckMaxRetries());
+
+        Retryer<Boolean> retryer = RetryerBuilder.<Boolean>newBuilder()
+            .retryIfResult(bool -> !bool)
+            .withWaitStrategy(WaitStrategies.fixedWait(1L, TimeUnit.SECONDS))
```
ssalinas: missed this one earlier. There is a healthcheck interval in the healthcheck options as well. Right now we are essentially treating the max retries as a timeout in seconds, which might be . bit confusing to the user given how the web health checks work. If we take the interval into account here then the math for total timeout works the same as it does for web service checks

ssalinas:
🚢

PR #1876, opened 2018-11-06
Better alerting for task lag

sjeropkipruto:


SingularityBase/src/main/java/com/hubspot/singularity/SingularityState.java
```diff
@@ -85,6 +86,7 @@ public SingularityState(@JsonProperty("activeTasks") int activeTasks,
                           @JsonProperty("oldestDeployStep") long oldestDeployStep,
                           @JsonProperty("activeDeploys") List<SingularityDeployMarker> activeDeploys,
                           @JsonProperty("lateTasks") int lateTasks,
+                          @JsonProperty("listLateTasks") List<SingularityPendingTask> listLateTasks,
```
ssalinas: we probably just want to use the SingularityPendingTaskId here. That will give us the basic information about the request/deploy it belongs to. The SingularityPendingTask can be a very large amount of data (includes lists of artifacts, environment variables, etc, etc), which we wouldn't want to be sending on every call for the state

SingularityBase/src/main/java/com/hubspot/singularity/SingularityScheduledTasksInfo.java
```diff
@@ -1,25 +1,28 @@
...
     this.timestamp = timestamp;
   }
 
-  public int getNumLateTasks() {
-    return numLateTasks;
+  public List<SingularityPendingTaskId> getLateTasks() {
+    return lateTasks;
   }
 
+  public int getNumLateTasks() { return getLateTasks().size(); }
```
ssalinas: formatting nit, we usually use the new lines for these cases still

```java
public int getNumLateTasks() {
  return getLateTasks().size();
}
```

ssalinas:
🚢

PR #1877, opened 2018-11-14
Updating the alert banner to only show when there's a widespread issue

sjeropkipruto:
Currently, the UI shows a banner when only a small count of tasks are overdue (when max task lag is greater than 3 minutes). I'm updating the UI to show an alert banner only when there's a widespread issue (when the count of distinct requests with overdue tasks is greater than 10).

### DIFF
--- a/SingularityUI/app/components/common/Application.jsx
+++ b/SingularityUI/app/components/common/Application.jsx
@@ -21,6 +21,10 @@ class Application extends Component {
     );
   }
 
+  componentDidMount() {
+    this.notifyLateRequests(this.props.listLateTasks);
+  }
+
   componentWillUnmount() {
     clearTimeout(this.reenableTaskLagNotificationTimeoutId);
   }
@@ -48,7 +52,6 @@ class Application extends Component {
   }
 
   render() {
-    this.notifyLateRequests(this.props.listLateTasks);
     return (
       <div>
         <Title routes={this.props.routes} params={this.props.params} />
--- a/SingularityUI/app/components/common/Application.jsx
+++ b/SingularityUI/app/components/common/Application.jsx
@@ -34,7 +34,7 @@ class Application extends Component {
 
   notifyLateRequests(listLateTasks) {
     const { canShowTaskLagNotification: canNotify } = this.state;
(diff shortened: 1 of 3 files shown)


SingularityUI/app/utils.es6
```diff
@@ -228,6 +228,17 @@ const Utils = {
     return splits.slice(0, splits.length - 5).join('-');
   },
 
+  getListOfUniqueRequestsFromListOfTasks(listOfTasks) {
+    const res = [];
+    for (const TaskId in listOfTasks) {
```
baconmania: With arrays, `for ... of` is preferable to `for ... in`

SingularityUI/app/utils.es6
```diff
@@ -228,6 +228,17 @@ const Utils = {
...
+      requestId = TaskId.requestId;
```
baconmania: nit: we generally want `UpperCamelCase` for classes and `lowerCamelCase` for instances.

SingularityUI/app/components/common/Application.jsx
```diff
@@ -32,23 +32,23 @@ class Application extends Component {
...
+          notified. (Number of distinct late requests: ${lateRequests.length})
         `,
       });
     }
   }
 
   render() {
-    this.notifyLag(this.props.maxTaskLag);
-
+    this.notifyLateRequests(this.props.listLateTasks);
```
kwm4385: I see this was here before, but this call should probably go in `componentDidMount`

SingularityUI/app/utils.es6
```diff
@@ -228,6 +228,17 @@ const Utils = {
...
+      if (!res.contains(requestId)){
+        res.push(requestId);
+      }
+    }
+    return res;
```
kwm4385: Underscore already has the utility function `_.uniq(listOfTasks)`

SingularityUI/app/components/common/Application.jsx
sjeropkipruto: Since we want to make this call whenever the listOfLateTasks prop changes, I'll put it in componentDidUpdate

baconmania:
🚢

PR #1878, opened 2018-11-15
Add support for inactive task history filtered by deploy ID

mlr46:
Inactive tasks can be filtered by deploy ID in the UI; adding support for this in the client as well.


ssalinas:
🚢

PR #1879, opened 2018-11-15
Make task history paging less flaky

ssalinas:
Fixes the task history page size to also be reflected in query params so that a refresh of this will not interrupt the current view

### DIFF
--- a/SingularityUI/app/actions/ui/requestDetail.es6
+++ b/SingularityUI/app/actions/ui/requestDetail.es6
@@ -19,7 +19,6 @@ export const refresh = (requestId, location) => (dispatch, getState) => {
 
   const { taskHistoryPage, taskHistoryCount } = location.query;
   if (taskHistoryPage == null || taskHistoryPage == 1) {
-  console.log(taskHistoryPage)
     dispatch(FetchTaskHistoryForRequest.trigger(requestId, taskHistoryCount == null ? 10 : taskHistoryCount, taskHistoryPage == null ? 1 : taskHistoryPage));
   }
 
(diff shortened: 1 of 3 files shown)


SingularityUI/app/actions/ui/requestDetail.es6
```diff
@@ -8,15 +8,21 @@ import {
...
     dispatch(FetchRequestHistory.trigger(requestId, 5, 1))
   ])
 
   dispatch(FetchActiveTasksForRequest.trigger(requestId));
   dispatch(FetchTaskCleanups.trigger());
-  dispatch(FetchTaskHistoryForRequest.trigger(requestId, 5, 1));
+
+  const { taskHistoryPage, taskHistoryCount } = location.query;
+  if (taskHistoryPage == null || taskHistoryPage == 1) {
+  console.log(taskHistoryPage)
```
baconmania: Can probably remove this if we're done testing?

baconmania:
🚢

SingularityUI/app/actions/ui/requestDetail.es6
```diff
@@ -8,15 +8,20 @@ import {
 import { FetchTaskCleanups, FetchScheduledTasksForRequest } from '../api/tasks';
 import { FetchRequestUtilization } from '../api/utilization';
 
-export const refresh = (requestId) => (dispatch, getState) => {
+export const refresh = (requestId, location) => (dispatch, getState) => {
```
andyhuang91: One option is to not pass in location and instead use default parameters like this
```
export const refresh = (requestId, taskHistoryPage = 1, taskHistoryCount = 10) =>
```
pageSize might also be a better name than count

baconmania:
🚢

PR #1880, opened 2018-11-16
Improve ExtendedTaskState readability

andybergon:
I know that by the [Google Style docs](https://google.github.io/styleguide/javaguide.html#s4.8.1-enum-classes) this is optional but it improves readability. Do we mind having those line breaks?


ssalinas:
I'm fine with putting it on multiple lines, however at the moment it's failing checkstyle due to trailing spaces on each line

ssalinas:
🚢

PR #1881, opened 2018-11-16
Don't count pending tasks limited by instance count in task lag 

sjeropkipruto:
Currently, if you submit pending tasks in quick succession on an on demand request that has an instance count limit, you can end up with tasks that will wait for the first to complete before launching. We shouldn't page for task lag in these cases. I've separated the late tasks on on-demand requests from the rest of the late tasks. This allows us to get better metrics on task lag and improve our alerts.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
@@ -10,6 +10,7 @@ import java.util.stream.Collectors;
 
 import javax.inject.Singleton;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.imps.CuratorFrameworkState;
 import org.apache.curator.utils.ZKPaths;
@@ -26,6 +27,7 @@ import com.google.inject.Inject;
 import com.google.inject.name.Named;
 import com.hubspot.mesos.CounterMap;
 import com.hubspot.mesos.JavaUtils;
+import com.hubspot.singularity.RequestType;
 import com.hubspot.singularity.SingularityCreateResult;
 import com.hubspot.singularity.SingularityDeployMarker;
 import com.hubspot.singularity.SingularityHostState;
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
```diff
@@ -322,6 +328,22 @@ public SingularityState generateState(boolean includeRequestIds) {
...
   }
 
+  private ImmutablePair<List<SingularityPendingTaskId>, List<SingularityPendingTaskId>> filterOnDemandRequests (List<SingularityPendingTaskId> lateTasks) {
+    List<SingularityPendingTaskId> onDemandLateTasks = new ArrayList<>();
+    List<SingularityPendingTaskId> nonOnDemandLateTasks = new ArrayList<>();
+
+    List<String> requestIds = lateTasks.stream().map(lateTask -> lateTask.getRequestId()).collect(Collectors.toList());
+    List<RequestType> requestTypes = requestManager.getRequests(requestIds).stream().map(s -> s.getRequest().getRequestType()).collect(Collectors.toList());
+    for (int i = 0; i < lateTasks.size(); i++) {
+      if (requestTypes.get(i).equals(RequestType.ON_DEMAND)){
```
baconmania: Hm, I'm not sure this'll work as intended. For example, if we have two late tasks from the same request ID (so `requestIds` would be `[requestIdA, requestIdA]`), I think `requestManager.getRequests([requestIdA, requestIdA])` would just return `[requestA]`. Then the `requestTypes.get(i)` indexing wouldn't work as expected.

baconmania: An alternative could be something like this:

```java
Map<Boolean, List<SingularityPendingTaskId>> lateTasksPartitionedByOnDemand = lateTasks.stream()
  .collect(
    Collectors.partitioningBy(lateTask -> requestManager.getRequest(lateTask.getRequestId()).getRequestType() == RequestType.ON_DEMAND)
  );

// lateTasksPartitionedByOnDemand.get(true) == on demand late tasks
// lateTasksPartitionedByOnDemand.get(false) == everything else
```

baconmania:
🚢

PR #1886, opened 2018-12-11
Syncing upstreams between load balancer and singularity

sjeropkipruto:
We recently had a critsit due to duplicate upstreams. The SingularityUpstreamsPoller runs every 10 minutes. Wrote a poller to sync between Singularity and Baragon (load balancer) to automatically fix this issue on a recurring basis. It fetches upstreams from active & healthy tasks in Singularity and upstreams in BaragonService. It compares the two lists and, if there are any extra upstreams in Baragon, it sends a load balancer request to remove them.


SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
```diff
@@ -68,6 +72,54 @@ public LoadBalancerClientImpl(SingularityConfiguration configuration, ObjectMapp
     this.mesosProtosUtils = mesosProtosUtils;
   }
 
+  public String getLoadBalancerUri() {
```
ssalinas: are these getters used anywhere? No need to make these public when we can just use the private variables

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,80 @@
...
+      final UpstreamInfo upstream = taskToUpstreamInfo(task, loadBalancerUpstreamGroup);
+      if (upstream.equals(null)){
+        upstreams.add(upstream);
+      }
+    }
+    return upstreams;
+  }
+
+  private void syncUpstreams( ) {
+    Collection<UpstreamInfo> upstreamsInBaragon = lbClient.getUpstreams();
```
ssalinas: While it's more api calls, we should probably do this one service at a time for a few reasons:
- We want to compare upstreams within the context of each service. Right now the PR is compiling a list of all upstreams across all tasks/requests. It's very possible that host1:port1 is valid for SomethingWeb but not for SomethingElseWeb (in which case we would remove it from the one it is no longer valid for). We need that context of which request the upstream belongs to to make that distinction (the `request` field of the Baragon UpstreamInfo points to the BaragonRequest id, it is not the same as the SingularityRequest id)
- Updates are happening frequently from other Singularity pollers and we want to be able to lock the request and make sure we are the only ones currently modifying the state of the system. See examples of `runWithRequestLock` for how to lock by request id on the leader. In this way, we can ensure that the BaragonState we are operating on is not also being modified by another Singularity poller at the same time.

SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
```diff
@@ -68,6 +72,54 @@ public LoadBalancerClientImpl(SingularityConfiguration configuration, ObjectMapp
...
+        LOG.info(String.format("Response status code %s", response.getStatusCode()));
+      }
+      return objectMapper.readValue(response.getResponseBodyAsBytes(), new TypeReference<List<BaragonServiceState>>() {});
+    } catch (IOException | InterruptedException | ExecutionException | TimeoutException e) {
+      e.printStackTrace();
+      return Collections.emptyList();
+    }
+  }
+
+  public Collection<UpstreamInfo> getUpstreams () {
```
ssalinas: related to my comment in the poller, we can probably make this take an argument of `String requestId` and return the list of upstreams for only that requestId

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
ssalinas: Additional piece, we can limit the number of requests we loop over by first filtering to only those that have load balancing enabled (boolean property on SingularityRequest object)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,80 @@
...
+  private final LoadBalancerClientImpl lbClient;
+  private final TaskManager taskManager;
+
+  @Inject
+  public SingularityUpstreamChecker(LoadBalancerClientImpl lbClient, TaskManager taskManager) {
+    this.lbClient = lbClient;
+    this.taskManager = taskManager;
+  }
+
+  private UpstreamInfo taskToUpstreamInfo (SingularityTask task, Optional<String> loadBalancerUpstreamGroup) {
```
ssalinas: since this is pretty much copied from the `LoadBalancerClientImpl`, could be worth factoring it out into a public or static method we can call from there

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,80 @@
...
+    UpstreamInfo upstreamInfo = null; //TODO: what is a better way to do this? We cannot initialize an empty upstream
```
ssalinas: we could always return an optional instead:

```
Optional<UpstreamInfo> taskToUpstream() {
  for () {
    ...
    return Optional.of(upstreamInfo);
  }
  return Optional.absent();
}

File: SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java (hunk 1 above)
Comment (sjeropkipruto) [reply SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java]: We use the getters in the SingularityUpstreamChecker, particularly in the taskToUpstreamInfo function

File: SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClient.java [hunk 6]
```diff
@@ -18,4 +21,29 @@
   SingularityLoadBalancerUpdate cancel(LoadBalancerRequestId loadBalancerRequestId);
 
   SingularityLoadBalancerUpdate delete(LoadBalancerRequestId loadBalancerRequestId, String requestId, Set<String> loadBalancerGroups, String serviceBasePath);
+
+  //TODO: complete the javadocs below
+  /**
+   * @param requestId
+   * @return
+   */
+  Collection<UpstreamInfo> getBaragonUpstreamsForRequest (String requestId);
```
ssalinas: formatting nit, no space after method name

SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClient.java
```diff
@@ -18,4 +21,29 @@
...
+
+  /**
+   * @param tasks
+   * @param requestId
+   * @param loadBalancerUpstreamGroup
+   * @return
+   */
+  List<UpstreamInfo> tasksToUpstreams(List<SingularityTask> tasks, String requestId, Optional<String> loadBalancerUpstreamGroup);
```
ssalinas: Maybe a better name is just `getUpstreamsForTasks`?

SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
```diff
@@ -68,6 +72,40 @@ public LoadBalancerClientImpl(SingularityConfiguration configuration, ObjectMapp
...
 
+  private String getStateUriFromRequestUri (){
+    return loadBalancerUri.replace("request", "state");
+  }
+
+  private String getLoadBalancerStateUri (String requestId){
+    return String.format(OPERATION_URI, getStateUriFromRequestUri(), requestId);
+  }
+
+  private Optional<BaragonServiceState> getBaragonServiceStateForLoadBalancerRequest (String requestId) {
```
ssalinas: similar formatting nit, no space after method name

SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
```diff
@@ -68,6 +72,40 @@ public LoadBalancerClientImpl(SingularityConfiguration configuration, ObjectMapp
...
+    } catch (IOException | InterruptedException | ExecutionException | TimeoutException e) {
+      e.printStackTrace();
+      return Optional.absent();
+    }
+  }
+
+  public Collection<UpstreamInfo> getBaragonUpstreamsForRequest (String requestId) {
+    Optional<BaragonServiceState> maybeBaragonServiceState = getBaragonServiceStateForLoadBalancerRequest(requestId);
+    if (maybeBaragonServiceState.isPresent()){
+      BaragonServiceState baragonServiceState = getBaragonServiceStateForLoadBalancerRequest(requestId).get();
```
ssalinas: don't need to call `getBaragonServiceStateForLoadBalancerRequest` twice here, which will make the http call twice. This can just be `maybeBaragonServiceState.get()`

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Inject
+  public SingularityUpstreamChecker(LoadBalancerClientImpl lbClient, TaskManager taskManager, RequestManager requestManager, DeployManager deployManager) {
+    this.lbClient = lbClient;
+    this.taskManager = taskManager;
+    this.requestManager = requestManager;
+    this.deployManager = deployManager;
+  }
+
+  private List<SingularityTask> getActiveSingularityTasksForRequest (String requestId) {
```
ssalinas: same formatting nit

ssalinas: There is a `public Map<SingularityTaskId, SingularityTask> getTasks(Iterable<SingularityTaskId> taskIds) {` method in `TaskManager` that will make this much easier. It will return the SingularityTask object for each of the provided task ids

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  private Collection<UpstreamInfo> getUpstreamsFromActiveTasks(String requestId, Optional<String> loadBalancerUpstreamGroup) {
+    return lbClient.tasksToUpstreams(getActiveSingularityTasksForRequest(requestId), requestId, loadBalancerUpstreamGroup);
+  }
+
+  private SingularityLoadBalancerUpdate syncUpstreamsForService(SingularityRequest request, String requestId, SingularityDeploy deploy, Optional<String> loadBalancerUpstreamGroup) {
+    Collection<UpstreamInfo> upstreamsInBaragonForRequest = lbClient.getBaragonUpstreamsForRequest(requestId);
+    Collection<UpstreamInfo> upstreamsInSingularityForRequest = getUpstreamsFromActiveTasks(requestId, loadBalancerUpstreamGroup);
+    upstreamsInBaragonForRequest.removeAll(upstreamsInSingularityForRequest);
```
ssalinas: we will only want to compare the `upstream` and `group` fields (maybe also `rackId` though not as important) of the `UpstreamInfo` object when determining if they are equal. The `requestId` field can be different and is an indicator of the unique baragon request id that originally added the upstream (not the Singularity request id). Right now we will not actually remove anything since the objects `equals` method also compares requestId

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,81 @@
...
+    return lbClient.makeAndSendBaragonRequest(loadBalancerRequestId, Collections.emptyList(), new ArrayList<>(upstreamsInBaragonForRequest), deploy, request);
+  }
+
+  public void syncUpstreams() {
+    for (SingularityRequestWithState singularityRequestWithState: requestManager.getActiveRequests()){
+      SingularityRequest request = singularityRequestWithState.getRequest();
+      if (request.isLoadBalanced()) {
+        //TODO: lock on the requestId
+        String requestId = singularityRequestWithState.getRequest().getId();
+        String deployId = deployManager.getInUseDeployId(requestId).get(); //TODO: handle when it's absent
```
ssalinas: can skip when this is absent since we don't have the required data to do lb operations

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private Collection<UpstreamInfo> getUpstreamsFromActiveTasks(String requestId, Optional<String> loadBalancerUpstreamGroup) {
+    return lbClient.tasksToUpstreams(getActiveSingularityTasksForRequest(requestId), requestId, loadBalancerUpstreamGroup);
+  }
+
+  private SingularityLoadBalancerUpdate syncUpstreamsForService(SingularityRequest request, String requestId, SingularityDeploy deploy, Optional<String> loadBalancerUpstreamGroup) {
+    Collection<UpstreamInfo> upstreamsInBaragonForRequest = lbClient.getBaragonUpstreamsForRequest(requestId);
+    Collection<UpstreamInfo> upstreamsInSingularityForRequest = getUpstreamsFromActiveTasks(requestId, loadBalancerUpstreamGroup);
+    upstreamsInBaragonForRequest.removeAll(upstreamsInSingularityForRequest);
+    LoadBalancerRequestId loadBalancerRequestId = new LoadBalancerRequestId(requestId, LoadBalancerRequestType.REMOVE, Optional.absent());
```
ssalinas: The `requestId` here will be the baragon service request id as well. We don't want this to be duplicated on any short time scale (I think Baragon purges old requests every few days right now). This should be a more unique id that uniquely identifies the individual update we're making to baragon.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,81 @@
...
+    return lbClient.makeAndSendBaragonRequest(loadBalancerRequestId, Collections.emptyList(), new ArrayList<>(upstreamsInBaragonForRequest), deploy, request);
```
ssalinas: we should check somewhere to be sure that this request succeeded. This code will enqueue it, but we don't know if it actually finished successfull.y Normally we have to poll baragon (using the request id mentioned in the comment above) to find out if the update actually succeeded. Not certain yet if we want to do that within the context of a single poller run (though this normally will only take a few seconds)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamPoller.java
```diff
@@ -0,0 +1,32 @@
...
+  SingularityUpstreamPoller(SingularityConfiguration configuration, SingularityUpstreamChecker upstreamChecker) {
+    super(configuration.getCheckUpstreamsEverySeconds(), TimeUnit.SECONDS, true);
+
+    this.upstreamChecker = upstreamChecker;
+
+  }
+
+  @Override
+  public void runActionOnPoll() {
+    LOG.info("Checking upstreams");
```
ssalinas: Can we add in a configuration param for whether or not this poller is enabled and short circuit here if needed? Will make it much easier to turn off the functionality in case something ever goes wrong with it. (vs rolling back builds/etc)

SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClient.java
```diff
@@ -18,4 +21,10 @@
   SingularityLoadBalancerUpdate cancel(LoadBalancerRequestId loadBalancerRequestId);
 
   SingularityLoadBalancerUpdate delete(LoadBalancerRequestId loadBalancerRequestId, String requestId, Set<String> loadBalancerGroups, String serviceBasePath);
+
+  Collection<UpstreamInfo> getBaragonUpstreamsForRequest(String requestId);
```
ssalinas: nit on naming. The original intent of the interface here was that we could eventually implement a non-baragon-specific one if needed. While it doesn't seem like that will actually happen, the more correct naming here would remove baragon from the method name (more like just `getUpstreamsForRequest`). Similar for the`makeAndSendBaragonRequest`

SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
```diff
@@ -68,6 +72,40 @@ public LoadBalancerClientImpl(SingularityConfiguration configuration, ObjectMapp
...
+    try {
+      ListenableFuture<Response> future = httpClient.executeRequest(request);
+      Response response = future.get(loadBalancerTimeoutMillis, TimeUnit.MILLISECONDS);
+      if (!JavaUtils.isHttpSuccess(response.getStatusCode())) {
+        LOG.info(String.format("Response status code %s", response.getStatusCode()));
+      }
+      return objectMapper.readValue(response.getResponseBodyAsBytes(), new TypeReference<BaragonServiceState>() {});
+    } catch (IOException | InterruptedException | ExecutionException | TimeoutException e) {
+      e.printStackTrace();
+      return Optional.absent();
```
ssalinas: I'm wondering if we should propagate this exception rather than swallowing it and returning absent here. I can see a case where Baragon is down/times out/etc and we then try to re-add upstreams it already has because this returned that the state was not present. Maybe we want to catch this exception in the actual sync poller since the poller will have more context about the appropriate action to take when that happens

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,106 @@
...
+
+    return lbClient.makeAndSendBaragonRequest(loadBalancerRequestId, Collections.emptyList(), extraUpstreams, deploy, singularityRequest);
+  }
+
+  public void syncUpstreams() {
+    for (SingularityRequestWithState singularityRequestWithState: requestManager.getActiveRequests()){
+      final SingularityRequest singularityRequest= singularityRequestWithState.getRequest();
+      if (singularityRequest.isLoadBalanced()) {
+        final String singularityRequestId = singularityRequest.getId(); //TODO: lock on the requestId
+        final Optional<String> maybeDeployId = deployManager.getInUseDeployId(singularityRequestId);
```
ssalinas: additional thing. I'm wondering if we should skip the sync if there is a deploy in progress. I'm trying to think through all the ways this could possibly corrupt state, and I could see a deploy and bounce being problematic. See my comment on `getActiveTasksForRequest` for more detail

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,106 @@
...
+  @Inject
+  public SingularityUpstreamChecker(LoadBalancerClientImpl lbClient, TaskManager taskManager, RequestManager requestManager, DeployManager deployManager) {
+    this.lbClient = lbClient;
+    this.taskManager = taskManager;
+    this.requestManager = requestManager;
+    this.deployManager = deployManager;
+  }
+
+  private List<SingularityTask> getActiveTasksForRequest(String requestId) {
+    final Map<SingularityTaskId, SingularityTask> activeTasksForRequest = taskManager.getTasks(taskManager.getActiveTaskIdsForRequest(requestId));
```
ssalinas: Here, we will want to make sure that we are looking specifically at healthy tasks. If a bounce or deploy is in progress, there will be a number of tasks that are considered 'active' in terms of mesos status, but not healthy and ready to be in the load balancer. We do not want to mistakenly add these too early. You can check out `RequestHelper.getTaskIdsByStatusForRequest` for an example of finding tasks that are active + healthy for a request

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  private List<SingularityTask> getActiveTasksForRequest(String requestId) {
+    final Optional<SingularityTaskIdsByStatus> taskIdsByStatusForRequest = requestHelper.getTaskIdsByStatusForRequest(requestId);
+    if (taskIdsByStatusForRequest.isPresent()) {
+      final List<SingularityTaskId> activeHealthyTaskIdsForRequest = taskIdsByStatusForRequest.get().getHealthy();
+      final Map<SingularityTaskId, SingularityTask> activeTasksForRequest = taskManager.getTasks(activeHealthyTaskIdsForRequest);
+      return new ArrayList<>(activeTasksForRequest.values());
+    }
+    return new ArrayList<>();
```
ssalinas: I think this is another case where I'd lean towards throwing an exception and essentially skipping to the next request we need to check. If the taskIds is absent that means we couldn't get task data for some reason. I'd be wary of making any sort of LB update when we did not correctly fetch task data.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  private boolean isEqualUpstreamGroupRackId(UpstreamInfo upstream1, UpstreamInfo upstream2){
+    return (upstream1.getUpstream().equals(upstream2.getUpstream()))
+        && (upstream1.getGroup().equals(upstream2.getGroup()))
+        && (upstream1.getRackId().equals(upstream2.getRackId()));
+  }
+
+  private Collection<UpstreamInfo> getEqualUpstreams(UpstreamInfo upstream, Collection<UpstreamInfo> upstreams) {
+    // We expect that the collection will have a maximum of one match, but we will keep it as a collection just in case
```
ssalinas: +1 for clarifying comment

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,157 @@
...
+        final Optional<SingularityDeploy> maybeDeploy = deployManager.getDeploy(singularityRequestId, deployId);
+        if (maybeDeploy.isPresent()) {
+          final SingularityDeploy deploy = maybeDeploy.get();
+          final Optional<String> loadBalancerUpstreamGroup = deploy.getLoadBalancerUpstreamGroup();
+          final SingularityLoadBalancerUpdate syncUpstreamsUpdate;
+          try {
+            syncUpstreamsUpdate = syncUpstreamsForService(singularityRequest, deploy, loadBalancerUpstreamGroup);
+            final LoadBalancerRequestId loadBalancerRequestId = syncUpstreamsUpdate.getLoadBalancerRequestId();
+            final BaragonRequestState syncUpstreamsState = lbClient.getState(loadBalancerRequestId).getLoadBalancerState();
+            switch (syncUpstreamsState) { //TODO: discuss what to do when the state is not SUCCESS, apart from logging
```
ssalinas: documenting from slack conversation:

the immediate response when you first post it will be `WAITING`
I’d imagine we’d want something like:
- POST it (should be waiting after that)
- Poll until no longer in WAITING (or until some timeout)
- if `SUCCESS` then :thumbsup: , if not then maybe we LOG.error (which will also throw up a sentry that notifies us)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,157 @@
...
+              case CANCELING:
+                LOG.info("Syncing upstreams for singularity request {} is canceling. Load balancer request id is {}.", singularityRequestId, loadBalancerRequestId.toString());
+              case UNKNOWN:
+                LOG.info("Syncing upstreams for singularity request {} is unknown. Load balancer request id is {}.", singularityRequestId, loadBalancerRequestId.toString());
+              case WAITING:
+                LOG.info("Syncing upstreams for singularity request {} is waiting. Load balancer request id is {}.", singularityRequestId, loadBalancerRequestId.toString());
+                break;
+            }
+          } catch (IOException | InterruptedException | ExecutionException | TimeoutException e) {
+            e.printStackTrace();
```
ssalinas: Generally prefer using a `LOG.error("message", e)` here. This will send the exception through our normal logging stack and will also trigger a sentry automatically for anything at error level

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,163 @@
...
+        final Optional<SingularityDeploy> maybeDeploy = deployManager.getDeploy(singularityRequestId, deployId);
+        if (maybeDeploy.isPresent()) {
+          final SingularityDeploy deploy = maybeDeploy.get();
+          final Optional<String> loadBalancerUpstreamGroup = deploy.getLoadBalancerUpstreamGroup();
+          final SingularityLoadBalancerUpdate syncUpstreamsUpdate;
+          try {
+            syncUpstreamsUpdate = syncUpstreamsForService(singularityRequest, deploy, loadBalancerUpstreamGroup);
+            final LoadBalancerRequestId loadBalancerRequestId = syncUpstreamsUpdate.getLoadBalancerRequestId();
+            BaragonRequestState syncUpstreamsState = lbClient.getState(loadBalancerRequestId).getLoadBalancerState();
+            while (syncUpstreamsState.equals(BaragonRequestState.WAITING)) { // continue polling until it's no longer WAITING
```
ssalinas: nit Since this is an enum, can use `==` here

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,170 @@
...
+            checkSyncUpstreamsState(loadBalancerRequestId, singularityRequestId);
+          } catch (Exception e) {
+            LOG.error("message", e);
+          }
+        }
+      }
+    }
+  }
+
+  private Predicate<BaragonRequestState> baragonRequestStateIsWaiting = baragonRequestState -> baragonRequestState == BaragonRequestState.WAITING;
```
ssalinas: for styling purposes, easier to find this if it is at the top of the class and made private static final
```java
private static final Predicate<BaragonRequestState> IS_WAITING_STATE = baragonRequestState -> baragonRequestState == BaragonRequestState.WAITING;
```

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,170 @@
...
+        .build();
+    try {
+      BaragonRequestState syncUpstreamsState = syncingRetryer.call(() -> lbClient.getState(loadBalancerRequestId).getLoadBalancerState());
+      if (syncUpstreamsState == BaragonRequestState.SUCCESS){
+        LOG.info("Syncing upstreams for singularity request {} is {}. Load balancer request id is {}.", singularityRequestId, syncUpstreamsState.name(), loadBalancerRequestId.toString());
+      } else {
+        LOG.error("Syncing upstreams for singularity request {} is {}. Load balancer request id is {}.", singularityRequestId, syncUpstreamsState.name(), loadBalancerRequestId.toString());
+      }
+    } catch (Exception e) {
+      LOG.error("message", e);
```
ssalinas: Note to update logging here eventually, e.g. 'Could not sync state for request id {}', requestid, e

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,170 @@
...
+        .retryIfException()
+        .withWaitStrategy(WaitStrategies.fixedWait(1, TimeUnit.SECONDS))
+        .retryIfResult(baragonRequestStateIsWaiting)
+        .build();
+    try {
+      BaragonRequestState syncUpstreamsState = syncingRetryer.call(() -> lbClient.getState(loadBalancerRequestId).getLoadBalancerState());
+      if (syncUpstreamsState == BaragonRequestState.SUCCESS){
+        LOG.info("Syncing upstreams for singularity request {} is {}. Load balancer request id is {}.", singularityRequestId, syncUpstreamsState.name(), loadBalancerRequestId.toString());
+      } else {
+        LOG.error("Syncing upstreams for singularity request {} is {}. Load balancer request id is {}.", singularityRequestId, syncUpstreamsState.name(), loadBalancerRequestId.toString());
```
ssalinas: Wonder if we should log the full content of `lbClient.getState(loadBalancerRequestId)` if it fails here, not just the load balancer state. The full json will have a message to it that details the reason the update failed, which will make debugging much easier should something go wrong

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  private SingularityLoadBalancerUpdate syncUpstreamsForService(SingularityRequest singularityRequest, SingularityDeploy deploy, Optional<String> loadBalancerUpstreamGroup) throws Exception {
+    final String singularityRequestId = singularityRequest.getId();
+    final LoadBalancerRequestId loadBalancerRequestId = new LoadBalancerRequestId(singularityRequestId, LoadBalancerRequestType.REMOVE, Optional.absent());
+    Collection<UpstreamInfo> upstreamsInBaragonForRequest = lbClient.getLoadBalancerUpstreamsForRequest(loadBalancerRequestId.toString());
+    Collection<UpstreamInfo> upstreamsInSingularityForRequest = getUpstreamsFromActiveTasksForRequest(singularityRequestId, loadBalancerUpstreamGroup);
+    final List<UpstreamInfo> extraUpstreams = getExtraUpstreams(upstreamsInBaragonForRequest, upstreamsInSingularityForRequest);
+    LOG.info("Making and sending load balancer request to remove {} extra upstreams.", extraUpstreams.size());
+    LOG.info("Upstreams removed are: {}", Arrays.toString(extraUpstreams.toArray()));
```
ssalinas: the logger will format this for you, can just put `extraUpstreams` as the arg here

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,170 @@
...
+        .retryIfException()
+        .withWaitStrategy(WaitStrategies.fixedWait(1, TimeUnit.SECONDS))
+        .retryIfResult(IS_WAITING_STATE)
+        .build();
+    try {
+      BaragonRequestState syncUpstreamsState = syncingRetryer.call(() -> lbClient.getState(loadBalancerRequestId).getLoadBalancerState());
+      if (syncUpstreamsState == BaragonRequestState.SUCCESS){
+        LOG.info("Syncing upstreams for singularity request {} is {}.", singularityRequestId, lbClient.getState(loadBalancerRequestId).toString());
+      } else {
+        LOG.error("Syncing upstreams for singularity request {} is {}.", singularityRequestId, lbClient.getState(loadBalancerRequestId).toString());
```
ssalinas: for this error case we will want to make sure the `message` from the full `SingularityLoadBalancerUpdate` (from `lbClient.getState`) is logged. That will contain the actual reason for why the update failed if we need it for debugging. So you'll need to just change the retryer to return the full object and the predicate to check the field

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,192 @@
...
+        .retryIfException()
+        .withWaitStrategy(WaitStrategies.fixedWait(1, TimeUnit.SECONDS))
+        .retryIfResult(IS_WAITING_STATE)
+        .build();
+    try {
+      SingularityLoadBalancerUpdate syncUpstreamsState = syncingRetryer.call(() -> lbClient.getState(loadBalancerRequestId));
+      if (syncUpstreamsState.getLoadBalancerState() == BaragonRequestState.SUCCESS){
+        LOG.info("Syncing upstreams for singularity request {} is {}.", singularityRequestId, syncUpstreamsState.toString());
+      } else {
+        LOG.error("Syncing upstreams for singularity request {} is {}.", singularityRequestId, syncUpstreamsState.toString());
```
ssalinas: Can we log out the `message` (or maybe the whole `SingularityLoadBalancerUpdate` object) here as well? That will have the specific error that caused the update to fail.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,192 @@
...
+    final long start = System.currentTimeMillis();
+    try {
+      Collection<UpstreamInfo> upstreamsInLoadBalancerForRequest = getLoadBalancerUpstreamsForService(singularityRequest.getId());
+      LOG.info("Upstreams in load balancer for service {} are {}.", singularityRequest.getId(), upstreamsInLoadBalancerForRequest);
+      Collection<UpstreamInfo> upstreamsInSingularityForRequest = getUpstreamsFromActiveTasksForRequest(singularityRequest.getId(), loadBalancerUpstreamGroup);
+      LOG.info("Upstreams in singularity for service {} are {}.", singularityRequest.getId(), upstreamsInSingularityForRequest);
+      final List<UpstreamInfo> extraUpstreams = getExtraUpstreamsInLoadBalancer(upstreamsInLoadBalancerForRequest, upstreamsInSingularityForRequest);
+      if (extraUpstreams.size() > 0){
+        LOG.info("Syncing upstreams for service {}. Making and sending load balancer request {} to remove {} extra upstreams. The upstreams removed are: {}.", singularityRequest.getId(), loadBalancerRequestId, extraUpstreams.size(), extraUpstreams);
```
ssalinas: This is probably the only one we need to have at info level. The rest of the log lines will get a bit noisy and can be debug/trace

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,192 @@
...
+        return Optional.of(lbClient.makeAndSendLoadBalancerRequest(loadBalancerRequestId, Collections.emptyList(), extraUpstreams, deploy, singularityRequest));
```
ssalinas: Should this have a separate catch around it? I feel like there are two cases we are lumping together here:
1) We catch an exception getting the upstreams (i.e. before we `makeAndSendLoadBalancerRequest`). Could log it and return Optional.absent() there since there is no request to poll for results
2) We catch an exception from `makeAndSendLoadBalancerRequest`, which can fall to the `BaragonRequestState.UNKNOWN` case you have in the current catch, since we don't know if the lb update went through or not

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,192 @@
...
+
+  private List<SingularityTask> getActiveHealthyTasksForRequest(String requestId) throws Exception {
+    final Optional<SingularityTaskIdsByStatus> taskIdsByStatusForRequest = requestHelper.getTaskIdsByStatusForRequest(requestId);
+    if (taskIdsByStatusForRequest.isPresent()) {
+      final List<SingularityTaskId> activeHealthyTaskIdsForRequest = taskIdsByStatusForRequest.get().getHealthy();
+      final Map<SingularityTaskId, SingularityTask> activeTasksForRequest = taskManager.getTasks(activeHealthyTaskIdsForRequest);
+      return new ArrayList<>(activeTasksForRequest.values());
+    }
+    LOG.error("TaskId not found for requestId: {}.", requestId);
+    throw new Exception("TaskId not found.");
```
ssalinas: Is there anything more specific than plain old Exception we can use here?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamPoller.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.hubspot.singularity.SingularityAction;
+import com.hubspot.singularity.config.SingularityConfiguration;
+import com.hubspot.singularity.data.DisasterManager;
+
+@Singleton
+public class SingularityUpstreamPoller extends SingularityLeaderOnlyPoller {
```
ssalinas: Last thing for this class, there is an `abortsOnError` method in `SingularityLeaderOnlyPoller`. We should override it to return false for this. For some of the leader only pollers that have to do with task state, singularity aborts + restarts to preserve state integrity. That is not necessary for this poller so we can add:
```java
  @Override
  protected boolean abortsOnError() {
    return false;
  }
```

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,197 @@
...
+    try {
+      LOG.trace("Sending load balancer request to sync upstreams for service {}.", singularityRequest.getId());
+      final LoadBalancerRequestId loadBalancerRequestId = new LoadBalancerRequestId(String.format("%s-%s-%s", singularityRequest.getId(), deploy.getId(), System.currentTimeMillis()), LoadBalancerRequestType.REMOVE, Optional.absent());
+      Collection<UpstreamInfo> upstreamsInLoadBalancerForService = getLoadBalancerUpstreamsForService(singularityRequest.getId(), loadBalancerUpstreamGroup);
+      LOG.trace("Upstreams in load balancer for service {} are {}.", singularityRequest.getId(), upstreamsInLoadBalancerForService);
+      Collection<UpstreamInfo> upstreamsInSingularityForService = getUpstreamsFromActiveTasksForService(singularityRequest.getId(), loadBalancerUpstreamGroup);
+      LOG.trace("Upstreams in singularity for service {} are {}.", singularityRequest.getId(), upstreamsInSingularityForService);
+      final List<UpstreamInfo> extraUpstreams = getExtraUpstreamsInLoadBalancer(upstreamsInLoadBalancerForService, upstreamsInSingularityForService);
+      if (extraUpstreams.size() == 0) {
```
ssalinas: another nitpick note, List<T> already has `isEmpty()` versus size == 0

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUpstreamChecker.java
```diff
@@ -0,0 +1,197 @@
...
+      LOG.error("Failed getting load balancer upstreams for singularity request {}. ", singularityRequestId, e);
+    }
+    return Collections.emptyList();
+  }
+
+  private Optional<SingularityLoadBalancerUpdate> syncUpstreamsForServiceHelper(SingularityRequest singularityRequest, SingularityDeploy deploy, Optional<String> loadBalancerUpstreamGroup) {
+    try {
+      LOG.trace("Sending load balancer request to sync upstreams for service {}.", singularityRequest.getId());
+      final LoadBalancerRequestId loadBalancerRequestId = new LoadBalancerRequestId(String.format("%s-%s-%s", singularityRequest.getId(), deploy.getId(), System.currentTimeMillis()), LoadBalancerRequestType.REMOVE, Optional.absent());
+      Collection<UpstreamInfo> upstreamsInLoadBalancerForService = getLoadBalancerUpstreamsForService(singularityRequest.getId(), loadBalancerUpstreamGroup);
```
ssalinas: are we taking into account the `SingularityDeploy.loadBalancerServiceIdOverride` here? That essentially lets something under one Singularity request id be under a different baragon service id.

e.g. MyService-web is normal and has group default, then MyService-userweb has `loadBalancerServiceIdOverride` set to MyService-web and `loadBalancerUpstreamGroup` set to `user` so that both share the same baragon service with different groups

ssalinas:
🚢

PR #1887, opened 2018-12-12
Remove all switch user bits from docker wrapper

ssalinas:
Opening docker to non-root users is a recipe for disaster anyways. This will now assume the Singularity executor is running as root in docker instances such that docker usage can be more effectively limited to the root user.

@kenbreeman


pschoenfelder:
🚢

kenbreeman:
:ship:

PR #1888, opened 2018-12-13
Report uploader/downloader metrics by writing to a file.

baconmania:


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/config/SingularityS3UploaderConfiguration.java
```diff
@@ -76,6 +76,10 @@
   @JsonProperty
   private List<SingularityS3UploaderContentHeaders> s3ContentHeaders = new ArrayList<>();
 
+  @NotNull
+  @JsonProperty
+  private String metricsFilePath = "/var/run/singularity/s3uploader-metrics.out";
```
ssalinas: random thought, if we don't want everyone to have to have this enabled, maybe this could be an optional and we only start the poller if it's set? (Or another boolean for enabled?)

ssalinas:
🚢

PR #1894, opened 2019-01-03
Add trace logging for header passthrough authenticator

ssalinas:
Will help us track down what else is still using it

cc @pschoenfelder @baconmania


pschoenfelder:
🚢

PR #1899, opened 2019-01-08
Add missing enum

pschoenfelder:


SingularityBase/src/main/java/com/hubspot/mesos/protos/MesosTaskStatusReason.java
```diff
@@ -20,14 +20,10 @@
   REASON_MASTER_DISCONNECTED,
   REASON_RECONCILIATION,
   REASON_RESOURCES_UNKNOWN,
-  REASON_SLAVE_DISCONNECTED,
-  REASON_SLAVE_REMOVED,
-  REASON_SLAVE_REMOVED_BY_OPERATOR,
-  REASON_SLAVE_RESTARTED,
-  REASON_SLAVE_UNKNOWN,
```
ssalinas: only comment on this. Do we know what mesos version removed the REASON_SLAVE ones? If it's in the 1.x.x family it's possible someone is still using Singularity with a version that has them. There is no harm in keeping extras in this enum, only in missing them

ssalinas:
🚢

PR #1906, opened 2019-02-01
Reduce S3Uploader memory usage during directory scan

ssalinas:
Our `JavaUtils. iterable` function was previously taking a DirectoryStream -> Lists.newArrayList(stream.iterator()). This caused all paths in a directory to be in memory at once. For particularly large directories this was causing ooms in the uploader. Files.walk does this in more of a streaming fashion and should save us from loading the entire directory tree in memory before iterating over it


ssalinas:
Confirmed fix in our QA env. uploader that previously grew to ~1600MB is now topping out at 13MB of heap

pschoenfelder:
🚢

PR #1907, opened 2019-02-14
On demand requests with instances shouldn't trigger task lag over instance count

sjeropkipruto:
Currently, if you submit pending tasks in quick succession on an on demand request that has an instance count limit, you can end up with tasks that will wait for the first to complete before launching. We shouldn't page for task lag in these cases. 

[We previously separated the late tasks from on-demand requests from the rest of the late tasks](https://github.com/HubSpot/Singularity/pull/1881) and configured different alerting rules for them. In this PR, we ensure that we count the on-demand late tasks are within the task limit only so that on-demand late tasks blocked from launching because of instance count limit won't send alerts.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
@@ -200,7 +200,7 @@ public class SingularityRequest {
     return scheduleTimeZone;
   }
 
-  @Schema(nullable = true, description = "A count limit of tasks to run for on-demand requests")
+  @Schema(nullable = true, description = "A count of tasks to run for long-running requests or the limit on the number of concurrent tasks for on-demand requests")
   public Optional<Integer> getInstances() {
     return instances;
   }
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
@@ -200,7 +200,7 @@ public class SingularityRequest {
     return scheduleTimeZone;
   }
 
-  @Schema(nullable = true, description = "A count of tasks to run for long-running requests")
+  @Schema(nullable = true, description = "A count limit of tasks to run for on-demand requests")
   public Optional<Integer> getInstances() {
     return instances;
   }
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
@@ -317,10 +317,13 @@ public class StateManager extends CuratorManager {
   private List<SingularityPendingTaskId> getOnDemandLateTasks (List<SingularityPendingTaskId> maybeOnDemandLateTasks) {
     List<SingularityPendingTaskId> onDemandLateTasks = new ArrayList<>();
     for (SingularityPendingTaskId maybeOnDemandLateTask : maybeOnDemandLateTasks) {
-      final Optional<Integer> maybeInstancesLimit = requestManager.getRequest(maybeOnDemandLateTask.getRequestId()).get().getRequest().getInstances();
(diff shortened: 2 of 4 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
```diff
@@ -200,7 +200,7 @@ public String getId() {
     return scheduleTimeZone;
   }
 
-  @Schema(nullable = true, description = "A count of tasks to run for long-running requests")
+  @Schema(nullable = true, description = "A count limit of tasks to run for on-demand requests")
```
ssalinas: this should be an addition, not a replacement. It functions as both, depending on the context of the request type. eg. `A count of tasks to run for long-running requests or the limit on the number of concurrent tasks for on-demand requests`

SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
```diff
@@ -312,6 +313,20 @@ private boolean requestTypeIsOnDemand(SingularityPendingTaskId taskId) {
     return false;
   }
 
+  private List<SingularityPendingTaskId> getOnDemandLateTasks (List<SingularityPendingTaskId> maybeOnDemandLateTasks) {
+    List<SingularityPendingTaskId> onDemandLateTasks = new ArrayList<>();
+    for (SingularityPendingTaskId maybeOnDemandLateTask : maybeOnDemandLateTasks) {
+      String requestId = requestManager.getRequest(maybeOnDemandLateTask.getRequestId()).get().getRequest().getId();
```
ssalinas: we don't need to fetch this here, requestId is already on the SingularityPendingTaskId, and you are fetching the SingularityRequest on the next line down

SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
```diff
@@ -312,6 +313,20 @@ private boolean requestTypeIsOnDemand(SingularityPendingTaskId taskId) {
...
+      Optional<Integer> maybeInstancesLimit = requestManager.getRequest(maybeOnDemandLateTask.getRequestId()).get().getRequest().getInstances();
```
ssalinas: since `requestManager.getRequest` returns an Optional, we should have a check for isPresent here to avoid throwing an exception up the stack. e.g.

Optional<SingularityRequest> maybeRequest = requestManager.getRequest(...)
if (maybeRequest.isPresent) { continue to do stuff }

SingularityService/src/test/java/com/hubspot/singularity/data/StateManagerTest.java
```diff
@@ -74,4 +77,48 @@ public void itDoesntFlagPendingRequestsForUnderOrOverProvisioning() {
...
   }
+
+  @Test
+  public void itDoesntCountInstancesOverLimitInOnDemandLateTasks() {
+    initOnDemandRequest();
+    SingularityRequest request = requestResource.getRequest(requestId, singularityUser).getRequest();
+    requestResource.postRequest(request.toBuilder().setInstances(Optional.of(2)).build(), singularityUser);
+    initFirstDeploy();
+
+    configuration.setDeltaAfterWhichTasksAreLateMillis(TimeUnit.MILLISECONDS.toMillis(10));
```
ssalinas: shouldn't need MILLISECONDS.toMillis, can just use the number. Also, why not set to `0`? Outcome otherwise may be different depending on how fast the machine running the test is

SingularityService/src/main/java/com/hubspot/singularity/data/StateManager.java
sjeropkipruto: Got the `maybeOnDemandLateTasks` from `lateTasksPartitionedByOnDemand` which calls `requestTypeIsOnDemand` that had already checked that the request is optional...checking again here will be repetitive

ssalinas:
🚢

PR #1908, opened 2019-02-15
Cleaner failover + update dependencies

ssalinas:
This consolidates our many Managed classes and scheduled executor usages so we have clear control over the order in which they shut down (and make sure they all actually do...). This should help keep state more consistent on singularity failover. Ideally I'll add in the ability for Singulairty to not fail over for loss of mesos connection, and instead try to reconnect


ssalinas:
notes:
- NP_NONNULL_PARAM_VIOLATION came with the new findbugs version. Wasn't going to fix all occurrences as part of this PR as many were pretty far reaching or covered by logic a few lines above (which findbugs does not account for). Suppressed for now
- metrics-jmx split off from the normal dropwizard package and is now a new dep
- abort emails should have a full stack trace now
- leader only pollers are auto-discovered. As long as they are bound in guice, the managed class will start/stop them

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -403,14 +424,25 @@ public SingularityPendingRequestParent scheduleImmediately(SingularityUser user,
 
     checkConflict(requestWithState.getState() != RequestState.PAUSED, "Request %s is paused. Unable to run now (it must be manually unpaused first)", requestWithState.getRequest().getId());
 
+    // Check these to avoid unnecessary calls to taskManager
+    Integer activeTasks = null;
+    Integer pendingTasks = null;
```
baconmania: I think we want to default these to `0` instead of `null` to avoid NPE's in `checkRunNowRequest()`.

ssalinas: Yeah I can do that 👍

baconmania:
:ship:

baconmania:
:ship:

PR #1909, opened 2019-02-27
Skip fuser/lsof check when uploader is marked as immediate

ssalinas:
One of the slower/expensive parts of the uploader right now when there are many files is checking if the files are still in use. The `immediate` flag is used to signify that the file is ready for upload and done being written to. If this flag is set, skip the lsof/fuser check. SingularityExecutor uses this flag on clean shutdown, so this should alleviate the need for the check in the majority of cases


pschoenfelder:
🚢

PR #1910, opened 2019-03-13
Do the entire deploy history persist under the request lock

ssalinas:
We were grabbing the `Map<String, SingularityRequestDeployState>` before grabbing the request lock. So if the deploy checker ran at the right time right before this poller, the deploy state would be out of date. This changes the poller to loop over requests, holding the request lock for each during teh entire persist process


pschoenfelder:
🚢

PR #1911, opened 2019-03-13
Fix non-paginated fetch of s3 logs for task

ssalinas:
We weren't checking if there were more results beyond the first page. Previously we never had this many logs, but now that we've added thread dumps, there are quite a few more than previously per task


pschoenfelder:
🚢

PR #1912, opened 2019-03-13
Also fetch the original file for log snippets in email

ssalinas:
We currently have a race between the executor moving the log to tail_of_finished_... and the emailer fetching tail_of. Make the emailer check the original log first, if the path contains `tail_of_finished_`


pschoenfelder:
🚢

PR #1913, opened 2019-03-14
Clarify logging in offer scheduler

ssalinas:
- Don't log all task IDs if there are more than 30, hard to even read the line
- Also log the cpus already committed to an offer when considering it overloaded


pschoenfelder:
🚢

PR #1916, opened 2019-03-18
Build on JDK 8 internally

stevegutz:
This will stop the build from breaking internally on Blazar when I switch things to JDK 11 tomorrow.

The failures should be fixable by upgrades to things like findbugs, jacoco, and mockito.

@ssalinas @jhaber @kmclarnon


ssalinas:
🚢

PR #1918, opened 2019-03-22
Use basepom 25.3 and build on JDK 11 internally

stevegutz:
@ssalinas @jhaber @kmclarnon


stevegutz:
It looks like mockito is having trouble on JDK 11 and I'll have to upgrade it. Sorry for the premature PR

ssalinas:
closing in favor of #1948

PR #1920, opened 2019-04-16
Add ability to do sns-based updates instead of webhooks

ssalinas:
This is the first in a series of "let's be nicer to zookeeper" PRs. This adds the ability for Singularity to produce messages to SNS instead of backing webhook queues with zookeeper. This should reduce calls to zk by (task updates/second * num subscribers)

TODOs:
- [x] trace error handling back up for errors thrown in SingularityEventListener and make sure an SNS outage wouldn't cause issues on any critical path


SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
```diff
@@ -572,7 +573,10 @@ public SingularityCreateResult saveTaskHistoryUpdate(SingularityTaskHistoryUpdat
 
   @Timed
   public SingularityCreateResult saveTaskHistoryUpdate(SingularityTaskHistoryUpdate taskHistoryUpdate, boolean overwriteExisting) {
-    singularityEventListener.taskHistoryUpdateEvent(taskHistoryUpdate);
+    Optional<SingularityTask> task = getTask(taskHistoryUpdate.getTaskId());
```
ssalinas: This is the only line I was hesitant about in the PR, but I still think it's a new improvement. With the previous version, we would be doing a write to zk here. Now we are doing a (mostly likely cached) read and a write to SNS/zk. The SNS update falls back to a zk write in case sns is down, so nothing should block here.

SingularityService/src/main/java/com/hubspot/singularity/hooks/SingularityWebhookSender.java
```diff
@@ -60,36 +61,33 @@ public SingularityWebhookSender(SingularityConfiguration configuration,
...
-        case REQUEST:
-          requestUpdates += checkRequestUpdates(webhook, webhookFutures);
-          break;
-        case DEPLOY:
-          deployUpdates += checkDeployUpdates(webhook, webhookFutures);
-          break;
-        default:
-          break;
+    for (WebhookType type : WebhookType.values()) {
+      for (SingularityWebhook webhook : webhookManager.getActiveWebhooksByType(type)) {
```
ssalinas: this is a switch to use the cached version of this data since it can be queried frequently

pschoenfelder:
🚢

PR #1921, opened 2019-04-17
Immediately persist items to history

ssalinas:
Another in the 'be nice to zookeeper' series. This will add hooks to immediately persist (or check for purge) on objects that are no longer needed in ZK to reduce/stabilize total node count. The immediate persist should always be async to not interfere with critical scheduling actions

Remaining TODOs here:
- [x] replicate for request persister
- [x] replicate for deploy persister
- [x] Check if any modifications can be made to blended history fetcher with the assumption that 'all not active things are already in sql'
- [x] Task history persister needs to use the request lock


SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequestParent.java
```diff
@@ -120,16 +114,6 @@ public SingularityRequest getRequest() {
     return taskIds;
   }
 
-  @Schema(description = "The most recent historical update to this request", nullable = true)
```
baconmania: Are we concerned about usages of these in the wild?

baconmania:
🚢

PR #1922, opened 2019-04-22
The zoo is under new management (zk cleanup)

ssalinas:
Be much nicer to zookeeper by avoiding possible large payloads, specifically:

- Namespace pending tasks by request id so that the list children call for this does not get to large when there are thousands of pending tasks
  - Avoid the list + filter call, which required fetching all pending task ids to find the ones for a particular request
- Remove usage of the /tasks/active node entirely since we have this data elsewhere
- Namespaces last active task statuses nodes by request id and use this data to power the active tasks list instead
- clean up nodes from old features that are no longer used

Still TODO on this PR:
- [x] automatic cleanup of request groups when a request is deleted
- [x] revisit caching to determine if any additional data can be leader cached
- [x] more aggressive pruning of inactive slaves data

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
@@ -1,5 +1,6 @@
 package com.hubspot.singularity.data;
 
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 
@@ -31,6 +32,7 @@ public abstract class AbstractMachineManager<T extends SingularityMachineAbstrac
   private final Transcoder<T> transcoder;
   private final Transcoder<SingularityMachineStateHistoryUpdate> historyTranscoder;
   private final Transcoder<SingularityExpiringMachineState> expiringMachineStateTranscoder;
+  private final int maxHistoryEntries;
 
   public AbstractMachineManager(CuratorFramework curator, SingularityConfiguration configuration, MetricRegistry metricRegistry, Transcoder<T> transcoder,
       Transcoder<SingularityMachineStateHistoryUpdate> historyTranscoder, Transcoder<SingularityExpiringMachineState> expiringMachineStateTranscoder) {
@@ -39,6 +41,7 @@ public abstract class AbstractMachineManager<T extends SingularityMachineAbstrac
     this.transcoder = transcoder;
     this.historyTranscoder = historyTranscoder;
     this.expiringMachineStateTranscoder = expiringMachineStateTranscoder;
+    this.maxHistoryEntries = configuration.getMaxMachineHistoryEntries();
   }
 
   protected abstract String getRoot();
@@ -206,6 +209,16 @@ public abstract class AbstractMachineManager<T extends SingularityMachineAbstrac
     final String historyChildPath = String.format("%s-%s", historyUpdate.getState().name(), historyUpdate.getTimestamp());
 
     return ZKPaths.makePath(getHistoryPath(historyUpdate.getObjectId()), historyChildPath);
+  }
+
+  public void clearOldHistory(String machineId) {
+    List<SingularityMachineStateHistoryUpdate> histories = getHistory(machineId);
+    histories.sort(Comparator.comparingLong(SingularityMachineStateHistoryUpdate::getTimestamp));
+    histories.stream()
+        .skip(maxHistoryEntries)
+        .forEach((history) -> {
+          delete(getHistoryUpdatePath(history));
+        });
   }
 
   private SingularityCreateResult saveHistoryUpdate(SingularityMachineStateHistoryUpdate historyUpdate) {
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
@@ -224,7 +224,7 @@ public abstract class AbstractMachineManager<T extends SingularityMachineAbstrac
 
   public void clearOldHistory(String machineId) {
     List<SingularityMachineStateHistoryUpdate> histories = getHistory(machineId);
-    histories.sort(Comparator.comparingLong(SingularityMachineStateHistoryUpdate::getTimestamp));
+    histories.sort(Comparator.comparingLong(SingularityMachineStateHistoryUpdate::getTimestamp).reversed());
     histories.stream()
         .skip(maxHistoryEntries)
         .forEach((history) -> {
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
@@ -120,6 +120,17 @@ public abstract class AbstractMachineManager<T extends SingularityMachineAbstrac
   }
 
   public Optional<T> getObject(String objectId) {
+    Optional<T> maybeCached = getObjectFromLeaderCache(objectId);
+    if(!maybeCached.isPresent()) {
+      return getObjectNoCache(objectId);
+    } else {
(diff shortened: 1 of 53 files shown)


ssalinas:
We weren't using the correct code paths to hit the leader cache for slave + rack data, causing lots of unneeded zk calls for those. Fixed in this PR now

SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
```diff
@@ -52,9 +55,16 @@ private String getHistoryPath(String objectId) {
   }
 
   public List<T> getObjects() {
-    return getObjects(getRoot());
+    List<T> fromCache = getObjectsFromLeaderCache();
+    if (fromCache != null) {
```
baconmania: There look to be at least two places where this'll never be null: `SingularityLeaderCache#getRacks()` and `SingularityLeaderCache#getSlaves()` do we want to check this for emptiness as well as nullity, and fall back to `getObjectsNoCache()` in either case?

ssalinas: ah, good point, maybe I need the `isActive` check in the leader cache to return null instead

SingularityService/src/main/java/com/hubspot/singularity/data/RackManager.java
```diff
@@ -39,33 +43,41 @@ protected String getRoot() {
...
-    leaderCache.cacheRacks(getObjects());
+    leaderCache.cacheRacks(getObjectsNoCache(getRoot()));
   }
 
-  public Optional<SingularityRack> getRack(String rackName) {
+  @Override
+  public Optional<SingularityRack> getObjectFromLeaderCache(String rackId) {
     if (leaderCache.active()) {
-      return leaderCache.getRack(rackName);
+      return leaderCache.getRack(rackId);
```
baconmania: We're using both `rackName` (in `SingularityLeaderCache`) and `rackId`. We should pick one for consistency

SingularityService/src/main/java/com/hubspot/singularity/data/zkmigrations/NamespaceActiveTasksMigration.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Override
+  public void applyMigration() {
+    try {
+      if (curatorFramework.checkExists().forPath(ACTIVE_TASKS_ROOT) != null) {
+        List<String> currentActive = curatorFramework.getChildren().forPath(ACTIVE_STATUSES_ROOT);
+        for (String taskIdString : currentActive) {
+          SingularityTaskId taskId = SingularityTaskId.valueOf(taskIdString);
+          String oldPath = ZKPaths.makePath(ACTIVE_STATUSES_ROOT, taskIdString);
```
baconmania: We're running this code based on whether `ACTIVE_TASKS_ROOT` exists in ZK, but we're never actually reading from that path during this migration. Just double-checking, is that intended?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerImpl.java
```diff
@@ -153,6 +153,9 @@ public void subscribed(Subscribed subscribed) {
         heartbeatIntervalSeconds = Optional.of(advertisedHeartbeatIntervalSeconds);
       }
 
+      // Should be called before activation of leader cache or cache could be left empty
+      startup.checkMigrations();
```
baconmania: Do we expect that this wouldn't have run at startup?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityHealthcheckAsyncHandler.java
```diff
@@ -36,32 +34,25 @@ public SingularityHealthcheckAsyncHandler(SingularityExceptionNotifier exception
...
-  @Override
-  public Response onCompleted(Response response) throws Exception {
-    Optional<String> responseBody = Optional.absent();
-
-    if (response.hasResponseBody()) {
-      responseBody = Optional.of(response.getResponseBodyExcerpt(maxHealthcheckResponseBodyBytes));
-    }
-
-    saveResult(Optional.of(response.getStatusCode()), responseBody, Optional.<String> absent(), Optional.<Throwable>absent());
+  public void setHealthcehckUri(String healthcheckUri) {
```
baconmania: nit: spelling

pom.xml
```diff
@@ -35,7 +35,7 @@
     <basepom.release.profiles>oss-release</basepom.release.profiles>
     <dep.classmate.version>1.3.1</dep.classmate.version>
     <dep.commons-lang3.version>3.7</dep.commons-lang3.version>
-    <dep.curator.version>2.12.0</dep.curator.version>
+    <dep.curator.version>4.2.0</dep.curator.version>
```
baconmania: Sweet glory

SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
```diff
@@ -203,20 +222,33 @@ private String getHistoryUpdatePath(SingularityMachineStateHistoryUpdate history
     return ZKPaths.makePath(getHistoryPath(historyUpdate.getObjectId()), historyChildPath);
   }
 
+  public void clearOldHistory(String machineId) {
+    List<SingularityMachineStateHistoryUpdate> histories = getHistory(machineId);
+    histories.sort(Comparator.comparingLong(SingularityMachineStateHistoryUpdate::getTimestamp));
```
baconmania: I think we want to reverse this sort, or we'll keep the oldest `n` entries instead of the newest.

SingularityService/src/main/java/com/hubspot/singularity/data/RackManager.java
ssalinas: consolidating on rackId 👍

SingularityService/src/main/java/com/hubspot/singularity/data/zkmigrations/NamespaceActiveTasksMigration.java
ssalinas: We remove it as the last step of the migration which is why it's my initial check. There isn't actually any data in those nodes, they are only an index we call getChildren on. The /tasks/statuses is where the actual data is and what I am namespacing. However, realizing that if this died half way though, some of the children on the next run would be request ids instead. Going to add a fix so this will be able to run on a half-migrated dataset as well

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerImpl.java
ssalinas: It only runs when we take leadership, not as a Managed class. Moved it here since we need to migrate before filling leader cache, since leader cache is using the newer taskManager at this point

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityHealthcheckAsyncHandler.java
ssalinas: 👍 fixed

pom.xml
ssalinas: 🎉

SingularityService/src/main/java/com/hubspot/singularity/data/AbstractMachineManager.java
ssalinas: Reversing the sort. Though I'm realizing we don't even really expose this data anywhere... Maybe don't even need it

baconmania:
:ship:

PR #1924, opened 2019-04-24
Fail unhealthy deploys instead of marking as overdue

pschoenfelder:


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -611,6 +611,11 @@ private SingularityDeployResult checkDeployProgress(final SingularityRequest req
       return checkCanMoveToNextDeployStep(request, deploy, pendingDeploy, updatePendingDeployRequest);
     }
 
+    final DeployHealth deployHealth = deployHealthHelper.getDeployHealth(request, deploy, deployActiveTasks, true);
+    if (request.getRequestType().equals(RequestType.WORKER) && deployHealth.equals(DeployHealth.UNHEALTHY)) {
```
ssalinas: Can do == for enum comparison

ssalinas: Other thing, can we leave a note somehow, or maybe a mini method with a descriptive name so we know this line is here specifically to catch non-web health check failures?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -79,12 +79,12 @@ private boolean runHealthcheck() {
 
       try {
         Retryer<Boolean> retryer = RetryerBuilder.<Boolean>newBuilder()
-            .retryIfResult(bool -> !bool)
+            .retryIfResult(bool -> bool)
```
ssalinas: if the process is not alive though, don't we want to have this whole method return false (i.e. did not pass health check). This is the place for the enum/ternary return type I mentioned. We need to differentiate waiting vs process dead vs succeeded when deciding if we need to retry

ssalinas:
🚢

PR #1925, opened 2019-04-26
Always remove from LB, even if add in WAITING

ssalinas:
We do this already on all individual task checks, but cleanup takes a different path during deploys. A badly timed lb add that ends up stuck in waiting at the exact time a deploy finishes can lead to an additional task in baragon state


baconmania:
🚢

PR #1926, opened 2019-04-29
Also attempt to get TaskHistory from history manager for mail

ssalinas:
Side effect of the immediate persist changes. Mailer also needs to check history


baconmania:
🚢

PR #1928, opened 2019-04-29
MySQL task resource usage storage

ssalinas:
Last (for now) in the 'be nice to zookeeper' series. We store cluster, slave, request, and task level resource usage, currently all in zookeeper. This does a few things to make that more scalable:
- For request/slave we only ever need the current. Stop saving history, and make sure recent slave usages are also leader cached
- Task usages, if in zk, need to be namespaced by request id to avoid extra large list children calls of all task ids
- Task usages will be stored in mysql if it is configured, which will require a migration to be run to create the table
- A zk migration has been added to clean all current usage data. This data will be rewritten the next time the poller runs and is not required for scheduling
- task usage timestamps are now stored in millis instead of seconds
- `usage/slaves/{slaveId}/tasks/current` endpoint is removed

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskUsage.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskUsage.java
@@ -9,7 +9,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 public class SingularityTaskUsage {
 
   private final long memoryTotalBytes;
-  private final long timestamp; // seconds
+  private final long timestamp; // epoch millis
   private final double cpuSeconds;
   private final long diskTotalBytes;
   private final long cpusNrPeriods;
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskUsage.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskUsage.java
@@ -9,7 +9,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 public class SingularityTaskUsage {
 
   private final long memoryTotalBytes;
-  private final double timestamp; // seconds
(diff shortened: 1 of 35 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskUsage.java
```diff
@@ -9,7 +9,7 @@
 public class SingularityTaskUsage {
 
   private final long memoryTotalBytes;
-  private final double timestamp; // seconds
+  private final long timestamp; // seconds
```
baconmania: We should update the comment here

ssalinas: thanks, missed that

SingularityService/src/main/java/com/hubspot/singularity/data/usage/TaskUsageJDBI.java
```diff
@@ -0,0 +1,24 @@
+package com.hubspot.singularity.data.usage;
+
+import java.util.Date;
+import java.util.List;
+
+import com.hubspot.singularity.SingularityTaskUsage;
+
+public abstract class TaskUsageJDBI {
+  static final String FILEDS = "requestId, taskId, memoryTotalBytes, timestamp, cpuSeconds, diskTotalBytes, cpusNrPeriods, cpusNrThrottled, cpusThrottledTimeSecs";
+  static final String FILED_VALUES = ":requestId, :taskId, :memoryTotalBytes, :timestamp, :cpuSeconds, :diskTotalBytes, :cpusNrPeriods, :cpusNrThrottled, :cpusThrottledTimeSecs";
```
baconmania: nit: spelling

baconmania:
🚢

PR #1929, opened 2019-04-30
Put a time limit on the uploader check so we don't get stuck

ssalinas:
We've seen slow leaks of uploader files, likely from the `chekcUploads` method getting stuck. IF this gets stuck and subsequent upload checks don't run, then files do not get cleaned up. Immediate uploads are still triggered fine, which is the majority of uploads from the SingularityExecutor.

To start, this just adds a timeout on the regular uploader check and some additional logging so we can know which future may be stuck


baconmania:
:ship:

PR #1931, opened 2019-04-30
Change persist strategy

ssalinas:
Immediate persist was causing more problems than it solved. Backing that out in favor of just running the poller more frequently


baconmania:
🚢

PR #1932, opened 2019-05-01
Refactor calling of offer evaluation

ssalinas:
While looking into some odd offer issues, I noticed that our invocations of checking offers differed slightly between the SchedulerImpl and the SchedulerPoller. This does two things:
- Consolidates all offer evaluation logic for new and cached offers in the SingularityMesosOfferScheduler class, so the same logic is called in all places
- Moves the drainPendingQueue operation to only the scheduler poller, since there is no need to call it on every offer run (which can be multiple times a second), making offer evaluation a bit more efficient
- Add logging on all possible offers that were used to launch a task so we have a better record of all of them


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -134,32 +162,163 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
...
   }
 
-  public Collection<SingularityOfferHolder> checkOffers(final Collection<Offer> offers) {
-    for (SingularityPendingTaskId taskId : taskManager.getPendingTasksMarkedForDeletion()) {
-      lock.runWithRequestLock(() -> taskManager.deletePendingTask(taskId), taskId.getRequestId(), String.format("%s#%s", getClass().getSimpleName(), "checkOffers -> pendingTaskDeletes"));
+  public void resourceOffers(List<Offer> uncached) {
+    final long start = System.currentTimeMillis();
+    LOG.info("Received {} offer(s)", uncached.size());
+    scheduler.checkForDecomissions();
+    boolean delclineImmediately = false;
```
baconmania: nit: old typo

ssalinas: fixed

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -134,32 +162,163 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
...
+    Map<String, CachedOffer> cachedOffers = new HashMap<>();
+    for (CachedOffer cachedOffer : cachedOfferList) {
+      if (isValidOffer(cachedOffer.getOffer())) {
+        cachedOffers.put(cachedOffer.getOfferId(), cachedOffer);
+        offersToCheck.add(cachedOffer.getOffer());
+      } else if (cachedOffer.getOffer().getId() != null && cachedOffer.getOffer().getId().getValue() != null) {
+        mesosSchedulerClient.decline(Collections.singletonList(cachedOffer.getOffer().getId()));
+        offerCache.rescindOffer(cachedOffer.getOffer().getId());
+      } else {
+        LOG.warn("Offer {} was not valid, but we can't decline it because we have no offer ID!");
```
baconmania: We'll want to pass `cachedOffer` to this log statement

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -134,32 +162,163 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
...
+        uncached.size() - acceptedOffers.size());
+  }
 
+  private void checkOfferAndSlave(Offer offer, List<Offer> offersToCheck) {
+    if (!isValidOffer(offer)) {
+      offersToCheck.remove(offer);
+      if (offer.getId() != null && offer.getId().getValue() != null) {
+        mesosSchedulerClient.decline(Collections.singletonList(offer.getId()));
+      } else {
+        LOG.warn("Offer {} was not valid, but we can't decline it because we have no offer ID!");
```
baconmania: Same here, we'll want to pass `offer` into the log

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -134,32 +162,163 @@ public SingularityMesosOfferScheduler(MesosConfiguration mesosConfiguration,
...
+          });
+        }
+      }
+
+      LOG.info("{} remaining offers not accounted for in offer check", cachedOffers.size());
+      cachedOffers.values().forEach(offerCache::returnOffer);
+    } catch (Throwable t) {
+      LOG.error("Received fatal error while handling offers - will decline all available offers", t);
+
+      mesosSchedulerClient.decline(offersToCheck.stream()
```
baconmania: Since `offersToCheck` is created via `Collections.synchronizedList()`, iterations over it need to be inside a `synchronized (offersToCheck) { }` block

baconmania:
🚢

PR #1933, opened 2019-05-01
More explicit choice of canSkipZk flag

ssalinas:
Another piece related to backing out immediate persist. I added this as an efficiency boost for zk since task history was immediately in sql most times. Now that it has to wait for the persister to run, it can appear like tasks are missing sometimes. I'm still leaving this as an optimization, with option to override, for many endpoints that are only ever used in task search. Those used by the UI will default to checking zookeeper as well

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
@@ -164,14 +164,15 @@ public class HistoryResource extends AbstractHistoryResource {
       @Parameter(required = true, description = "Request ID for deploy") @PathParam("requestId") String requestId,
       @Parameter(required = true, description = "Deploy ID") @PathParam("deployId") String deployId,
       @Parameter(description = "Maximum number of items to return") @QueryParam("count") Integer count,
-      @Parameter(description = "Which page of items to view") @QueryParam("page") Integer page) {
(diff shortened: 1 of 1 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/HistoryResource.java
```diff
@@ -164,14 +164,15 @@ public SingularityDeployHistory getDeploy(
       @Parameter(required = true, description = "Request ID for deploy") @PathParam("requestId") String requestId,
       @Parameter(required = true, description = "Deploy ID") @PathParam("deployId") String deployId,
       @Parameter(description = "Maximum number of items to return") @QueryParam("count") Integer count,
-      @Parameter(description = "Which page of items to view") @QueryParam("page") Integer page) {
+      @Parameter(description = "Which page of items to view") @QueryParam("page") Integer page,
+      @Parameter(description = "Skip checking zookeeper, items that have not been persisted yet may not appear") @QueryParam("skipZk") Boolean skipZk) {
```
jhaber: Maybe add `@DefaultValue("true")`? Then you can switch the type to a primitive `boolean` and remove the ternary below

ssalinas: Good call, I always forget that annotation exists

pschoenfelder:
🚢

PR #1935, opened 2019-05-08
Run a configured shell command before killing a task for thread overuse.

baconmania:


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/SingularityExecutorThreadChecker.java
```diff
@@ -101,13 +111,51 @@ private void checkThreads() {
...
+          );
+
+          SingularityExecutorShellCommandUpdater updater = new SingularityExecutorShellCommandUpdater(
+              objectMapper, shellRequest, taskProcess.getTask()
+          );
+
+          SingularityExecutorShellCommandRunner shellRunner = new SingularityExecutorShellCommandRunner(shellRequest, configuration, taskProcess.getTask(),
+              taskProcess, monitor.getShellCommandExecutorServiceForTask(taskProcess.getTask().getTaskId()), updater);
+
+          Futures.addCallback(shellRunner.start(), new FutureCallback<Integer>() {
```
ssalinas: curious, any reason for Futures vs CompletableFuture here?

baconmania: Mostly just staying consistent with shell commands kicked off elsewhere in the codebase, which all use the same `Futures.addCallback()` style here

ssalinas:
🚢

PR #1937, opened 2019-05-08
Shuffle tasks on hosts with overutilized memory resources

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -82,7 +82,9 @@
 
   private int maxConcurrentUsageCollections = 15;
 
-  private boolean shuffleTasksForOverloadedSlaves = false; // recommended 'true' when oversubscribing cpu for larger clusters
+  private boolean shuffleTasksForOverloadedSlaves = false; // recommended 'true' when oversubscribing resources for larger clusters
+
+  private double shuffleTasksWhenSlaveMemoryUtilizationPercentageExceeds = 0.82;
```
ssalinas: was there any math behind this value?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -111,68 +114,146 @@ public void runActionOnPoll() {
...
+      }
 
       for (TaskIdWithUsage taskIdWithUsage : possibleTasksToShuffle) {
         if (requestsWithShuffledTasks.contains(taskIdWithUsage.getTaskId().getRequestId())) {
           LOG.debug("Request {} already has a shuffling task, skipping", taskIdWithUsage.getTaskId().getRequestId());
           continue;
         }
-        if (cpuOverage <= 0 || shuffledTasksOnSlave > configuration.getMaxTasksToShufflePerHost() || currentShuffleCleanupsTotal >= configuration.getMaxTasksToShuffleTotal()) {
-          LOG.debug("Not shuffling any more tasks (overage: {}, shuffledOnHost: {}, totalShuffleCleanups: {})", cpuOverage, shuffledTasksOnSlave, currentShuffleCleanupsTotal);
+        if ((mostOverusedResource.overusage <= 0) || shuffledTasksOnSlave > configuration.getMaxTasksToShufflePerHost() || currentShuffleCleanupsTotal >= configuration.getMaxTasksToShuffleTotal()) {
```
ssalinas: `cpuOverage <= 0` worked as a condition for 'should this put us back in green?' because we updated it at the end of the loop. I don't currently see `mostOverusedResource.overusage` getting updated anywhere

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
baconmania: Sort of an arbitrary rule of thumb. Assuming we'd like to set a default "target" memory utilization of 85%, setting this config at 82% should give us enough time to shuffle tasks before actually hitting our target.

ssalinas:
🚢

ssalinas:
🚢

PR #1938, opened 2019-05-09
Make Slave/Rack Resource use proxy to leader

ssalinas:
Fixes an issue where leader cache can get out of sync with zk for decommissions


baconmania:
🚢

PR #1939, opened 2019-05-22
Add option to fetch a batch of requests

ssalinas:
For @gchomatas . Still need to test it out, but this should be generally what you're looking for to speed up these fetches


baconmania:
:ship'

baconmania:
🚢

PR #1940, opened 2019-05-22
Files from S3Downloader should be world readable, not read-writeable

ssalinas:
Currently these end up as `-rw-rw-rw`, changing this so only the owner has write
cc @kenbreeman


kenbreeman:
:ship:

baconmania:
🚢

PR #1942, opened 2019-05-22
Check assigned ports are available in SingularityExecutor

ssalinas:
Fail fast if some other process is using a port it isn't supposed to (as opposed to letting the process start and eventually throwing something like a BindException)

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorTaskBuilder.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorTaskBuilder.java
@@ -1,7 +1,7 @@
 package com.hubspot.singularity.executor.config;
 
 import java.io.IOException;
-import java.net.Socket;
+import java.net.ServerSocket;
 import java.nio.file.Path;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


baconmania:
🚢

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorTaskBuilder.java
```diff
@@ -83,6 +92,32 @@ public SingularityExecutorTask buildTask(String taskId, ExecutorDriver driver, T
...
+          }
+        }
+      }
+    }
+
+  }
+
+  private boolean isPortInUse(int port) {
+    try {
+      new Socket("127.0.0.1", port).close();
```
PaulFurtado: You should create a ServerSocket on `0.0.0.0` to attempt to bind the port, rather than attempting to connect to it. There are a lot of reasons the connection to the port could fail, and it could even block for a bit trying to connect which would cause this function to hang.

Additionally, there's no real guarantee in singularity that the app will bind to 127.0.0.1, it could just bind to the main IP of the host and you'd miss it, so binding 0.0.0.0 would ensure it's not listening to the port on any address.

ssalinas: good point, will switch to ServerSocket

baconmania:
🚢

PR #1943, opened 2019-05-22
Upstream validation

sjeropkipruto:
We never want upstreams duplicated. Adding upstream validation to ensure that we don't ever get in a situation where we routing traffic to the wrong service. Instead, the task should fail to start.

Related Baragon PRs: https://github.com/HubSpot/Baragon/pull/298, https://github.com/HubSpot/Baragon/pull/300


ssalinas:
🚢

PR #1944, opened 2019-05-22
Handle status updates from recovering agents appropriately

ssalinas:
This is a first pass to make sure we don't leave stale tasks running. If we already thought a task was gone, make it gone again


ssalinas:
Agent recovery updates are working fine now. Still need to fix one bug around cleanup with rebalance racks that gets triggered and starts an additional replacement task that immediately gets shut down

baconmania:
🚢

PR #1945, opened 2019-05-24
Bump to mesos 1.8.0

ssalinas:
No ground breaking changes in the recent versions on the framework side. The change with the most impact here will likely be the introduction of rapidjson in the mesos master. Jackson should handle reading the updated json just fine still.


baconmania:
🚢

PR #1946, opened 2019-05-24
Alternative way to specify auth for the mesos scheduler api

ssalinas:
Allow as a masterUri of user:host@hostname or as separate mesosUsername/mesosPassword config fields in the yaml. Will allow easier rotation of these parameters separately


baconmania:
🚢

PR #1948, opened 2019-05-30
Upgrade to basepom 25.4

jhaber:
I was hoping this would allow me to remove `.build-jdk8` however it seems like that will need to wait for https://github.com/spotbugs/spotbugs/issues/756

@ssalinas @baconmania @sjeropkipruto

### DIFF
--- a/EmbedSingularityExample/pom.xml
+++ b/EmbedSingularityExample/pom.xml
@@ -36,7 +36,3 @@
     </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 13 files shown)


jhaber:
It's already on basepom 25.3 and compatible with Maven 3.6.1 so there's no urgency around this upgrade

ssalinas:
Folding this into #1975 , going to find which places need temporary annotations so we don't lag behind the rest of the org in upgrading to java 11

PR #1951, opened 2019-06-06
Add builder methods to `SingularityRequestBuilder` & `SingularityDeployBuilder`

kdorosh:
* Adds `setRequestId` to `SingularityDeployBuilder `
* Adds `setRequestType`, and `setId` to `SingularityRequestBuilder`

These methods are useful for "cloning" other requests and deploys from template requests, removing a lot of unnecessary boilerplate.


ssalinas:
🚢

PR #1963, opened 2019-06-20
Also purge deploy and request history

ssalinas:
we currently don't ever remove anything from these tables. We should stop doing that


pschoenfelder:
🚢

pschoenfelder:
🚢

PR #1969, opened 2019-06-27
Calculate max task lag after excluding on demands with instance limit

ssalinas:
The on demands over their instance count are expected to be late, and should not contribute to this value


sjeropkipruto:
🚢

PR #1970, opened 2019-06-27
Proxy deploy cancellations to leader

pschoenfelder:


ssalinas:
Can we do this for the `updatePendingDeploy` method below this as well?

ssalinas:
🚢

PR #1972, opened 2019-07-05
More efficient active tasks call for executor cleanup

ssalinas:
With many slaves these calls can build up when they all run on a cron at once. The endpoint they use fetches full tasks rather than just ids, so it causes too many zk calls. This adds an endpoint to just grab the ids since that is all cleanup needs. The endpoint is essentially the same as the getTasksOnSlave one without the extra call to hydrate each task


baconmania:
:ship:

PR #1973, opened 2019-07-10
tweak cooldown thresholds and evaluation logic

ssalinas:
- There isn't a case where we want to evaluate cooldown on TASK_KILLED, skip if that's the status
- Check for the most recent failure being outside the time window wasn't correct
- Make cooldown clear a bit faster


sjeropkipruto:
🚢

PR #1974, opened 2019-07-15
Fix task history page size on refresh

ssalinas:
Any refreshing in the UI tends to break the user experience paging through lots of task history. Resetting them to the beginning. Pull in the current page + page size when doing this


sjeropkipruto:
🚢

PR #1975, opened 2019-07-16
junit5

ssalinas:
https://github.com/HubSpot/Singularity/pull/1965 is not working out as well as I had hoped. So, I'm going to pick apart the usable pieces I found along the way. The first is upgrading to junit 5, which cuts our build time in half and then some. This is generally due to setting up TestingServer/hk2/etc once per class instead of once per test method (Thanks to @BeforeAll now allowing us to still us dependency injection in a Before method)


pschoenfelder:
🚢

PR #1976, opened 2019-07-17
UI Performance Updates

ssalinas:
Nothing major in terms of backend work, but I found while doing the caching work in another PR that our UI uses some more expensive endpoints than needed for the active and scheduled tasks pages. We can make these much faster by omitting a few columns and instead requiring an extra click to load the expensive data.


pschoenfelder:
🚢

PR #1978, opened 2019-07-17
Fix deploy link on requests page + pending tasks table

ssalinas:
Was missing a property in the property filtering list


pschoenfelder:
🚢

baconmania:
🚢

PR #1979, opened 2019-07-18
Fix missing extension for additional logrotate file

ssalinas:
In cases where the file name is like `stdout` there will be no extension. the Optional.or(null) will throw an error in that case, but we need the null value to differentiate things properly in the template


tpetr:
:ship:

PR #1980, opened 2019-07-18
Add a base timeout to executor healthchecks + make configurable in yaml

ssalinas:
reverts functionality back to when we had the startup delay, but makes it configurable in executor yaml instead and continues polling at same interval instead of waiting first


jonathanwgoodwin:
🚢

PR #1984, opened 2019-07-19
bump blazar and other versions

ssalinas:
Some of these hadn't been touched in a while. Get them all up to date


pschoenfelder:
🚢

PR #1985, opened 2019-07-19
Fix api reference docs link

ssalinas:
Something weird with the gitbook here. It doesn't like if I give it a full url, but also seems to weirdly hang if I use the relative one. Navigating directly to api.html works though


baconmania:
🚢

PR #1986, opened 2019-07-19
tech debt cleanup

ssalinas:
Figured right after a release is a good time to do this since we have a new snapshot version to start working off of. Currently includes:

- guava -> java util Optional
- remove Throwables.propagate
- clean up various other deprecated methods


ssalinas:
Note for posterity, ObjectMapper was updated to NOT_ABSENT setting. GuavaModule had this behavior by default but jdk8 module does not

baconmania:
:ship:

PR #1987, opened 2019-08-02
Also check sql for task/directory

ssalinas:
We have some cases of task emails not being sent on finish. If the persister runs between the task finish and email send, we won't find the task/directory and won't send the email. This updates the smtp mailer to fall back to the sql data if present


sjeropkipruto:
🚢

PR #1988, opened 2019-08-09
Do not allow user to override STARTED_BY_USER variable

ssalinas:
This should cover all the basic cases I can think of where a user might try to override this. Still need to hunt down a solution for some more complicated edge cases


sjeropkipruto:
🚢

PR #1992, opened 2019-08-12
Remove extra NotNull annotations

ssalinas:
Had these on Optionals in some places, which works with guava but can make jackson confused with java.util.Optional


sjeropkipruto:
🚢

PR #1993, opened 2019-08-12
Revert AWS builder update for now

ssalinas:
Going to have to be more careful and deliberate about how we specify regions. Want to give that one a bit more thought on how to make it backwards compatible


sjeropkipruto:
🚢

PR #2000, opened 2019-08-15
Module + ObjectMapper cleanup

ssalinas:
- Remove a problematic metrics module that we aren't really using anymore anyways
- more object mapper flexibility


sjeropkipruto:
🚢

PR #2003, opened 2019-08-23
Task history ui updates for sort

ssalinas:
- Make the backend sort correctly when combining zk + sql data. Most times this is already sorted correctly, but other times it is not, depending on history persister times
- Make the sort more obvious (but still not currently editable) in the UI


ssalinas:
Realized after looking further that a bunch of our code sorts by updatedAt or assumes it is sorted as such. But the sql and default query args sort by startedAt. Fixing that so everything is consistent on updatedAt

sjeropkipruto:
🚢

PR #2007, opened 2019-09-04
More efficient use of state data

ssalinas:
Right now when we restart/redeploy singulairty, we slam zk with a ton of requests from startup as well as regenerating state on every instance at once. This updates the state manager to prefer pulling the existing state from zk. The state will be updated async by the leading Singularity instance. This makes the endpoint respond much faster as well as lessening the load on zk when several instances try to generate state at once


baconmania:
🚢

PR #2011, opened 2019-09-05
Retry more TASK_LOST cases on deploy

ssalinas:
Most of these are Singularity/mesos issues, not an issue with the deployed task itself


baconmania:
🚢

PR #2013, opened 2019-09-05
Validate number of ports in run now resource overrides

ssalinas:
We can get in a crash loop state with an error in the scheduler if a run now override specifies less than the ports in the deploy. This is due to the fact that port mappings for docker can be specified on the deploy that then do not have an available port given the new resource overrides


pschoenfelder:
🚢

PR #2014, opened 2019-09-11
Don't immediately clean all history items, let the history purger do it

ssalinas:
Often we have cases where a request gets immediately recreated as a new type. We don't want to lose the history in these cases. Let the history purger do this later on instead


pschoenfelder:
🚢

PR #2016, opened 2019-09-13
Ensure delayed oneoff tasks get rescheduled on new deploy

ssalinas:
Right now we clean all pending tasks not belonging to the active deploy. However, for a oneoff that a user may have scheduled for a few hours in the future. They expect that task to still run. This recreates the task under the updated active deploy id

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
@@ -171,7 +171,9 @@ public class TaskResource extends AbstractLeaderAwareResource {
   public List<SingularityPendingTaskId> getScheduledTaskIds(
       @Parameter(hidden = true) @Auth SingularityUser user,
       @Parameter(description = "Use the cached version of this data to limit expensive api calls") @QueryParam("useWebCache") Boolean useWebCache) {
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/TaskResource.java
```diff
@@ -171,7 +171,9 @@ public TaskResource(TaskRequestManager taskRequestManager, TaskManager taskManag
   public List<SingularityPendingTaskId> getScheduledTaskIds(
       @Parameter(hidden = true) @Auth SingularityUser user,
       @Parameter(description = "Use the cached version of this data to limit expensive api calls") @QueryParam("useWebCache") Boolean useWebCache) {
-    return authorizationHelper.filterByAuthorizedRequests(user, taskManager.getPendingTaskIds(useWebCache(useWebCache)), SingularityTransformHelpers.PENDING_TASK_ID_TO_REQUEST_ID, SingularityAuthorizationScope.READ);
+    List<SingularityPendingTaskId> taskIds = authorizationHelper.filterByAuthorizedRequests(user, taskManager.getPendingTaskIds(useWebCache(useWebCache)), SingularityTransformHelpers.PENDING_TASK_ID_TO_REQUEST_ID, SingularityAuthorizationScope.READ);
+    Collections.sort(taskIds);
```
mjball: is `taskIds` guaranteed to be mutable? It might be safer to return a sorted copy. 🤷‍♂

pschoenfelder:
🚢

PR #2017, opened 2019-09-18
Add an hour key to the format spec

axiak:
this will allow us to put files in `/uuuu/MM/dd/HH/` if we so choose


ssalinas:
Replaced by https://github.com/HubSpot/Singularity/pull/2018

PR #2018, opened 2019-09-19
Add hour to s3 key

ssalinas:
@axiak , I added one more commit to finish off adding `%H`. Going to replace #2017 with this one


baconmania:
🚢

PR #2019, opened 2019-09-20
Remount RequestDetailPage when jumping between requests

benheng:
Users can jump between requests using the global search. React decides not to remount the RequestDetailPage component because only the route param changes.

Applying a `key` based on the `requestId` will force React to remount the component. Another way to fix this would be in the component itself, but I think it's better done here.


ssalinas:
🚢

PR #2020, opened 2019-10-01
Update ProcessUtils to work with java11

ssalinas:
Looks like in newer java versions (or at least in openjdk11) UNIXProcess is gone and replaced with ProcessImpl. There is only a public method to get the pid as of java 9, so we'll have to wait until we are fully upgraded to start using that. For now continue using reflections against either class


pschoenfelder:
🚢

PR #2028, opened 2019-10-25
Rework scheduler connect + reconnect logic

ssalinas:
Goals of this PR:
- Don't require an abort/restart when mesos connection is lost. Try to reconnect for up to 60s first
- Don't require an abort/restart when losing zk connection. Wait up to 30s for reconnect first
  - When regaining zk connection, as long as we did not also lose mesos connection, don't re-run reconciliation
- When losing leadership, don't immediately abort, stop the pollers, clear leader cache and continue running as a hot standby

TODOs:
- [x] fix up unit tests. Need some newer guice magic to change how the unit test version of the scheduler works
- [x] Should we retry forever on mesos/zk connection? Wondering if there is a downside to spinning forever versus restarting. I _think_ as long as we've properly stopped all our leader pollers and leader cache, that it should be safe to continue waiting. Need to verify
- [x] Figure out what to do about buffering status updates while we are in a state where we have mesos connection but not zk. Should we reject these? Eventually buffer to disk? Status updates need zk to finish processing so we either have to buffer, or re-run reconciliation when restarting.

cc @jkebinger @baconmania @jschlather

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
@@ -93,6 +93,10 @@ public class SingularityLeaderController implements LeaderLatchListener, Connect
       if (!newState.isConnected()) {
         LOG.info("No longer connected to zk, pausing scheduler actions and waiting up to {}ms for reconnect",
             configuration.getZooKeeperConfiguration().getAbortAfterConnectionLostForMillis());
+        if (lostConnectionStateChecker != null) {
+          LOG.warn("Already started connection state check, due in {}ms", lostConnectionStateChecker.scheduledExecutionTime() - System.currentTimeMillis());
+          return;
+        }
         lostConnectionStateChecker = new TimerTask() {
           @Override
           public void run() {
@@ -110,6 +114,7 @@ public class SingularityLeaderController implements LeaderLatchListener, Connect
       } else if (lostConnectionStateChecker != null) {
         LOG.info("Reconnected to zk, scheduler actions resumed");
         lostConnectionStateChecker.cancel();
+        lostConnectionStateChecker = null;
       }
     } finally {
       stateHandlerLock.unlock();
--- a/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
@@ -101,6 +101,10 @@ public class SingularityLeaderController implements LeaderLatchListener, Connect
           @Override
           public void run() {
             stateHandlerLock.lock();
+            if (scheduler.getState().getZkConnectionState().isConnected()) {
+              LOG.debug("Reconnected to zk, will not abort");
+              return;
+            }
             try {
               LOG.error("Aborting due to loss of zookeeper connection for {}ms. Current connection state {}",
                   configuration.getZooKeeperConfiguration().getAbortAfterConnectionLostForMillis(), newState);
--- a/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
@@ -166,7 +166,26 @@ public class SingularityLeaderController implements LeaderLatchListener, Connect
       LOG.info("We are not the leader! Current state {}", scheduler.getState());
       master = false;
       if (!isTestMode()) {
-        scheduler.notLeader();
+        LOG.info("Might not be the leader, pausing scheduler actions");
+        scheduler.pauseForDatastoreReconnect();
+        // Check again if we are still not leader in a few seconds. LeaderLatch.reset can often get called on reconnect, which
+        // will call notLeader/isLeader is quick succession.
+        TIMER.schedule(new TimerTask() {
+          @Override
+          public void run() {
+            stateHandlerLock.lock();
+            try {
+              if (master) {
+                LOG.debug("Reconnected as leader before shutdown timeout");
+              } else {
+                LOG.info("No longer the leader, stopping scheduler actions");
+                scheduler.notLeader();
+              }
+            } finally {
+              stateHandlerLock.unlock();
+            }
+          }
+        }, 5000);
       }
     } finally {
       stateHandlerLock.unlock();
--- a/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
@@ -3,14 +3,20 @@ package com.hubspot.singularity;
 import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
 import java.util.Optional;
+import java.util.Timer;
+import java.util.TimerTask;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import javax.inject.Singleton;
 
+import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.recipes.leader.LeaderLatchListener;
+import org.apache.curator.framework.state.ConnectionState;
+import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.mesos.v1.Protos.MasterInfo;
 import org.apache.mesos.v1.Protos.Offer;
 import org.slf4j.Logger;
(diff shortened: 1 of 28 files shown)


SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
```diff
@@ -40,6 +47,11 @@
   private final StatePoller statePoller;
   private final SingularityMesosScheduler scheduler;
   private final OfferCache offerCache;
+  private final SingularityConfiguration configuration;
+
+  private AtomicReference<ConnectionState> lastState;
+  private final ReentrantLock stateHandlerLock;
+  private TimerTask lostConnectionStateChecker;
```
jschlather: This should be volatile.

SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
```diff
@@ -40,6 +47,11 @@
   private final StatePoller statePoller;
   private final SingularityMesosScheduler scheduler;
   private final OfferCache offerCache;
+  private final SingularityConfiguration configuration;
+
+  private AtomicReference<ConnectionState> lastState;
```
jschlather: This should be final?

SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
```diff
@@ -71,6 +85,40 @@ public void start() {
...
+  }
+
+  @Override
+  public void stateChanged(CuratorFramework client, ConnectionState newState) {
+    stateHandlerLock.lock();
+    try {
+      scheduler.setZkConnectionState(newState);
+      ConnectionState previous = lastState.get();
+      if (!newState.isConnected()) {
+        scheduler.notLeader();
```
jkebinger: Logging here that connection has been lost and we're going to hang out for X millis would be useful for debugging

SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
```diff
@@ -71,6 +85,40 @@ public void start() {
...
+                  configuration.getZooKeeperConfiguration().getAbortAfterConnectionLostForMillis(), newState);
+              abort.abort(AbortReason.LOST_ZK_CONNECTION, Optional.empty());
+            } finally {
+              stateHandlerLock.unlock();
+            }
+          }
+        };
+        TIMER.schedule(lostConnectionStateChecker, configuration.getZooKeeperConfiguration().getAbortAfterConnectionLostForMillis());
+      } else if (previous != null && !previous.isConnected()) {
+        // Reconnected in time!
```
jkebinger: Logging here that connection was regained and we're GTG will be useful too

ssalinas:
Found a few more things:
- We were treating `SUSPENDED` as an error state in CuratorFramework itself (this is apparently the default). Without the `SessionConnectionStateErrorPolicy` added, LeaderLatch will lose leadership on any zk state change
- notLeader/isLeader can be called in quick succession on reconnections. So, it's possible that we still have leadership even when notLeader gets called. Wait a few seconds before fully shutting down the scheduler here

SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
```diff
@@ -71,6 +81,39 @@ public void start() {
...
+  @Override
+  public void stateChanged(CuratorFramework client, ConnectionState newState) {
+    LOG.info("Received update for new zk connection state {}", newState);
+    stateHandlerLock.lock();
+    try {
+      scheduler.setZkConnectionState(newState);
+      if (!newState.isConnected()) {
+        LOG.info("No longer connected to zk, pausing scheduler actions and waiting up to {}ms for reconnect",
+            configuration.getZooKeeperConfiguration().getAbortAfterConnectionLostForMillis());
+        lostConnectionStateChecker = new TimerTask() {
```
jschlather: The connection state can potentially go `SUSPENDED -> LOST -> RECONNECTED`, if the session times out. In this case, we would schedule a second `lostConnectionStateChecker`, which would override the first one. If we then reconnected, we would cancel the second lost state checker, but not the first. We probably want some sort of guard around this, either cancelling the existing task if it's present or not adding if there's a recent enough time task.

SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
```diff
@@ -71,6 +81,39 @@ public void start() {
...
+          @Override
+          public void run() {
+            stateHandlerLock.lock();
```
jschlather: There's a small race here. If a reconnect happens and acquires the lock, then this TimerTask executes and begins waiting for the lock. The cancel does not interrupt the TimerTask, so it will still execute the abort. You'd probably need to some sort of boolean guard to this, because TimerTask doesn't expose whether or not its been cancelled.

ssalinas:
Staring down the SingularityLeaderController code a bit more, but I think I got all the races handling reconnects taken care of now

SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
ssalinas: Added a re-check of the current connection state after acquiring the lock, that should prevent this

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerImpl.java
```diff
@@ -327,10 +358,65 @@ private void callWithStateLock(Runnable function, String name, boolean ignoreIfN
     }
   }
 
+  public void reconnectMesos() {
+    callWithStateLock(() -> {
+      state.setMesosSchedulerState(MesosSchedulerState.PAUSED_FOR_MESOS_RECONNECT);
+      mesosSchedulerClient.close();
```
jschlather: Can this throw an exception?

ssalinas: Yes, but that will bubble to our uncaught exception handling for anything in this class (which will end up causing an abort). I could work in some more handling for weirdness stopping the mesos connection, but to start with I think it's a good improvement to be able to retry it, and still abort if something goes wrong in that retrying flow

jschlather: If the exception gets handled higher then it should be fine. I was mostly concerned about it getting stuck in a paused state indefinitely.

SingularityService/src/main/java/com/hubspot/singularity/mesos/StatusUpdateQueue.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public void add(TaskStatus update) throws IOException {
+    if (inMemoryQueue.size() < configuration.getMesosConfiguration().getMaxStatusUpdateQueueSize()) {
+      inMemoryQueue.add(update);
+    } else {
+      onDiskQueue.add(update);
+    }
+  }
+
+  public void iterate(Function<TaskStatus, CompletableFuture<StatusUpdateResult>> function) throws IOException {
```
jschlather: Should this class be thread safe? Maybe make `add` and `iterate` synchronized?

ssalinas: Theoretically shouldn't need to be. But added it to be safe

ssalinas:
staging running smoothly after temporarily blocking the service machine from hitting zk for 1s, 5s, 10s, 25s, and aborts properly after 30s

jschlather:
Great, can we test a leader election too? If ZK is on kubernetes in staging, then it's easy to induce by killing the leader pod.

ssalinas:
Reconnect to zk has been working great so far. Reconnect to mesos had a few issues I addressed in the recent commits. Also need to take one more look at the semaphore we are using to chew through queued status updates. We seem to be overloading it too easily

ssalinas:
Documenting some additional changes that were added:
- Consider a status update a heartbeat from mesos. We often fell behind on status updates, and the next hearbeat was behind a pile of 20k status updates, so we then failed over thinking we lost mesos.
- startup pending request creations now need a request lock since they can be run concurrent with other pollers during reconnect events
- Fix shtudown ordering and mark scheduler as STOPPED earlier in the shutdown flow to avoid a bunch of needless RejectedExecutionExceptions on shutdown
- Throttle status update submissions to the semaphore when catching up on queued backlog. We often queued thousands then submitted all at once, causing the async updater's queue to overflow
- Need to catch Interrupted and wrapped Interrupted exceptions properly in the subscriberThread. observable.unsubscribe causes the original thread sitting at `openStream.await()` to get an InterruptedException wrapped in a RuntimeException

baconmania:
🚢

ssalinas:
consolidating in #2032

PR #2029, opened 2019-10-29
Timing-related fixes

ssalinas:
Few that will be in this PR
- [x] When skip healthchecks expires, any tasks that were initially marked healthy due to that should remain as healthy. Accomplishing this with a dummy healthcheck result, since pulling in request history to check at the time of startup would pull a dep on mysql on the critical scheduling path
- [x] Fix handling of baragon request already in queue with different params. This should not result in zero upstreams when rolling back a deploy (odd failover race condition)
- [x] Fix race condition where pending deploy gets stuck even though it already has a result in zk (odd failover mode)

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
@@ -48,6 +48,7 @@ public class LoadBalancerClientImpl implements LoadBalancerClient {
 
   private static final String CONTENT_TYPE_JSON = "application/json";
   private static final String HEADER_CONTENT_TYPE = "Content-Type";
+  private static final String ALREADY_ENQUEUED_ERROR = "is already enqueued with different parameters";
 
   private final String loadBalancerUri;
   private final Optional<Map<String, String>> loadBalancerQueryParams;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/hooks/LoadBalancerClientImpl.java
```diff
@@ -130,6 +130,14 @@ private BaragonResponse readResponse(Response response) {
   private SingularityLoadBalancerUpdate sendRequestWrapper(LoadBalancerRequestId loadBalancerRequestId, LoadBalancerMethod method, Request request, BaragonRequestState onFailure) {
     final long start = System.currentTimeMillis();
     final LoadBalancerUpdateHolder result = sendRequest(loadBalancerRequestId, request, onFailure);
+
+    if ((method != LoadBalancerMethod.CHECK_STATE && method != LoadBalancerMethod.PRE_ENQUEUE) &&
+        result.state == BaragonRequestState.FAILED
+        && result.message.orElse("").contains("is already enqueued with different parameters")) {
+      LOG.info("Baragon request {} already in the queue, will fetch current state instead", loadBalancerRequestId.getId());
```
baconmania: Bit of a nit, but is there any cleaner way to check for this state other than matching the against the message string? Or if we do have to do a string match, can we put the string into a constant somewhere?

ssalinas: Yeah, the longer term way would be to make baragon return a different status code. Right now this is the only thing that differentiates it from other non-specific failures. Can definitely put it in a constant for the time being

baconmania:
🚢

PR #2030, opened 2019-10-30
Subscribe on other thread for faster status updates

ssalinas:
Found recently that most methods in SingularityMesosSchedulerImpl were all being processed on a single thread. For some things like status updates this is fine since they are then kicked over to another pool of threads. However for things like offers, we hog the thread for a while, making status updates very delayed. Let's use more than one thread....

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
@@ -39,6 +39,7 @@ import com.hubspot.mesos.rx.java.SinkOperation;
 import com.hubspot.mesos.rx.java.SinkOperations;
 import com.hubspot.mesos.rx.java.protobuf.ProtobufMesosClientBuilder;
 import com.hubspot.mesos.rx.java.util.UserAgentEntries;
+import com.hubspot.singularity.SingularityManagedThreadPoolFactory;
 import com.hubspot.singularity.config.MesosConfiguration;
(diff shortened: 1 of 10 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
```diff
@@ -173,7 +180,8 @@ private void connect(URI mesosMasterURI, FrameworkInfo frameworkInfo, Singularit
 
     subscribe.processStream(unicastEvents -> {
 
-      final Observable<Event> events = unicastEvents.share();
+      final Observable<Event> events = unicastEvents.share()
+          .subscribeOn(subscribeScheduler);
```
baconmania: Is there any risk in having events pertaining to the same Mesos host end up on two different subscriber threads simultaneously?

ssalinas: I still need to stare at it more, but I don’t think so. Most of SchedulerImpl was written assuming it was already concurrent. Offer related traffic is behind a lock, as are any updates to state. Lost executors are mostly just logging. Status updates are a sync and behind request locks. Think that’s all?

ssalinas:
One additional change, gave the offer-related events and status updates separate pools. It's possible we could get 4,5,6,etc offer events backed up and still block status updates if they aren't separated

baconmania:
🚢

ssalinas:
consolidating in #2032

PR #2031, opened 2019-10-31
Only delay for status updates when many requests are delayed

ssalinas:
Attempt at fixing the delay issues for status updates. After the other current open PRs are merged, we still have one case where a particularly active request with many tasks being launched will have status updates delayed. Offer processing to launch 200+ tasks ma hold the request lock for a fair amount of time, delaying status updates for that particular request. Other request updates are fine, but the overall delay still looks high and can short circuit offer polling.

This changes the logic to only delay other pollers when a higher percentage of requests have delayed status updates. One or two requests with large piles of status updates will then not be able to slow down processing of everything else.


ssalinas:
conflicts heavily with https://github.com/HubSpot/Singularity/pull/2028 , merging these changes into there

PR #2032, opened 2019-10-31
Reconnect, offer, and startup performance improvements

ssalinas:
Based off of #2028 to not cause merge conflicts. Need to test if this helps our scheduling issues we've seen lately


sjeropkipruto:
🚢

PR #2034, opened 2019-11-04
More performance updates

ssalinas:
- remove all parallelStream calls and repalce with a managed executor
- reduce default concurrency everywhere
- bail out of offer checks and decline if offer processing is too delayed
- add a good enough threshold on offer scores to skip processing additional checks
- sql history purger doesn't need to hold the lock since it doesn't touch zk

more to come soon....


ssalinas:
Next round of updates:
- bump the status update concurrency back up a slight bit since we spend much of the time waiting for locks, not in cpu-intensive things
- load leader cache in parallel. This is 99% of our startup time right now once subscribed
- update the offer loop to wait for locks in parallel, but run offer checking in serial. Have to test this more but it should make the overall offer loop much faster
- use a timeout when acquiring the offers lock for net-new offers. If we don't get it in time, cache these offers instead. This will let the offer loop run as fast as it can without backing up, and in general process larger batches of offers, with more possibility for finding a match
- bump the UI notification back to 100 late requests instead of 20

ssalinas:
latest additions:
- don't let the persister hold the lock for more than 10s at a time. We can always catch up on persisting later

SingularityService/src/main/java/com/hubspot/singularity/hooks/SnsWebhookManager.java
```diff
@@ -73,8 +72,7 @@ public SnsWebhookManager(@Singularity ObjectMapper objectMapper,
       this.snsClient = AmazonSNSClientBuilder.defaultClient();
     }
     this.webhookManager = webhookManager;
-    this.publishSemaphore = AsyncSemaphore.newBuilder(configuration::getMaxConcurrentWebhooks, executorServiceFactory.get("webhook-publish-semaphore", 1)).build();
-    this.publishExecutor = managedCachedThreadPoolFactory.get("webhook-publish");
+    this.publishExecutor = managedCachedThreadPoolFactory.get("webhook-publish", configuration.getMaxConcurrentWebhooks());
```
baconmania: Leaving a note here to double check our naming here, for both `managedCachedThreadPoolFactory` and `maxConcurrentWebhooks`

SingularityService/src/main/java/com/hubspot/singularity/SingularityLeaderController.java
```diff
@@ -141,8 +141,8 @@ public void isLeader() {
       master = true;
       try {
         if (!isTestMode()) {
-          scheduler.start();
           statePoller.wake();
+          scheduler.start();
```
baconmania: Small change, but I'm curious: why?

ssalinas: If we call the state poller first we get instant feedback in the UI about who is leader versus having to wait until after startup is already finished to know which host gained leadership

SingularityService/src/main/java/com/hubspot/singularity/hooks/SnsWebhookManager.java
ssalinas: did the rename for these. Going to leave the config var the same though for easier backwards compatibility. It is essentially still controlling the same thing, we are just being limited by the size of a fixed thread pool vs the async semaphore now. A future update could be to consolidate all the thread pool size configs in one spot

baconmania:
:ship:

PR #2036, opened 2019-11-07
Add endpoint to fetch all active task states

ssalinas:
cc @szabowexler , this should be a faster way to fetch what you're looking for


baconmania:
🚢

PR #2037, opened 2019-11-07
offer executor needs at least 2 threads

ssalinas:
:faceplam: had this commit locally and never pushed. This makes the other piece of the newer thread model work, allowing subsequent offer traffic to come in, time out on the lock, and instead add to the cache. With a single thread it just gets queued instead


baconmania:
🚢

PR #2038, opened 2019-11-12
Catch more SQL exceptions in persister queries

ssalinas:
If Singulairty was interrupted between the history persist and the removal from zk, subsequent attempts to run the persister would fail out on unique key errors. request and deploy history can easily be updated on duplicate primary key instead. The task history persister does not have this problem because it checks in a different way

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/history/MySQLHistoryJDBI.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/history/MySQLHistoryJDBI.java
@@ -18,13 +18,11 @@ import com.hubspot.singularity.data.history.SingularityMappers.SingularityReques
 
 public interface MySQLHistoryJDBI extends AbstractHistoryJDBI {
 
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/history/MySQLHistoryJDBI.java
```diff
@@ -18,11 +18,13 @@
 
 public interface MySQLHistoryJDBI extends AbstractHistoryJDBI {
 
-  @SqlUpdate("INSERT INTO requestHistory (requestId, json, createdAt, requestState, user, message) VALUES (:requestId, :json, :createdAt, :requestState, :user, :message)")
+  @SqlUpdate("INSERT INTO requestHistory (requestId, json, createdAt, requestState, user, message) VALUES (:requestId, :json, :createdAt, :requestState, :user, :message) " +
+      "ON DUPLICATE KEY UPDATE json = :json, requestState = :requestState, user = :user, message = :message")
```
ssalinas: this is causing tests to fail but so far can't figure out what's wrong with this query

baconmania:
🚢

PR #2039, opened 2019-11-12
Make task history persister not use the request level lock

ssalinas:
This was added when we attempted to work on immediate persisting of task history items. However, when looking through more recent changes to the code, the request level lock should no longer be needed here, since data persisted here is not part of the active task set. This will free up the lock during scheduling for further improvements to scheduling efficiency

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityTaskHistoryPersister.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityTaskHistoryPersister.java
@@ -1,11 +1,12 @@
 package com.hubspot.singularity.data.history;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityTaskHistoryPersister.java
```diff
@@ -65,43 +56,36 @@ public void runActionOnPoll() {
...
-      LOG.info("Transferred {} out of {} inactive task ids (total {}) in {}", numTransferred, numTotal, allTaskIds.size(), JavaUtils.duration(start));
     } finally {
       persisterLock.unlock();
     }
   }
 
+  private Map<String, List<SingularityTaskId>> getInactiveTaskIdsByRequest() {
+    final List<SingularityTaskId> taskIds = new ArrayList<>(taskManager.getAllTaskIds());
+    taskIds.removeAll(taskManager.getActiveTaskIds());
+    taskIds.removeAll(taskManager.getLBCleanupTasks());
```
baconmania: Not a huge deal, but can we do these `removeAll`s against a Set instead of a List?

ssalinas: yeah definitely, originally had it as a list because i was sorting it, but then switched to the stream at the end instead. Will update

baconmania:
🚢

PR #2041, opened 2019-11-20
Findbugs fixes

ssalinas:
Not sure how this didn't trigger on our staging/qa/stable branches, but it triggered when I tried to build master, so fixing


pschoenfelder:
🚢

PR #2043, opened 2019-11-20
Corrected annotation for findbugs

ssalinas:
newer maven/findbugs seems to pick up on SuppressWarnings, but older does not. Trying it this way instead


pschoenfelder:
🚢

PR #2046, opened 2019-11-21
Process a reconciliation for all tasks with missing state in zk

ssalinas:
We've had a few orphaned tasks recently. It seems the mesos master still knew about them but we weren't marking them as active. I'm still not certain how they got into that state, but this will at least reconcile it afterwards so they get cleaned up properly

one possibility for how they got there is a bug in the persister, so adding logging there as well

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
@@ -128,8 +128,7 @@ public class SingularityMesosStatusUpdateHandler {
     if (!previousTaskStatusHolder.isPresent() // Task was already removed from the active list
         && !taskState.isDone()
         && newTaskStatusHolder.getTaskStatus().isPresent()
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -128,8 +128,7 @@ private boolean isRecoveryStatusUpdate(Optional<SingularityTaskStatusHolder> pre
     if (!previousTaskStatusHolder.isPresent() // Task was already removed from the active list
         && !taskState.isDone()
         && newTaskStatusHolder.getTaskStatus().isPresent()
-        && ACTIVE_STATES.contains(newTaskStatusHolder.getTaskStatus().get().getState())
-        && reason == Reason.REASON_AGENT_REREGISTERED) {
```
ssalinas: Removing this additional condition will let us process and reconcile all tasks that are in mesos master but not zk, not only the cones with an explicit reconciliation update state

baconmania:
🚢

PR #2048, opened 2019-12-09
Don't mutate args in lb sync comparison

ssalinas:
Didn't realize this was happening when the Pr went through. A check a few lines down later validates the size of the argument list against the one it is about to use for removal.


baconmania:
🚢

PR #2049, opened 2019-12-11
Prioritize memory shuffles over cpu

ssalinas:
Memory ones are generally more urgent, take care of them first

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
@@ -22,7 +22,6 @@ import com.hubspot.singularity.RequestUtilization;
 import com.hubspot.singularity.SingularityAction;
 import com.hubspot.singularity.SingularityClusterUtilization;
 import com.hubspot.singularity.SingularityDeploy;
-import com.hubspot.singularity.SingularityManagedScheduledExecutorServiceFactory;
 import com.hubspot.singularity.SingularityManagedThreadPoolFactory;
 import com.hubspot.singularity.SingularityPendingRequest;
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsagePoller.java
```diff
@@ -160,6 +159,13 @@ private void shuffleTasksOnOverloadedHosts(Map<SingularitySlaveUsage, List<TaskI
         .sorted((usage1, usage2) -> {
           OverusedResource mostOverusedResource1 = getMostOverusedResource(usage1, getSystemLoadForShuffle(usage1), usage1.getMemoryBytesUsed());
           OverusedResource mostOverusedResource2 = getMostOverusedResource(usage2, getSystemLoadForShuffle(usage2), usage2.getMemoryBytesUsed());
+          if (mostOverusedResource1.resourceType != mostOverusedResource2.resourceType) {
+            if (mostOverusedResource1.resourceType == Type.MEMORY) {
+              return -1;
+            } else {
+              return 1;
+            }
+          }
```
baconmania: Hmm, this short circuits _all_ resource comparisons when the two resources being compared aren't the same. So like, if we're comparing `DISK` vs `CPU`, whichever one is `mostOverusedResource1` would always sort as higher. Is that the intent? Should we scope this logic to the specific scenario when one of the resources being compared is `MEMORY`?

ssalinas: We don't have shuffles for disk so far. Type can only be MEM or CPU

baconmania: Ahhh, derp. cool

baconmania:
🚢

PR #2050, opened 2019-12-17
Better preferential scheduling

sjeropkipruto:
#### Description
Mesos hosts aren't nearly as well utilized as they could be. The high mem ones fill up on cpu before we can make full use of the extra memory and vice versa for the high cpu ones. Similar for spot, not all things that can run on spot actually are.

We want to improve our scheduling to really improve our utilization and make scheduling faster & easier because there are more available large spots of resources.

#### Testing
- [x] unit testing 
- [x] deploying to staging

#### Todos
- [x] Implement preference by allowed attributes. If a deployable specifies an allowed attribute, give preference to that type of machine. For example, if it can run on spot, prefer spot. If it is allowed to run on spark hosts, prefer those.
- [x] Implement preference by resource preference. Identify mesos hosts as normal or having a particular resource preference (e.g. a high memory host). Do the same for tasks based on their measured used resources. Matches between tasks and hosts should be preferred (e.g. high memory task on a high memory host).
- [x] Write unit tests 
- [x] Deploy and test in staging
- [x] Deploy to QA 
- [x] Deploy to PROD


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -579,7 +578,12 @@ private double score(SingularityOfferHolder offerHolder, SingularityTaskRequestH
     final SlaveMatchState slaveMatchState = slaveAndRackManager.doesOfferMatch(offerHolder, taskRequest, activeTaskIdsForRequest, isPreemptibleTask(taskRequest, deployStatsCache));
```
ssalinas: I think your overall strategy here is good, but we need to be a bit more efficient. We already iterate over allowed/required attributes during this doesOfferMatch call here. Maybe we can communicate back a flag in the `SlaveMatchState` that lets this code know to boost the score? That way we don't have to do the attribute comparison and calculation a second time. Instead of just `SlaveMatchState.OK` could be `SlaveMatchState.PREFERRED_SLAVE` or something like that

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularitySlaveAndRackHelper.java
```diff
@@ -116,4 +118,57 @@ public boolean containsAllAttributes(Map<String, String> attributes, Map<String,
...
+  }
+
+  private double getCpuMemoryRatioForRequest(RequestUtilization requestUtilization) {
+    double cpuUsageForRequest = getEstimatedCpuUsageForRequest(requestUtilization);
+    double memUsageForRequest = requestUtilization.getAvgMemBytesUsed();
+    return cpuUsageForRequest/memUsageForRequest;
+  }
+
+  private double getEstimatedCpuUsageForRequest(RequestUtilization requestUtilization) {
+    // To account for cpu bursts, tend towards max usage if the app is consistently over-utilizing cpu, tend towards avg if it is over-utilized in short bursts
```
ssalinas: if this was copied from the other offers class, can we make sure we don't duplicate it? Maybe give it a new home in another class

ssalinas:
Overall concept looks good 👍 , let's get some unit tests in to make sure it behaves as expected then give it a go in staging

ssalinas:
🚢

PR #2051, opened 2019-12-17
Better crash loop detection

ssalinas:
The previous changes to Singularity's cooldown checking did not accomplish what we had hoped. We are still in the boat of it being either too noisy or missing things. This PR is aimed at fixing cooldown and splitting out a new crash loop concept. It will:

- Return cooldown to only being a protection against fast-failing services. It was originally intended to protect us from scheduling things too often that we know will fail. I am returning it to only this purpose
- Create a new api for the detection of 'crash loops' that is separate from cooldown. These will be divided into different cases such as repeated healthcheck failures, unexpected exits, etc so that Singularity can provide more detail on the reasoning for it

So far the PR sets up the plumbing for exposing this on the api. Still TODO:
- [x] Actual logic to return cooldown to a fast-failure only state
- [x] Logic to detect each type of crash loop, will likely need to update the data stored in the deploy statistics object for this
- [x] Unit tests for ^
- [x] (Maybe) Add a webhook for when new crash loops are detected?


ssalinas:
Finished off most of the wiring up and POJOs. Meat of the logic is outlined and will be implemented in SingularityCrashLoops/SingularityCrashLoopChecker

sjeropkipruto:
🚢

PR #2052, opened 2019-12-18
Carry forward "Task launched because" message when retrying task.

baconmania:


ssalinas:
🚢

baconmania:
image
Works, merging through.

PR #2054, opened 2019-12-23
fix mismatch in http/https protocol

yiwork:
This should fix issues where API reference page will not load when there's a protocol mismatch when this page is accessed with https and the swagger ui JS loads the swagger reference page using http.

I've not been able to get the public API reference page for months now due to this error


ssalinas:
🚢

PR #2055, opened 2019-12-31
Health Checks V2

pschoenfelder:


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -115,4 +131,84 @@ private boolean runHealthcheck(Process process) {
...
+
+    for (HealthcheckResult result : results) {
+      HealthCheckStatus status = result.getStatus();
+      if (status != HealthCheckStatus.PASSED) {
+        executorUtils.sendStatusUpdate(task.getDriver(), task.getTaskInfo().getTaskId(), TaskState.TASK_FAILED, String.format("Healthcheck failed: {}", result.getMessage()), task.getLog());
+        return false;
+      }
+    }
+    executorUtils.sendStatusUpdate(task.getDriver(), task.getTaskInfo().getTaskId(), Protos.TaskState.TASK_RUNNING, String.format("Task running process %s.", getCurrentProcessToString()), task.getLog());
+    return true;
```
pschoenfelder: My biggest concern here is that any given healthcheck
- doesn't prevent another from running
- doesn't prevent early exit due to an earlier failing healthcheck

pschoenfelder: I think this does this, but definitely extra eyes here would be good

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -72,6 +81,13 @@ public String toString() {
 
   private boolean runHealthcheck(Process process) {
     Optional<HealthcheckOptions> maybeOptions = task.getTaskDefinition().getHealthcheckOptions();
+    if (maybeOptions.isPresent() && maybeOptions.get().getExpectHealthcheckConfigFile()) {
```
ssalinas: Can we change this up a bit. The need for `expectHealthcheckConfigFile` doesn't solve the problem we're having. Right now we have the issue that we don't know at deploy time (when the SingularityDeploy is created) if the underlying things being deployed will or won't have file healthchecks implemented. So, I was thinking this would be more of a check like 'if healthcheck config file present -> run v2 -> else run v1'

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -115,4 +131,93 @@ private boolean runHealthcheck(Process process) {
       return false;
     }
   }
+
+  private boolean runHealthcheckV2(Process process, HealthcheckOptions maybeOptions) {
+    String taskAppDirectory = task.getTaskDefinition().getTaskAppDirectory();
```
ssalinas: just confirming, does this equate to essentially `./healthcheck-config.json` from within task code?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
```diff
@@ -115,4 +131,93 @@ private boolean runHealthcheck(Process process) {
...
+        return false;
+      }
+    }
+    executorUtils.sendStatusUpdate(task.getDriver(), task.getTaskInfo().getTaskId(), Protos.TaskState.TASK_RUNNING, String.format("Task running process %s.", getCurrentProcessToString()), task.getLog());
+    return true;
+  }
+
+  private CompletableFuture<HealthcheckResult> startHealthcheck(Process process, HealthcheckConfig healthcheckConfig) {
+    Retryer<HealthcheckResult> retryer = buildHealthcheckRetryer(healthcheckConfig);
+    return CompletableFuture.supplyAsync( () -> {
```
ssalinas: We should make sure to use an ExecutorService with this. Ideally shut it down right after as well

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployHealthHelper.java
```diff
@@ -246,6 +246,12 @@ public DeployHealth getTaskHealth(SingularityDeploy deploy, boolean isDeployPend
         return DeployHealth.HEALTHY;
       }
     }
+    if (deploy.getHealthcheck().isPresent() && deploy.getHealthcheck().get().getExpectHealthcheckConfigFile()) {
```
ssalinas: not sure how to do this part without the getExpectHealthcheckConfigFile. Not sure if it's even necessary in that case?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
pschoenfelder: I was considering that but saw it as an opportunity for falsely passing health checks in the case that a config file doesn't appear in time. Is the timing of that something we're comfortable making a guarantee about?

My thinking is that this flag would tell the executor to fail health checks if the task doesn't provide a configuration when it should, and eliminate the false pass scenario.

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskProcessCallable.java
pschoenfelder: Yea exactly.

pschoenfelder:
Closing in favor of a newfound way to work with the old health checks.

PR #2057, opened 2020-01-13
Slightly Smarter Task Shuffling

WH77:
The current task shuffling/eviction logic used by Singularity prioritizes shuffling tasks with the highest resource utilization (memory or CPU, depending on which is most overused), on each overcommitted slave. Though this is a good way to ensure the shuffle succeeds at reducing load to acceptable levels, it is also much more likely to shuffle tasks performing large amounts of work or holding large amounts of in-memory state. This makes sense for high CPU tasks, but is undesirable for high memory ones - they'll likely return to their previous memory consumption and possibly trigger another shuffle on their new host.

To address these issues, this PR attempts to:
- Alter the shuffling logic for memory overusage to favor shuffling "low utilization" tasks.
- Create a simple scoring metric to identify "low utilization" tasks. Currently, this is just the average of a task's memory/CPU usage. Suggestions on whether to use max, min, or other methods would be welcome.
- Refactor shuffling logic from `SingularityUsagePoller` to a new `SingularityTaskShuffler` class, while preserving existing functionality/signatures.

Because of these changes, it is very likely that tasks that were not shuffled before will now be shuffled, which in turn has a fair chance of causing issues with these tasks in QA/prod. I don't think there's a good way to avoid this and also actually change the shuffling logic, though self-service shuffle opt-out may help mitigate things.

TODO:
- [ ] Add more complex shuffling unit tests.
- [ ] Evaluation in staging cluster.
- [ ] Ensure that enough resources are freed up while respecting shuffle limits in configuration.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskShuffler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskShuffler.java
@@ -1,20 +1,25 @@
 package com.hubspot.singularity.scheduler;
 
-import java.util.*;
-import java.util.stream.Collectors;
-
-import com.hubspot.singularity.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.google.inject.Inject;
+import com.hubspot.singularity.SingularityPendingRequest;
 import com.hubspot.singularity.SingularityPendingRequest.PendingType;
+import com.hubspot.singularity.SingularitySlaveUsage;
+import com.hubspot.singularity.SingularityTaskCleanup;
+import com.hubspot.singularity.TaskCleanupType;
 import com.hubspot.singularity.config.SingularityConfiguration;
 import com.hubspot.singularity.data.RequestManager;
 import com.hubspot.singularity.data.TaskManager;
 import com.hubspot.singularity.scheduler.SingularityTaskShuffler.OverusedResource.Type;
-
 import io.dropwizard.util.SizeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
 
 public class SingularityTaskShuffler {
   private static final Logger LOG = LoggerFactory.getLogger(SingularityTaskShuffler.class);
@@ -33,7 +38,9 @@ public class SingularityTaskShuffler {
   }
 
   static class OverusedResource {
-    enum Type {MEMORY, CPU};
+    enum Type {MEMORY, CPU}
+
+    ;
 
     double overusage;
     Type resourceType;
@@ -93,7 +100,7 @@ public class SingularityTaskShuffler {
       long shufflingTasksOnSlave = 0;
       double cpuUsage = getSystemLoadForShuffle(slave.usage);
       double memUsageBytes = slave.usage.getMemoryBytesUsed();
-      
+
       TaskCleanupType shuffleCleanupType = slave.resource.toTaskCleanupType();
       List<TaskIdWithUsage> shuffleCandidates = getPrioritizedShuffleCandidates(slave);
 
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskShuffler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskShuffler.java
@@ -124,8 +124,8 @@ public class SingularityTaskShuffler {
 
       long shufflingTasksOnSlave = shufflingTasksPerHost.getOrDefault(getHostId(slave).orElse(""), 0L);
       long availableTasksOnSlave = shuffleCandidates.size();
-      double cpuUsage = getSystemLoadForShuffle(slave.usage);
-      double memUsageBytes = slave.usage.getMemoryBytesUsed();
+      double cpuUsage = getSystemCpuLoadForShuffle(slave.usage);
+      double memUsageBytes = getSystemMemLoadForShuffle(slave.usage);
 
       for (TaskIdWithUsage task : shuffleCandidates) {
         availableTasksOnSlave--;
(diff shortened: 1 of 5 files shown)


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskShuffler.java
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  private void bounce(TaskIdWithUsage task, TaskCleanupType cleanupType, Optional<String> message) {
+    taskManager.createTaskCleanup(new SingularityTaskCleanup(
+        Optional.empty(),
+        cleanupType,
+        System.currentTimeMillis(),
+        task.getTaskId(),
+        message,
+        Optional.of(UUID.randomUUID().toString()),
```
ssalinas: Small nit, it would be good for the actionId here and in the SingularityPendingRequest to match. For this particular case it doesn't have any lasting effect on the actions themselves, but makes it easier to trace through the data for debugging if they are equal

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskShuffler.java
```diff
@@ -0,0 +1,279 @@
...
+
+  private double getUtilizationScoreForCPUShuffle(TaskIdWithUsage task) {
+    return task.getUsage().getCpusUsed() / task.getRequestedResources().getCpus();
+  }
+
+  private double getUtilizationScoreForMemoryShuffle(TaskIdWithUsage task) {
+    double memoryUtilization = task.getUsage().getMemoryTotalBytes() / task.getRequestedResources().getMemoryMb();
+    double cpuUtilization = task.getUsage().getCpusUsed() / task.getRequestedResources().getCpus();
+
+    return (memoryUtilization + cpuUtilization) / 2;
```
ssalinas: Still thinking more on the metric for this. You were correct to point out the fact that, by shuffling more/smaller tasks we may very well hit the shuffling limits more often. I'm not certain if it will be often enough to be of consequence given the size of our cluster, but there are 2 things I could see us doing to compensate for it:
- Possibly make the shuffle limits a percentage of the overall task count (i.e. so we don't have to worry about it being more restrictive as the cluster grows larger)
- Possibly make a slightly more complex scoring mechanism. I could see a case where there is maybe a threshold for cpu (something simple like a 50-60% cutoff) in memory shuffle. If tasks are _below_ that, they are given a higher priority for high memory usage (inverse of now), if they are above that they are scored as they are now. This could serve to put a few higher memory, but more likely less busy tasks at the top of the queue, resulting in fewer overall shuffles

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityUsageTest.java
```diff
@@ -688,6 +706,249 @@ public void itLimitsTheNumberOfTaskCleanupsToCreate() {
...
+      Assertions.assertFalse(taskManager.getTaskCleanup(task3.getId()).isPresent());
+
+      // Fourth task is cleaned up, as it has the lowest utilization.
+      Assertions.assertEquals(TaskCleanupType.REBALANCE_MEMORY_USAGE, taskManager.getTaskCleanup(task4.getId()).get().getCleanupType());
+    } finally {
+      configuration.setShuffleTasksForOverloadedSlaves(false);
+    }
+  }
+
+  protected void scheduleTask(String rqid, double requiredCpus, double requiredMemoryMb) {
```
ssalinas: we have a few shortcuts available already that should make the stuff in this method easier/shorter, though I'm impressed at all of the pieces you uncovered here :)

- `saveRequest` will take care of the requestManager.activate step you have below
- `initAndFinishDeployWithResources` will do most of the first section here
- `launchTask` can take the request/deploy, instance, and a state you want to get it to, as well as possible hostname args, and could be a possible shortcut for getting these tasks to an active state. A normal flow for a task is similar to what you had, where a pending request is created, drainPendingQueue turns that to a pending task, which then is launched with resource offers.

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityUsageTest.java
```diff
@@ -688,6 +706,249 @@ public void itLimitsTheNumberOfTaskCleanupsToCreate() {
...
+    deployManager.saveDeploy(rq, marker, dp);
+    deployManager.deletePendingDeploy(marker.getRequestId());
+    deployManager.saveDeployResult(marker, Optional.of(dp), new SingularityDeployResult(DeployState.SUCCEEDED));
+    deployManager.saveNewRequestDeployState(new SingularityRequestDeployState(marker.getRequestId(), Optional.of(marker), Optional.empty()));
+
+    requestManager.addToPendingQueue(pending);
+    scheduler.drainPendingQueue();
+  }
+
+  protected Map<String, Map<String, SingularityTaskId>> getTaskIdMapByHostByRequest() {
```
ssalinas: I don't know that we really need to build a map of this here. It seems in each case you are just using the getActiveTaskIds to then loop through all of them and get them into an active/starting state which can then be viewed by the usage poller. Can probably just do a for loop on taskManager.getActiveTaskIds() after maybe validating the size of the list

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityUsageTest.java
WH77: I really should've checked for review comments before I kept going, will change to use these methods.

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityUsageTest.java
WH77: I feel the map is more intuitive than looping through the task ids and matching host/request names, especially when working with multiple hosts, but it's definitely not something that needs to be built. I'd prefer to leave it, but will remove it if you prefer a simpler structure.

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityUsageTest.java
WH77: Since `launchTask` bypasses the normal task flow, it breaks/complicates testing of relocation of bounced tasks. I'm sure there's a way around this, that also makes these tests less horribly messy, but it's probably not worth the trouble.

ssalinas:
🚢

Issue #1641, opened 2017-11-02
Web interface JS caches forever (or just too long)


carlf:
I have found on upgrade, that a shift-reload is required to reload the javascript. When I do a curl, I find the following:

```
$ curl -I http://singularity/static/js/app.bundle.js
HTTP/1.1 200 OK
Date: Wed, 01 Nov 2017 23:55:10 GMT
Last-Modified: Fri, 04 Aug 2017 17:35:54 GMT
ETag: "17e67d2bec249a010b97c4af40cee9ad"
Content-Type: application/javascript
Content-Length: 2112229
```

That Last-Modified is earlier than when I upgraded to singularity 0.16.2 so it never gets loaded after an upgrade. It's not a huge deal but I usually field 10 or so users asking what's wrong with singularity after every update.

ssalinas:
Thanks for letting us know. We can take a look at how dropwizard is determining the Last-Modified header. I'm guessing it is based on the time the build was done. We can maybe update it to me something like the time the current singularity process started

PR #2058, opened 2020-01-15
Version SingularityUI assets

ssalinas:
This has been a problem for a while, that updates to the UI would take a hard refresh to load. We got around this internally by serving assets direct from nginx with slightly different structure and cache headers. However, this solves it so that things work direct from SingularityService.

gul-rev will add assets with a hash in their path, also templating that into the index file. These can essentially be cached forever since they will change when the UI code changes. Swagger ui assets are also now served direct from the backend so all assets are contained in SingularityService

/fixes https://github.com/HubSpot/Singularity/issues/1641


pschoenfelder:
🚢

PR #2059, opened 2020-01-16
Do not retry ON_DEMAND tasks which were killed by user request.

baconmania:


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -849,10 +849,19 @@ private boolean shouldRetryImmediately(SingularityRequest request, SingularityDe
...
+    if (task.isPresent()) {
+      Optional<SingularityTaskHistoryUpdate> taskHistoryUpdate = taskManager.getTaskHistoryUpdate(task.get().getTaskId(), ExtendedTaskState.TASK_CLEANING);
+
+      if (taskHistoryUpdate.isPresent()
+          && task.get().getTaskRequest().getPendingTask().getPendingTaskId().getPendingType() == PendingType.ONEOFF
+          && taskHistoryUpdate.get().getStatusMessage().orElse("").contains("USER_REQUESTED")) {
+        return false; // don't retry one-off launches of on-demand jobs if they were killed by the user
+      }
+
+      if (task.get().getTaskRequest().getPendingTask().getPendingTaskId().getPendingType() == PendingType.IMMEDIATE
```
ssalinas: small nit, can we do this check first for efficiency since it doesn't require the task history lookup?

baconmania:
Tested live successfully.

ssalinas:
🚢

PR #2060, opened 2020-01-16
Self Service Shuffle Opt Out

WH77:
Though all Singularity tasks should be able to handle bounces, some can't. Even if they can, it's understandable that users may want some high-importance tasks to avoid bouncing if possible. Therefore, we should allow users to indicate they don't want their tasks to be shuffled without having to edit the configuration and redeploy Singularity.

Concerns:
- Currently, it's possible for no tasks to be considered for shuffling on overloaded hosts. This is fine for a static configuration, but may change when we switch to a dynamic shuffle blacklist. Though leaving tasks off the blacklist by default is probably enough to keep Singularity in the clear, we should also refactor the shuffling logic to override the blacklist if absolutely necessary.
- Might be painful to add expiration to shuffle opt outs later.
- This PR adds a field to SingularityRequestParent.

TODO:
- [x] Unbreak SingularityUI.
- [x] Support shuffle opt out as an optional deployment flag.
- [x] Unit testing of the new manager.
- [x] Unit testing that task shuffling respects the dynamic shuffle blacklist.
- [x] Testing on staging.

cc - @ssalinas


SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequestParent.java
```diff
@@ -24,28 +24,32 @@
...
-                                  @JsonProperty("requestDeployState") Optional<SingularityRequestDeployState> requestDeployState,
-                                  @JsonProperty("activeDeploy") Optional<SingularityDeploy> activeDeploy,
-                                  @JsonProperty("pendingDeploy") Optional<SingularityDeploy> pendingDeploy,
-                                  @JsonProperty("pendingDeployState") Optional<SingularityPendingDeploy> pendingDeployState,
-                                  @JsonProperty("expiringBounce") Optional<SingularityExpiringBounce> expiringBounce,
-                                  @JsonProperty("expiringPause") Optional<SingularityExpiringPause> expiringPause,
-                                  @JsonProperty("expiringScale") Optional<SingularityExpiringScale> expiringScale,
-                                  @JsonProperty("expiringSkipHealthchecks") Optional<SingularityExpiringSkipHealthchecks> expiringSkipHealthchecks,
-                                  @JsonProperty("taskIds") Optional<SingularityTaskIdsByStatus> taskIds) {
+  public SingularityRequestParent(
```
ssalinas: Few things here:
- If we are changing this constructor, we'd need to also keep the old one (minus annotations) so that we don't break anyone's builds
- This field could likely default to `false` rather than being an Optional (we made that mistake in a few places and have tried to not do it for net new stuff)
- I'm not sure that we need to go through adding this to the request parent object. If this is just for the UI, we may want to just make a separate api call to load up the blacklist (Which should always be a small amount of data. The SingularityRequestParent object is build and used in a lot of places, some on expensive paths where speed matters. So, adding one additional zk call to every one of them could really add up. The alternative of making a small/cheap API call from the UI seems more favorable

SingularityService/src/main/java/com/hubspot/singularity/resources/ShuffleConfigurationResource.java
```diff
@@ -0,0 +1,117 @@
...
+
+    if (request.isPresent()) {
+      authorizationHelper.checkForAuthorization(request.get().getRequest(), user, scope);
+    } else {
+      throw WebExceptions.notFound("No request matching id %s", requestId);
+    }
+  }
+
+  @GET
+  @Path("/shuffle/config/blacklist")
```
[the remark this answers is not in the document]
WH77: thanks for catching that, pretty sure I actually used /shuffle/blacklist in the frontend

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityUsageHelper.java
```diff
@@ -286,11 +293,14 @@ private SingularityTaskUsage getUsage(MesosTaskMonitorObject taskUsage) {
   }
 
 
-  private boolean isEligibleForShuffle(SingularityTaskId task) {
+  private boolean isEligibleForShuffle(SingularityTaskId task, Set<String> requestBlacklist) {
     Optional<SingularityTaskHistoryUpdate> taskRunning = taskManager.getTaskHistoryUpdate(task, ExtendedTaskState.TASK_RUNNING);
 
+    // requestBlacklist = dynamic list from zookeeper
+    // configuration = constant list from yaml file
```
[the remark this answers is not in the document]
WH77: Yeah, as is the split leaves the source of truth unclear. I was considering treating ZK as a "soft" blacklist and the configuration file as a "hard" blacklist, but that's non-intuitive and likely more trouble than it's worth.

SingularityService/src/main/java/com/hubspot/singularity/resources/ShuffleConfigurationResource.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+import java.util.Optional;
+
+@Path(ApiPaths.STATE_RESOURCE_PATH)
```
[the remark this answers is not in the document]
WH77: I'm not surprised I forgot to change this, just disappointed.

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -815,10 +815,10 @@ public SingularityRequestBatch getRequestsBatch(
   public List<SingularityRequestParent> getFinishedRequests(
       @Parameter(hidden = true) @Auth SingularityUser user,
       @Parameter(description = "Fetched a cached version of this data to limit expensive operations") @QueryParam("useWebCache") Boolean useWebCache,
-                                                            @QueryParam("filterRelevantForUser") Boolean filterRelevantForUser,
```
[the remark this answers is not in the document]
WH77: My bad, good to know for future work. I'll try to recover the original formatting, if possible.

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequestParent.java
WH77: Yeah, it definitely felt dirty as I did it. Though I agree that the overhead isn't worth it for this feature and will change it, counterargument just in case - attaching this information to SingularityRequestParent keeps this more consistent with similar request attributes accessible through the UI, like skipping healthchecks.

For the first bullet, I assumed that this class was a wrapper Singularity put around raw SingularityRequest instances to attach its own metadata for the UI. If it's used outside this project or on critical paths, then I'm definitely switching back to the UI making direct API calls to the shuffle resource.

ssalinas: Most clients just receive this object, but there are plenty of places where it could be constructed in things like unit tests. As for the first argument, many of those fields are used in things that are a bit more on the critical path for launching tasks, while shuffle is really only relevant to that single usagepoller

SingularityService/src/main/java/com/hubspot/singularity/resources/ShuffleConfigurationResource.java
```diff
@@ -0,0 +1,115 @@
...
+      RequestManager requestManager
+  ) {
+    super(httpClient, leaderLatch, objectMapper);
+
+    this.shuffleCfgManager = shuffleCfgManager;
+    this.authorizationHelper = authorizationHelper;
+    this.requestManager = requestManager;
+  }
+
+  private void checkReadAuthorization(SingularityUser user, String requestId) {
```
ssalinas: FYI, authorizationHelper has a `checkForAuthorizationByRequestId` which will do the work of these private methods for you

SingularityService/src/test/java/com/hubspot/singularity/data/ShuffleConfigurationManagerTest.java
```diff
@@ -0,0 +1,63 @@
...
+    Assertions.assertTrue(shuffleCfgManager.isOnShuffleBlacklist(r1));
+    Assertions.assertTrue(shuffleCfgManager.isOnShuffleBlacklist(r2));
+    Assertions.assertTrue(shuffleCfgManager.getShuffleBlacklist().contains(r1));
+    Assertions.assertTrue(shuffleCfgManager.getShuffleBlacklist().contains(r2));
+  }
+
+  @Test
+  public void itSupportsRemovingBlacklistedRequestIds() {
+    shuffleCfgManager.addToShuffleBlacklist(r1);
+    shuffleCfgManager.removeFromShuffleBlacklist(r1);
```
ssalinas: small nit, this test is likely more useful if it asserts that the added r1 is present before removing it and asserting it is gone (i.e. the add may fail and this test could still pass)

SingularityUI/app/components/requestForm/RequestForm.jsx
```diff
@@ -628,6 +630,15 @@ const RequestForm = (props) => {
     />
   );
 
+  const avoidShuffle = (
```
ssalinas: One note on placement of this. We will likely want to make a button+modal for this in the top bar of the request detail page (example below) rather than putting it on the request form, for two reasons:
1) We hide the request form from developers in most cases, since, internally, the request json is managed by our deploy api
2) Given that the request form is meant to essentially represent the request json, and shuffle is under a separate root/api/object, it makes sense to separate shuffle a bit

side note, a nice to have, but maybe we could actually show the button on the task detail page as well when the task in question was killed due to shuffling? Would have to identify that based on the cleanup message in the status updates

WH77: I've implemented the side note, but to avoid complicating the redux flow for the TaskDetail page the button will always prompt the user to disable shuffling for the request instead of being a toggle - no technical issues (safe to post the same request id to the blacklist multiple times), but maybe a slight source of confusion.

SingularityUI/app/components/common/modalButtons/ShuffleOptOutModal.jsx
```diff
@@ -0,0 +1,76 @@
...
+    return (
+      <FormModal
+        name={isOptedOut ? "Enable Shuffling For Request" : "Disable Shuffling For Request"}
+        ref="shuffleOptOutModal"
+        action={isOptedOut ? "Enable Shuffling" : "Disable Shuffling"}
+        onConfirm={(data) => this.confirm(data)}
+        buttonStyle="primary"
+        formElements={[]}>
+        <p>{this.renderActionText()} shuffling for {requestIds.length > 1 ? 'these' : 'this'} request{requestIds.length > 1 && 's'}?</p>
+        <p>Note that Singularity may still bounce your tasks if necessary, such as during host failures.</p>
```
ssalinas: Two things, we can probably add a first note that this is specifically 'Disable task shuffling due to memory pressure or cpu overusage', with a follow up of 'Note that this does not prevent all restarts of your task. A task may still be bounced or shuffled for events such as host decommissions'

WH77: image

ssalinas: Looks good, let's give this a go in QA 👍

ssalinas:
🚢

SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java
```diff
@@ -18,6 +18,7 @@
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
+import javax.ws.rs.HEAD;
```
baconmania: Leftover from a merge conflict?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskShuffler.java
```diff
@@ -0,0 +1,358 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class SingularityTaskShuffler {
```
baconmania: +1 for splitting this out

PR #2061, opened 2020-01-27
Properly reconnect after losing mesos master connection

ssalinas:
This will fix reconnects in a situation where all masters are temporarily unavailable. Currently if the reconnect ends up throwing a ConnectException in the subscriber thread, that calls onConnectException which, correctly, starts the process over again. However, because this is still synchronously called on the subscriber thread, when the reconnect tries to interrupt that thread, waiting for the join then throws an InterruptedException (because it is called from the same thread being interrupted)

Moving the reconnect to a separate thread should solve this problem. It is currently a single thread, as no more than one of them should be running at a time anyway

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/MesosConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/MesosConfiguration.java
@@ -77,7 +77,7 @@ public class MesosConfiguration {
   private Optional<String> mesosUsername = Optional.empty();
   private Optional<String> mesosPassword = Optional.empty();
 
-  private long reconnectTimeoutMillis = 60000;
+  private long reconnectTimeoutMillis = 120000;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/config/MesosConfiguration.java
```diff
@@ -77,7 +77,7 @@
   private Optional<String> mesosUsername = Optional.empty();
   private Optional<String> mesosPassword = Optional.empty();
 
-  private long reconnectTimeoutMillis = 60000;
+  private long reconnectTimeoutMillis = 120000;
```
ssalinas: This is lengthened because the attempt time now takes into account the startup period, not just the time from POST -> subscribed event like it did before

pschoenfelder:
🚢

PR #2063, opened 2020-01-29
Debounce UI Request/Task Search

WH77:
Should make the request/task search filters feel much snappier, though there's still occasionally perceptible amounts of lag for very short queries. I've tested it against the QA Singularity backend, can DM you a gif if you feel like it.

Tangential issue, seems like #2058 broke `gulp serve` - this PR fixes that as well, though I'm not proud of how long it took me to realize what was the issue.

cc - @ssalinas

### DIFF
--- a/SingularityUI/app/components/requests/RequestFilters.jsx
+++ b/SingularityUI/app/components/requests/RequestFilters.jsx
@@ -78,6 +78,16 @@ export default class RequestFilters extends React.Component {
     }
   ];
 
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      filter: _.extend({}, this.props.filter),
+    };
+
+    this.onFilterChange = _.debounce(this.props.onFilterChange, 250);
+  }
+
   handleStatusSelect(selectedKey) {
     this.props.onFilterChange(_.extend({}, this.props.filter, {state: RequestFilters.REQUEST_STATES[selectedKey].filterVal}));
   }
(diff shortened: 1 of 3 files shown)


SingularityUI/app/components/requests/RequestFilters.jsx
```diff
@@ -78,6 +78,16 @@ export default class RequestFilters extends React.Component {
...
+
+    this.state = {
+      filter: _.extend({}, this.props.filter),
+    };
+
+    this.debouncedOnFilterChange = _.debounce(this.props.onFilterChange, 250);
+  }
+
   handleStatusSelect(selectedKey) {
     this.props.onFilterChange(_.extend({}, this.props.filter, {state: RequestFilters.REQUEST_STATES[selectedKey].filterVal}));
```
WH77: left the other filter handlers as is because they're not causing any issues, though it does make the code a bit inconsistent

SingularityUI/gulpfile.js
```diff
@@ -72,6 +72,9 @@ gulp.task('static', ['clean'], function() {
 })
 
 gulp.task('debug-html', ['static'], function () {
+  templateData['appJsPath'] = 'js/app.bundle.js'
```
ssalinas: can't believe I forgot this in the other PR 🤦‍♂

ssalinas:
Yeah, this feels way snappier in staging. I know the filters are still a bit inconsistent, but the user experience is definitely a huge improvement here. So 🚢 for this one, and we can come back to cleaning up the filters later if needed, or if we have another round of changes for them

mjball:
Thank you! This was a very noticeable improvement for me. I really appreciate it.

PR #2064, opened 2020-01-29
Prevent request action modals from closing upon regaining focus

WH77:
Fixes a bug introduced by #2060, which caused all action modals to close when a user regained focus on the request detail page. TLDR - the shuffle opt out status for a request should only be fetched on initialization of the details page, not as part of the existing refresh process, to avoid forcing reconstruction (which resets modal visibility) of the request action buttons and their child modals.

cc - @ssalinas


ssalinas:
Thanks for the follow up PR. Let's test in staging/qa 👍

ssalinas:
🚢

PR #2068, opened 2020-02-03
Consistently compare memory in GB, not MB/B

ssalinas:
Came across this when looking at other things. Looks like the cpu/mem ratio to determine if a slave/request was high mem or high cpu was comparing bytes in one spot and MB in another. This should compare everything in GB instead, which also makes the default cutoffs of 1.5 and 0.5 better. Should be the case that (for AWS at least) m5 is AVERAGE, r5 is high mem, c5 is high cpu


pschoenfelder:
:ship:

PR #2069, opened 2020-02-05
Fix run now modal not being able to tail task launch

WH77:
More fun caused by Redux, seems like the run now modal was being re-rendered due to a transitive dependency on the shuffle opt out fetch. This PR removes that dependency, and should prevent similar issues for other modals.

cc - @ssalinas


ssalinas:
🚢

PR #2070, opened 2020-02-07
Less Terminal

WH77:
Adds experimental support for a shell-like less terminal in the task file browser, behind a new configuration flag.

cc - @ssalinas, @baconmania


SingularityUI/package.json
```diff
@@ -62,7 +62,10 @@
     "singularityui-tailer": "0.3.13",
     "typeahead.js": "^0.11.1",
     "underscore": "~1.8.0",
-    "uuid": "^3.0.0"
+    "uuid": "^3.0.0",
+    "xterm": "^4.3.0",
+    "xterm-addon-attach": "^0.4.0",
```
WH77: this dependency currently breaks `gulp build` due to a stray typescript file in the package causing a syntax error in uglifyjs

ssalinas:
One overall thing we need to discuss in this case is that we hadn't initially planned on open sourcing the shel socket backend, so we need to determine if this UI component is best suited here or somewhere else. Can chat offline about that as well

SingularityUI/app/assets/index.mustache
```diff
@@ -53,7 +53,8 @@
             authCookieName: "{{{ authCookieName }}}",
             authTokenKey: "{{{ authTokenKey }}}",
             quickLinks: {{{quickLinks}}},
-            navTitleLinks: {{{navTitleLinks}}}
+            navTitleLinks: {{{navTitleLinks}}},
+            lessTerminalPort: {{{lessTerminalPort}}}
```
ssalinas: quick note for this. To fully wire it up to Singularity backend, we'll have to put something in maybe the [UI Config](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/config/UIConfiguration.java) that is then wired to the [IndexView](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/views/IndexView.java). That will wire the value through from the Singularity config yaml through the the frontend config

PR #2071, opened 2020-02-11
Better exposure of LB status in UI

ssalinas:
The Singularity backend knows if a task has been added to the load balancer or not. We should better expose this in the health icon in the UI to show if a task is currently serving traffic or not. Still debating if this should be a separate column. Didn't want to make the active tasks table too much wider, so for the moment I've tried to see if we can encode this all into the existing health icon in a way that is easy enough to understand

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdsByStatus.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdsByStatus.java
@@ -33,13 +33,6 @@ public class SingularityTaskIdsByStatus {
     this.killed = killed != null ? killed : Collections.emptyList();
   }
 
-  public SingularityTaskIdsByStatus(List<SingularityTaskId> healthy,
-                                    List<SingularityTaskId> notYetHealthy,
-                                    List<SingularityPendingTaskId> pending,
-                                    List<SingularityTaskId> cleaning) {
-    this(healthy, notYetHealthy, pending, cleaning, Collections.emptyList(), Collections.emptyList());
-  }
-
   @Schema(description = "Active tasks whose healthchecks and load balancer updates (when applicable) have finished successfully")
   public List<SingularityTaskId> getHealthy() {
     return healthy;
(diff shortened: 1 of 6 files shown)


SingularityUI/app/components/requestDetail/ActiveTasksTable.jsx
```diff
@@ -45,12 +45,34 @@ const ActiveTasksTable = ({request, requestId, tasksAPI, healthyTaskIds, cleanin
 
   const tasksWithHealth = _.map(tasks, (task) => {
```
WH77: could move this to mapStateToProps for cleanliness, but given that tasks might still be loading there I think it's fine to leave this as is

SingularityService/src/main/java/com/hubspot/singularity/helpers/RequestHelper.java
```diff
@@ -319,15 +310,27 @@ public void updateRequest(SingularityRequest request, Optional<SingularityReques
...
     }
 
-    return Optional.of(new SingularityTaskIdsByStatus(healthyTaskIds, notYetHealthyTaskIds, pendingTaskIds, cleaningTaskIds));
+    List<SingularityTaskId> loadBalanced = new ArrayList<>();
+    if (requestWithState.getRequest().isLoadBalanced()) {
+      healthyTaskIds.stream()
+          .filter(taskManager::isInLoadBalancer)
+          .forEach(loadBalanced::add);
+      cleaningTaskIds.stream()
+          .filter(taskManager::isInLoadBalancer)
```
WH77: does the task manager automatically update itself when tasks go in and out of the load balancer?

SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
```diff
@@ -364,24 +364,6 @@ public SingularityCreateResult savePendingTask(SingularityPendingTask task) {
     return getChildrenAsIdsForParents("getAllTaskIds", paths, taskIdTranscoder);
   }
 
-  private List<SingularityTaskId> getTaskIds(String root) {
```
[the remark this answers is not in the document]
ssalinas: yeah all unused

SingularityService/src/main/java/com/hubspot/singularity/helpers/RequestHelper.java
ssalinas: yeah, the isInLoadBalancer here is a live call to fetch the state (cached on the leader in memory)

SingularityBase/src/main/java/com/hubspot/singularity/SingularityTaskIdsByStatus.java
```diff
@@ -14,16 +15,29 @@
...
     this.cleaning = cleaning;
+    this.loadBalanced = loadBalanced != null ? loadBalanced : Collections.emptyList();
+    this.killed = killed != null ? killed : Collections.emptyList();
+  }
+
+  public SingularityTaskIdsByStatus(List<SingularityTaskId> healthy,
+                                    List<SingularityTaskId> notYetHealthy,
+                                    List<SingularityPendingTaskId> pending,
+                                    List<SingularityTaskId> cleaning) {
+    this(healthy, notYetHealthy, pending, cleaning, Collections.emptyList(), Collections.emptyList());
```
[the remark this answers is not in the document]
ssalinas: Was debating that, figured I'd tend towards treating as immutable. In most places in the code we try and do that, and make copies fo lists where appropriate if we are going to build on them

WH77:
🚢

PR #2072, opened 2020-02-11
Quiet down some exceptions that don't matter

ssalinas:
These all get retried or are exposed in other places already and don't need to spam sentry


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
```diff
@@ -245,9 +246,13 @@ private int checkUploads() {
           toRemove.add(entry.getKey());
         }
       } catch (Throwable t) {
-        metrics.error();
-        LOG.error("Waiting on future", t);
-        exceptionNotifier.notify(String.format("Error waiting on uploader future (%s)", t.getMessage()), t, ImmutableMap.of("metadataPath", uploader.getMetadataPath().toString()));
+        if (t instanceof NoSuchFileException) {
```
WH77: can this be a separate catch block / is there a reason why you prefer to use instanceof checks instead?

ssalinas: fixed, thanks! was copy/pasting things around and didn't even notice

PR #2078, opened 2020-02-26
Expose host sandbox path

ssalinas:
Needed for communication with s3uploader from docker tasks. The uploader needs to know the full path to the sandbox, not a relative path within the container


WH77:
🚢

PR #2081, opened 2020-03-12
Fix `logrotateadditionalfiles` deletion.

baconmania:
This PR aims to fix a scenario whereby `logrotateAdditionalFiles` marked with `deleteInExecutorCleanup` would never actually be deleted. The scenario was:

- Task terminates successfully
- Executor runs the “immediate” cleanup logic
- `logrotateAdditionalFiles` deletion is skipped because it hasn't been 15 minutes since task termination
- Cleanup logic flows all the way through to https://github.com/HubSpot/Singularity/blob/4dea95764bdeee9f41046b4ad1c6a94a1e5dab8c/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskCleanup.java#L79
- Task definition file is deleted
- Scheduled executor cleanup will never be aware of this sandbox again

I’ve plumbed through some logic for log cleanup which is similar to the logic for the sandbox `app/` directory cleanup. Which is to say, if it hasn’t been 15 minutes since task termination, we return a `WAITING` state from the “immediate” cleanup process. This way, we allow scheduled cleanup to take another look at it later, and properly delete files which we had been logrotating for the lifetime of the task.

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTask.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTask.java
@@ -84,10 +84,11 @@ public class SingularityExecutorTask {
     ExtendedTaskState extendedTaskState = MesosUtils.fromTaskState(org.apache.mesos.v1.Protos.TaskState.valueOf(state.toString()));
 
     boolean cleanupAppTaskDirectory = !extendedTaskState.isFailed() && !taskDefinition.getExecutorData().getPreserveTaskSandboxAfterFinish().orElse(Boolean.FALSE);
+    boolean cleanupLogs = false; // Task has just died, so we don't want to delete logs yet.
 
     boolean isDocker = (taskInfo.hasContainer() && taskInfo.getContainer().hasDocker());
 
-    taskCleanup.cleanup(cleanupAppTaskDirectory, isDocker);
+    taskCleanup.cleanup(cleanupAppTaskDirectory, cleanupLogs, isDocker);
   }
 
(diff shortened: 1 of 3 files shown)


baconmania:
Note that there are, broadly, three things to be cleaned up:
1. app dir
2. log files uploaded by Singularity (`service.log` and `s3UploaderAdditionalFiles`)
3. log files _rotated_ by Singularity

With this PR, we're set up so that 2 and 3 are either performed or skipped _together_. Is this what we want? Should 1 and 2 happen together, with the 15 minute delay applying only to 3?

ssalinas:
I think I agree that the 15 minute delay should apply to both 2 and 3. In particular because the S3Uploader needs a chance to get a at the logs for 2, which the 15 minute window is nice for.

Overall PR seems fine. I think it's gonna be a matter of testing the common cases in staging. Good find on where we clean the task definition file. Couldn't remember earlier where we actually were informed about a sandbox still hanging around

baconmania:
Tested successfully. The scenario described above now proceeds as follows:

- Task terminates successfully
- Executor runs the “immediate” cleanup logic
- `app/` directory is cleaned
- All configured logfiles that are configured to be rotated, are rotated (and therefore, `tail_of_finished_service.log` is available approximately immediately, same as current `master` behavior)
- `logrotateAdditionalFiles` deletion is skipped because it hasn't been 15 minutes since task termination
- The immediate cleanup process returns a `WAITING` state and the task definition file isn't deleted
- ... 15 minutes later...
- `logrotateAdditionalFiles` deletion proceeds
- Task definition file is deleted
- Scheduled cleanup process returns `SUCCESS`

ssalinas:
🚢

PR #2083, opened 2020-03-19
S3 uploader + executor cleanup efficiency improvements

ssalinas:
In looking through the code, there was a path that could cause us to build up an unnecessary amount of open uploaders if a large count of failed tasks ended up on the same host. In the current cleanup setup, because we were preserving the task app directory, we would short circuit before hitting any lines that finalized uploaders, allowing them to run to completion in the S3Uploader. Since the task is finished, there is no reason to keep the uploader files open

This also cleans up a few noisier exceptions and log statements that aren't adding a ton of value across the different runnables

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskCleanup.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskCleanup.java
@@ -72,25 +72,24 @@ public class SingularityExecutorTaskCleanup {
       return finishTaskCleanup(dockerCleanSuccess);
     }
 
-
-    if (!cleanupTaskAppDirectory) {
-      log.info("Not finishing cleanup because taskApp directory is being preserved");
-      return TaskCleanupResult.WAITING;
-    }
-
-    boolean cleanupTaskAppDirectorySuccess = cleanupTaskAppDirectory();
     boolean logTearDownSuccess = taskLogManager.teardown();
-
-    log.info("Cleaned up task app directory ({}) and rotated logs ({})", cleanupTaskAppDirectorySuccess, logTearDownSuccess);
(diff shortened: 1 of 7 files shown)


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityUploader.java
```diff
@@ -156,6 +158,8 @@ int upload(boolean isFinished) throws IOException {
           throw new RuntimeException(ioe);
         }
       });
+    } catch (UncheckedIOException|NoSuchFileException nsfe) {
+      LOG.debug("Parent file {} did not exist, skipping", directory);
```
ssalinas: This is because some directories we scanned contain tmp files, so even though we checked for the directory directly above this, it's possible it got deleted out from under us before we started the filesystem walk

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/task/SingularityExecutorTaskCleanup.java
```diff
@@ -72,26 +72,25 @@ public TaskCleanupResult cleanup(boolean cleanupTaskAppDirectory, boolean cleanu
...
     }
 
-
-    if (!cleanupTaskAppDirectory) {
-      log.info("Not finishing cleanup because taskApp directory is being preserved");
-      return TaskCleanupResult.WAITING;
-    }
-
-    boolean cleanupTaskAppDirectorySuccess = cleanupTaskAppDirectory();
     boolean logTearDownSuccess = taskLogManager.teardown();
```
ssalinas: This was the important line to move @baconmania . This shouldn't delete any log files, so the previous logging PR is still in effect. But it moves the task app directory cleanup to the end, so that we don't wait forever to clean logs just because we haven't cleaned taskapp yet

pschoenfelder:
🚢

PR #2085, opened 2020-03-20
Periodically reconcile the list of metadata files

ssalinas:
This is the simplest fix I could think of for our longer-term uploader leak issue. The most likely case is that we are occasionally missing a file update or not handling ti properly, since we always end up with more files in the filesystem than in memory. So, this periodically rescans the directory just like on startup. For the most part it will be a no-op, but will also serve to clean up files we may have missed so that things don't grow out of control

### DIFF
--- a/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
+++ b/SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
@@ -82,6 +82,7 @@ public class SingularityS3UploaderDriver extends WatchServiceHelper implements S
   private final Map<S3UploadMetadata, CompletableFuture<Integer>> immediateUploadersFutures;
 
   private ScheduledFuture<?> future;
+  private ScheduledFuture<?> fileSyncFuture;
 
   @Inject
   public SingularityS3UploaderDriver(SingularityRunnerBaseConfiguration baseConfiguration, SingularityS3UploaderConfiguration configuration, SingularityS3Configuration s3Configuration,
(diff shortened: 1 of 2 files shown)


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
```diff
@@ -187,6 +188,17 @@ public void startAndWait() {
       }
     }, configuration.getCheckUploadsEverySeconds(), configuration.getCheckUploadsEverySeconds(), TimeUnit.SECONDS);
 
+    fileSyncFuture = scheduler.scheduleAtFixedRate(() -> {
+      runLock.lock();
+      try {
+        readInitialFiles();
```
WH77: checking - do you know how long this process usually takes, and is there a concern that it'll hold the lock long enough to cause issues for the S3 uploads?

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
```diff
@@ -209,6 +221,10 @@ public void shutdown() {
       runLock.unlock();
     }
 
+    if (fileSyncFuture != null) {
+      fileSyncFuture.cancel(true);
```
WH77: why allow interrupts here but not for the upload future?

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
ssalinas: This whole process is async/background, nothing hits it live or like a service, and it only holds the lock for a second or two

SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityS3UploaderDriver.java
ssalinas: Good Q, I don't remember why I put that there... Will take a second look at it

ssalinas: Ah right, this one is just checking for new files, interrupting in the middle won't actually interrupt an upload. The uploaders themselves we want to give a better chance to finish

pschoenfelder:
🚢

PR #2086, opened 2020-03-24
Configurable variables for docker container to inherit from executor

ssalinas:
Mesos allows setting default variables. These will be set for each task, but for the executor process. Because our executor manages docker container creation itself, we need to manually move some of these mesos-provided or system level variables to the docker container. This adds a configurable set of variables to pass from the executor wrapper script to the container env file


pschoenfelder:
🚢

PR #2088, opened 2020-04-17
Allow service module override

ssalinas:
Allows us to do things like add in an override module or override the db connections. WIP


WH77:
🚢

SingularityService/src/main/java/com/hubspot/singularity/SingularityService.java
```diff
@@ -78,7 +82,7 @@ public void initialize(final Bootstrap<T> bootstrap) {
 
     guiceBundle = GuiceBundle.defaultBuilder(SingularityConfiguration.class)
         .modules(
-            new SingularityServiceModule(),
+            getServiceModule(),
             getObjectMapperModule(),
             new SingularityAuthModule()
```
WH77, line 87: might want to add a get method for this module as well, for consistency.

ssalinas: 👍 I have some groups/scopes work coming up soon, will likely do that in a future PR

PR #2089, opened 2020-04-28
Uncaught exception should result in failed deploy

ssalinas:
...instead of causing the whole poller to error and trigger abort


tpetr:
:ship:

PR #2090, opened 2020-04-30
Enable prettier java

ssalinas:
Yes it's opinionated, but much easier to have automatic formatting. `mvn prettier:write` will take care of formatting anything for those that don't have the intellij plugin. Will resolve conflicts on the othe branches with this in a bit, but it's a good time to do it since we don't ahve any large pending PRs


WH77:
🚢

PR #2093, opened 2020-05-19
Pin xterm versions to avoid compatibility issues

WH77:
Avoid issues by specifying the set of versions from https://github.com/xtermjs/xterm.js/releases/tag/4.6.0.

cc - @ssalinas


ssalinas:
🚢

PR #2094, opened 2020-05-19
Render page while fetching history tables

pschoenfelder:
History tables can sometimes be slow to fetch, so we should wait to load the entire page.

My frontend knowledge here is minimal, but I think taking advantage of this should take care of the issue: https://github.com/HubSpot/Singularity/blob/timeout/SingularityUI/app/components/common/table/UITable.jsx#L439


ssalinas:
@pschoenfelder this will give those pages nicer loading icons, and we can keep that However, I think the overall issue is more here:
https://github.com/HubSpot/Singularity/blob/master/SingularityUI/app/components/requestDetail/RequestDetailPage.jsx#L47-L90

in mapStateToProps there, we set the `deleted` boolean to true if the request returns a 404 or the history state for the request is empty. I think the Utils.maybe code in there will return an empty list for things if the request hasn't returned anything and populated the state yet. So, the active requests table doesn't try to render. We can probably have that condition be just the 404 and make sure each table can handle the empty table state properly

ssalinas:
🚢

PR #2096, opened 2020-05-22
Fix less tailer URLs for large files

WH77:
Several log links didn't take file size into account when they should have.

cc - @ssalinas


ssalinas:
🚢

PR #2097, opened 2020-05-28
Sandbox permissions update should be recursive

ssalinas:
At the moment we do a chown of `$MESOS_SANDBOX`. This is set by mesos agent to the root of the sandbox. However, most operations on the sandbox are in the directory named after the task id, and that ends up still owned by root. This will make all directories under the sandbox owned by the correct user


WH77:
🚢

PR #2098, opened 2020-06-02
Add button for manual leader failover

ssalinas:
Ends up being a common debugging step if other things are not responding. Adding an admin-only button for it so it can be triggered remotely


pschoenfelder:
🚢

PR #2100, opened 2020-06-04
Additional deploy validation

ssalinas:
Two main pieces:
- Add the ability to lock down artifact specification to only signed artifacts (s3 + signature file). This does not validate everything within a list, assuming that if the list is signed, the things within the list are valid
- Add valid docker registry list so Singularity can lock usage down to only an internal docker registry if needed

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
@@ -683,7 +683,11 @@ public class SingularityValidator {
         String image = deploy.getContainerInfo().get().getDocker().get().getImage();
         checkBadRequest(
           validDockerRegistries.contains(URI.create(image).getHost()),
-          String.format("%s does not point to an allowed docker registry", image)
+          String.format(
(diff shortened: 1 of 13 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -638,20 +638,63 @@ public SingularityDeploy checkDeploy(
...
+
+    if (!validDockerRegistries.isEmpty()) {
+      if (
+        deploy.getContainerInfo().isPresent() &&
+        deploy.getContainerInfo().get().getDocker().isPresent()
+      ) {
+        String image = deploy.getContainerInfo().get().getDocker().get().getImage();
+        checkBadRequest(
+          validDockerRegistries.contains(URI.create(image).getHost()),
+          String.format("%s does not point to an allowed docker registry", image)
```
PaulFurtado, line 686: Maybe include the allowed registries in the error message to make this obvious?

pschoenfelder:
🚢

PR #2101, opened 2020-06-09
Add ability to auth to mesos http endpoints

ssalinas:
Ability to still hit mesos master api endpoints when authenticate_http_readonly and authenticate_http_readwrite flags are set. Uses the same credentials as specified in MesosConfiguration that are used to register the framework (e.g. Singularity just needs a single principal with ability to register and to hit the read/write apis)


pschoenfelder:
🚢

PR #2103, opened 2020-06-11
Agent rename

ssalinas:
At the moment this is only a surface rename, but at least updates the most public-facing pieces. The openapi json descriptions, docs, and ui menus/titles have been updated from slave -> agent. There is obviously more we can do, but migrating the internal java bits is a much larger undertaking if we want to do it in a backwards compatible way.


pschoenfelder:
🚢

ssalinas:
docs and ui router fixed now

pschoenfelder:
🚢

PR #2105, opened 2020-06-11
Force index for task history requests by host and start/update times

fabatef:
Forces task history search queries to use the `hostUpadtedAt` index for requests that only specify `host` and `startedAt/updatedAt` fields. These queries currently timeout intermittently since the `hostUpdatedAt` index isn't consistently chosen by the optimizer.


ssalinas:
Mentioned this to @baconmania , but documenting here as well. If we are adding something non-postgres-compatible, we should make this into an abstract method instead. It can be implemented only in the mysql version and essentially a no-op on the postgres side

ssalinas:
@baconmania will let you review here

ssalinas:
@fabatef can we get the merge conflicts resolved here at some point. Realized this one never actually made it through to stable

ssalinas:
🚢

PR #2106, opened 2020-06-15
Specify which artifacts are missing signature in error

ssalinas:
Providing a better detailed error message. Also had one condition swapped that is now fixed


pschoenfelder:
🚢

PR #2107, opened 2020-06-16
Only keep the most recent 50 failure events

ssalinas:
For very long lived deploys in a constant crash loop, the statistics node can essentially grow out of control, which will eventually cause Singularity to reach its jute buffer limit. This will trim it to the most recent 50, which is more than enough for the cooldown and crash loop checks


pschoenfelder:
🚢

PR #2108, opened 2020-06-29
Track shutdown time in status message

ssalinas:
Right now this data is only available in the executor logs, which disappear after being cleaned up by mesos. This adds the time between a kill being sent and the process shutting down in the status message for easier viewing


pschoenfelder:
🚢

PR #2109, opened 2020-07-01
Wire up a `size` override for logrotateAdditionalFiles

baconmania:


ssalinas:
Overall looks good. Want to see it in action on staging then can give the ⛵

ssalinas:
🚢

PR #2111, opened 2020-07-06
Bump version of Apache Mesos to 1.9.0

ajammala:


mesos-docker/mesos-master/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM hubspot/mesos:1.0
```
ssalinas, line 1: Can we plan to tag this after whatever the version of mesos we are building into it is? e.g. 1.9.0 or 1.10.0? Will keep it consistent with the other images and patterns we used to use

pom.xml
```diff
@@ -74,7 +74,7 @@
     <dropwizard.version>1.3.12</dropwizard.version>
     <horizon.version>0.1.2</horizon.version>
     <mesos.docker.tag>1.7.1</mesos.docker.tag> <!-- 1.8.0 not available yet, still compatible -->
```
ssalinas, line 76: This is the tag we'd have to update for the SingularityExecutor/Dockerfile one. If I remember correctly, this controls the output tag from the maven build

PR #2112, opened 2020-07-08
Upgrading apache mesos version

ajammala:
Bump version for apache mesos and create docker containers for apache mesos v 1.9.0


ssalinas:
@ajammala gonna close this one since it had all the extra stuff in it

PR #2113, opened 2020-07-08
More complete info level logging on offers

ssalinas:
summarize overloaded hosts, also turns down the preferred host lines to debug to save some space


pschoenfelder:
🚢

PR #2116, opened 2020-07-13
Add warning about override and lb updates

ssalinas:
The kill -9 bypasses LB updates, which was a surprise to some users. Making it more obvious in the modal


pschoenfelder:
🚢

PR #2117, opened 2020-07-15
Filter getRequestHistory queries by createdAt times

fabatef:
Instead of looping through the entire request history for some `requestId` to find the `SingularityRequestHistory`s in a specific time range, this change allows us to get the request history entries filtered by `createdBefore` and `createdAfter` times via [SingularityClient#getHistoryForRequest ](https://github.com/HubSpot/Singularity/compare/get-request-history-by-createAt-times?expand=1#diff-33d09d764454092917dadaf4d4f34cd6L1679)

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/history/PostgresHistoryJDBI.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/history/PostgresHistoryJDBI.java
@@ -2,7 +2,6 @@ package com.hubspot.singularity.data.history;
 
 import com.hubspot.singularity.SingularityDeployHistory;
 import com.hubspot.singularity.SingularityRequest;
-import com.hubspot.singularity.SingularityRequestHistory;
 import com.hubspot.singularity.SingularityTaskHistory;
 import com.hubspot.singularity.data.history.SingularityMappers.SingularityRequestIdCount;
 import java.util.Date;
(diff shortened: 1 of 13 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/history/PostgresHistoryJDBI.java
```diff
@@ -253,5 +241,10 @@ void setDeployJson(
     @Bind("json") @Json SingularityDeployHistory deployHistory
   );
 
+  @Override
+  default String getRequestHistoryBaseQuery() {
```
ssalinas, line 245: If this is just a constant let's declare it as a string on the class, not as a method, same for the mysql jdbi version

ssalinas:
Just the one comment on the static string vs method. Only other thing is to confirm that the history paging in the UI still seems to work as intended once deployed on staging 👍 , will give the ⛵ once you've tested there

SingularityService/src/main/java/com/hubspot/singularity/data/history/PostgresHistoryJDBI.java
fabatef: The user column is named `user` in the mysql version, and `f_user` in the postgres version. That's why I didn't declare it as a constant.

ssalinas: ah, totally read over that difference. It can still just be a `String REQUEST_HISTORY_BASE_QUERY = "...";` on each class though, vs a method

ssalinas: sry, read that again, ignore me...

fabatef:
Tested on `iad03-test` via api goggles. I've spot-checked the request history paging in the UI as well, and things look good.

ssalinas:
🚢

PR #2118, opened 2020-07-17
Make blazar match what's in the pom

ssalinas:
Think this got missed, or not merged properly, after a previous release


pschoenfelder:
🚢

PR #2119, opened 2020-07-17
Add orderDirection param

fabatef:
Allows specifying sort order for request history entries. Forgot to add this on https://github.com/HubSpot/Singularity/pull/2117


ssalinas:
🚢

PR #2120, opened 2020-07-17
Convert to string

fabatef:
Fixing my mistake on https://github.com/HubSpot/Singularity/pull/2119 here. orderDirection should be converted to String due to https://github.com/HubSpot/Singularity/blob/7d53ef8460c8ad0687c0a4389fb2d3f9f9e518ed/SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java#L581


pschoenfelder:
🚢

Issue #2122, opened 2020-07-23
Continuous Reconciliation on a 5 second basis


relistan:
Hey folks, I have spent a fair bit of time debugging this and I believe there is a problem with the code introduced in https://github.com/HubSpot/Singularity/commit/fb05da3250756b857d2b9ea4d5d391219e4683b9 related to stuck launching tasks. It appears to continually request reconciliation for all tasks in the system after they have been running for a short while.

Here's my setup I validated this on:
 * Local Docker stack
 * One agent
 * One master
 * One ZK node
 * Singularity 1.2.0 and current `master` - same behavior

Steps:
 1. Deploy nginx with `OPTIMISTIC` placement strategy
 2. Wait a couple of minutes
 3. See it start to log `Requested explicit reconcile of task ...` every 5 seconds when the `SchedulerPoller` runs. It's calling the Mesos master for every one of these.

## Logs
```
singularity_1   | INFO  [2020-07-23 13:57:03,437] com.hubspot.singularity.scheduler.SingularityScheduler: Requested explicit reconcile of task dev_nginx-latest_6dc0b72-1595512077034-1-docker1-DEFAULT
singularity_1   | INFO  [2020-07-23 13:57:03,437] com.hubspot.singularity.mesos.SingularityMesosOfferScheduler: Received 0 offer(s)
singularity_1   | INFO  [2020-07-23 13:57:03,444] com.hubspot.singularity.mesos.SingularityMesosOfferScheduler: 0 remaining offers not accounted for in offer check
singularity_1   | INFO  [2020-07-23 13:57:03,444] com.hubspot.singularity.mesos.SingularityMesosOfferScheduler: Finished handling 0 new offer(s) 0 from cache (00:00.007), 0 accepted, 0 declined/cached
singularity_1   | INFO  [2020-07-23 13:57:08,447] com.hubspot.singularity.scheduler.SingularityScheduler: Requested explicit reconcile of task dev_nginx-latest_6dc0b72-1595512077034-1-docker1-DEFAULT
singularity_1   | INFO  [2020-07-23 13:57:08,447] com.hubspot.singularity.mesos.SingularityMesosOfferScheduler: Received 0 offer(s)
singularity_1   | INFO  [2020-07-23 13:57:08,455] com.hubspot.singularity.mesos.SingularityMesosOfferScheduler: 0 remaining offers not accounted for in offer check
[...]

PR #2123, opened 2020-07-24
Filter to only starting tasks when reconciling

relistan:
This should fix https://github.com/HubSpot/Singularity/issues/2122 by filtering properly to only launching tasks. It was previously targeted all tasks that were _not_ starting which I am pretty certain the opposite of the intent. Please review #2122 for all the details.


ssalinas:
Thanks for the PR, my internet is out and I haven’t been able to work on much via LTE :cry:

ssalinas:
@pschoenfelder can you help make sure this makes it through the branches and gets merged?

pschoenfelder:
👍 Will start rolling out on Monday.

relistan:
Looks like the merge resolution screwed up the formatting and the formatter is angry (why CI failed). I can fix Monday morning.

ssalinas:
Coming back to this, it seems to have had the opposite effect on our end. Reading the logic again if we want all tasks in a launching state and we are inspecting the the presence of a 'starting' history event in zk, the previous logic was actually correct. If the 'starting' event is _not_ present in zk, explicitly reconcile the task because we want to see if we missed the first event back from the executor for it

This seems to point to #2122 being caused by a separate issue and tasks not actually making it from launching -> starting for some reason in your cluster

I'm going to open a PR to revert this for the moment as it is dramatically slowing down some of our internals. Feel free to stay on the older commit for the moment and we can keep looking into the cause of your issue as well @relistan

PR #2126, opened 2020-08-25
Use leader cache first for launching check, also check staging

ssalinas:
Follow up to #2122 and #2123 , I think I was able to fix both issues:

1) The previous zk check for !exists... was actually correct. However, I realized that some mesos executors do not utilize the task starting state at all, only task staging or task running. So I added both of those to the check
2) This could be more efficient anyway since it should almost always run on the leading instance where the leader cache is already present. When running on the leader, this will now avoid those zk checks entirely and just use the list of history updates already in memory to determine what the most recent state is. This gives us a very simple comparison of == TASK_LAUNCHED to check, not having to worry about which states specific executors use

cc @relistan


relistan:
Aaah, this makes sense @ssalinas. After reading your comments on the other issue I had just come to the conclusion that it must have something to do with the behavior of your executor vs the regular Mesos Docker containerizer and our own executor. We will try this out when we next upgrade. 🍻

PR #2127, opened 2020-08-27
Get exception causes

pschoenfelder:


pschoenfelder:
Not actually necessary

PR #2128, opened 2020-09-04
Make crash loops resolve faster

ssalinas:
Currently we bucket failures into time windows. For example:

|O|O|X|X|O|O|O|O|

and we take into account the number of overall buckets that have failures. However, the one piece I thought the code compensated well for, but does not is that there is a very big difference between:

|X|X|O|O|O|O|O|O|

and

|O|O|O|O|O|X|X|O|

If you imagine each of those buckets is a 3 minute window, that overall window can stretch 30mins. This PR adds an additional condition that the most recent failure timestamp must be in the most recent X% of buckets (e.g. one of the failures must be in the most recent 8 minutes). This should cover cases where something failed for a while then recovered on its own

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/CrashLoopConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/CrashLoopConfiguration.java
@@ -8,19 +8,22 @@ public class CrashLoopConfiguration {
   private int startupFailureThreshold = 5;
 
   private int evaluateOomsOverMinutes = 30;
-  private int oomFailureThreshold = 3;
+  private int oomFailureThreshold = 4;
 
   private int singleInstanceFailureBucketSizeMinutes = 3;
   private int singleInstanceFailureBuckets = 10;
   private double singleInstanceFailureThreshold = 0.25;
+  private int singleInstanceMinBucketIndexPercent = 75; // i.e. must have a failure in most recent 50% of bucket
 
   private int multiInstanceFailureBucketSizeMinutes = 3;
   private int multiInstanceFailureBuckets = 10;
   private double multiInstanceFailureThreshold = 0.4;
+  private int multiInstanceMinBucketIndexPercent = 75; // i.e. must have a failure in most recent 50% of bucket
 
   private int slowFailureBucketSizeMinutes = 30;
   private int slowFailureBuckets = 15;
   private double slowFailureThreshold = 0.7;
+  private int slowFailureMinBucketIndexPercent = 75; // i.e. must have a failure in most recent 70% of bucket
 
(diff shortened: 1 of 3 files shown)


mjball:
Thank you!

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityCrashLoops.java
```diff
@@ -429,20 +434,39 @@ boolean hasCooldownExpired(
...
-    Map<Long, List<Long>> bucketedFailures = recentFailures
-      .stream()
-      .filter(t -> t > thresholdFailTimeMillis)
-      .collect(Collectors.groupingBy(e -> e / bucketSizeMillis));
-    long bucketsWithFailure = bucketedFailures
-      .entrySet()
-      .stream()
-      .filter(e -> !e.getValue().isEmpty())
-      .count();
+    long[] bucketThresholds = new long[numBuckets];
```
mjball, line 441: Why use an array rather than a `List`, OOC?

ssalinas: Ah, I was originally counting something based on the index in that next loop, then changed the implementation a bit. Doesn't particularly matter in current form. Could change to a stream -> List or something like that

ssalinas: updated since List is generally more readable/flexible

SingularityService/src/main/java/com/hubspot/singularity/config/CrashLoopConfiguration.java
```diff
@@ -8,19 +8,22 @@
   private int startupFailureThreshold = 5;
 
   private int evaluateOomsOverMinutes = 30;
-  private int oomFailureThreshold = 3;
+  private int oomFailureThreshold = 4;
 
   private int singleInstanceFailureBucketSizeMinutes = 3;
   private int singleInstanceFailureBuckets = 10;
   private double singleInstanceFailureThreshold = 0.25;
+  private int singleInstanceMinBucketIndexPercent = 75; // i.e. must have a failure in most recent 50% of bucket
```
pschoenfelder, line 16: 75% or 50%?

ssalinas: whoops, will update comment, thanks

ssalinas: fixed

pschoenfelder:
lgtm 🚢

PR #2131, opened 2020-09-18
Fix PostgreSQL typos

iMats:
Noticed you fixed the PostgreSQL support, very cool!
I started testing it and found a couple of simple typos, which I fixed.


ssalinas:
🚢

PR #2140, opened 2020-09-28
Add a cheaper endpoint to return the counts of tasks for a request

ssalinas:
For cases of large counts or where the actual task ids aren't needed. This is much cheaper and faster.


pschoenfelder:
🚢

PR #2141, opened 2020-09-29
Add SNS events for JITA access

ajammala:


SingularityService/src/main/java/com/hubspot/singularity/event/SingularityEventController.java
```diff
@@ -144,6 +145,28 @@ public Void call() {
...
   }
 
+  @Override
+  public void elevatedAccessEvent(ElevatedAccessEvent elevatedAccessEvent) {
+    ImmutableSet.Builder<ListenableFuture<Void>> builder = ImmutableSet.builder();
+
+    for (final SingularityEventSender eventListener : eventListeners) {
+      builder.add(
+        listenerExecutorService.submit(
+          new Callable<Void>() {
```
ssalinas, line 155: feel free to change any of these to lambdas too. Much of this is pre-java8 code which is why there are callables/runnables all over

ssalinas (commented):
Looks good overall. I think the only thing for completness to match other webhook setups would be to add it to the webhooks UI page as well in case folks want to configure it that way. The api for that _should_ 'just work' already, and the UI should just need to extra enum type added I think

ssalinas:
🚢

ssalinas:
🚢 for merge commit

Issue #2145, opened 2020-10-02
Auth issues in Singularity 1.3.0


yuriy-filatov:
Hey, @ssalinas 
Here's a report for 1.3.0 build.
~~Our env: cluster of four 1.2.0 + one 1.3.0, leader is 1.2.0.~~

~~RO user (requiredGroups + globalReadOnlyGroups)~~
- ~~able to deploy/edit/pause/unpause/remove requests, everything actually (shouldn't be)~~
- ~~able to delete racks (shouldn't be)~~
- ~~unable to manage active agents: freeze/decomission throw 405 and flag as inactive -- 403 (seems OK but different codes)~~
- ~~able to delete inactive agents (shouldn't be)~~
- utilization is rendering, but broken (access to /usage/agents is not available)
looks OK

~~Jita user (requiredGroups+jitaGroups):~~
- ~~able to deploy/edit/pause/unpause/remove requests, everything actually (looks OK for jita)~~
- ~~able to delete racks (shouldn't be)~~
- ~~unable to manage active agents: freeze/decomission throw 405 and flag as inactive -- 403 (seems OK but different codes)~~
- ~~able to delete inactive agents (shouldn't be)~~
- utilization is rendering, but broken (access to /usage/agents is not available)
looks OK

~~Admin user (requiredGroups+adminGroup):~~

- ~~admin -> agents -> freeze/decomission: 405 method not allowed (should be able; might this be related to a bunch of recent renames?)~~
- utilization is not rendering at all
rest looks OK

Any mail-reported actions (pause/unpause for example) don't have singularityUserId (actually this is a long-runner, unrelated to current issue and is observed at least for last 12 months). Can singularityUserId be also parsed from http://URI/singularity/api/auth/user from user:id field if auth is enabled?

ssalinas:
Hey, first comment is that Singularity wasn't generally set up to run mixed versions like that. For example the UI moved to the new naming, so if you got the UI from 1.3 but then hit a 1.2 instance it wouldn't work in all cases. Upgrading all should account for any of the 405s

utilization is a known issue

Can look closer at some of the other descriptions in a bit

yuriy-filatov:
Fixed my comment above, thanks.

yuriy-filatov:
Thanks for the update, we've rebuilt our conf according to new docs and rechecked everything -- looks very nice, thank you!

Only annoying thing is that we're not getting userId in email logback, even for authorized user:
[...]

PR #2146, opened 2020-10-02
Better auth docs

ssalinas:
@yuriy-filatov , realized my docs on the usage of the updated auth never made it into 1.3.0. Adding them here in a PR now at least so you can try it out more:

cc https://github.com/HubSpot/Singularity/issues/2145


yuriy-filatov:
Really helpful, thanks for this!

PR #2147, opened 2020-10-02
Add actions to group/scope auth

pschoenfelder:
Some requests need to be runnable ONLY by someone other than the group owner for compliance purposes. For example, Team A would write and deploy a sensitive request, and only Team B could run it (or bounce, scale, kill tasks, or pause).

This provides a more generic solution in the form of an action permissions override map.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityAuthorizationScope.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityAuthorizationScope.java
@@ -4,5 +4,6 @@ public enum SingularityAuthorizationScope {
   READ,
   WRITE,
   ADMIN,
-  DEPLOY
+  DEPLOY,
+  EXEC // run, bounce, kill, pause, scale
 }
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityAuthorizationScope.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityAuthorizationScope.java
@@ -3,7 +3,5 @@ package com.hubspot.singularity;
 public enum SingularityAuthorizationScope {
   READ,
   WRITE,
-  ADMIN,
-  DEPLOY,
-  EXEC // run, bounce, kill, pause, scale
+  ADMIN
 }
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserFacingAction.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserFacingAction.java
@@ -1,0 +1,12 @@
+package com.hubspot.singularity;
+
+public enum SingularityUserFacingAction {
+  RUN_REQUEST,
+  BOUNCE_REQUEST,
+  PAUSE_REQUEST,
+  SCALE_REQUEST,
+  KILL_TASK,
+  BOUNCE_TASK,
+  RUN_SHELL_COMMAND,
+  DELETE_SCHEDULED_TASK
+}
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserFacingAction.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserFacingAction.java
@@ -1,12 +1,11 @@
 package com.hubspot.singularity;
 
 public enum SingularityUserFacingAction {
-  RUN_REQUEST,
-  BOUNCE_REQUEST,
-  PAUSE_REQUEST,
-  SCALE_REQUEST,
+  EXEC,
+  BOUNCE,
+  PAUSE,
+  SCALE,
   KILL_TASK,
-  BOUNCE_TASK,
-  RUN_SHELL_COMMAND,
-  DELETE_SCHEDULED_TASK
+  DELETE_SCHEDULED_TASK,
+  RUN_SHELL_COMMAND
 }
--- a/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityUpdateGroupsRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityUpdateGroupsRequest.java
@@ -2,7 +2,7 @@ package com.hubspot.singularity.api;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.hubspot.singularity.SingularityAuthorizationScope;
+import com.hubspot.singularity.SingularityUserFacingAction;
 import io.swagger.v3.oas.annotations.media.Schema;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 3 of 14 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityAuthorizationScope.java
```diff
@@ -4,5 +4,6 @@
   READ,
   WRITE,
   ADMIN,
-  DEPLOY
+  DEPLOY,
+  EXEC // run, bounce, kill, pause, scale
```
ssalinas, line 8: Since we are adding granularity, would it be too much harder to actually make each of these their own separate scope? Would let us much more easily divide things up later if needed

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityGroupsScopesAuthorizer.java
```diff
@@ -211,6 +144,19 @@ public boolean isAuthorizedForRequest(
     }
   }
 
+  private void warnJita(
```
ssalinas, line 147: just FYI that this will have some merge conflicts with #2141

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityGroupsScopesAuthorizer.java
```diff
@@ -239,52 +185,41 @@ private boolean isJita(SingularityUser user) {
...
-
-  private boolean hasWriteScope(SingularityUser user) {
-    return groupsIntersect(scopesConfiguration.getWrite(), user.getScopes());
+  private Set<String> getScopes(SingularityAuthorizationScope scope) {
+    switch (scope) {
+      case READ:
+        return Sets.union(scopesConfiguration.getRead(), scopesConfiguration.getWrite());
+      case WRITE:
+        return scopesConfiguration.getWrite();
+      case DEPLOY:
```
ssalinas, line 197: If it makes it simpler in the overall implementation, I'm fine with having the DEPLOY scope behave similar to the new ones (i.e. falls under WRITE unless otherwise overriden by a request). I had originally separated out DEPLOY for future use with our internal automation, but it isn't turned on or enforced by default

pschoenfelder: Do you mean change DEPLOY to not fall back to use WRITE? I set up EXEC so that it doesn't fall back to write, but that the request group gets EXEC privs unless explicitly removed by the overrides.

pschoenfelder: We're going to have to add all these scopes for all the appropriate people anyways, it would be easy to add DEPLOY for them too

SingularityBase/src/main/java/com/hubspot/singularity/api/SingularityUpdateGroupsRequest.java
```diff
@@ -13,20 +15,25 @@
   private final Optional<String> group;
   private final Set<String> readWriteGroups;
   private final Set<String> readOnlyGroups;
+  private final Optional<Map<String, Set<SingularityAuthorizationScope>>> groupScopeOverrides;
```
ssalinas, line 18: should this also be renamed/updated to actionPermissions?

SingularityService/src/main/java/com/hubspot/singularity/config/ScopesConfiguration.java
```diff
@@ -14,6 +14,9 @@
   @NotEmpty
   private Set<String> read = Collections.singleton("SINGULARITY_READONLY");
 
+  @NotEmpty
+  private Set<String> exec = Collections.singleton("SINGULARITY_EXEC");
```
ssalinas, line 18: can probably remove this since we went with the actions strategy instead

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserFacingAction.java
```diff
@@ -0,0 +1,12 @@
+package com.hubspot.singularity;
+
+public enum SingularityUserFacingAction {
+  RUN_REQUEST,
```
ssalinas, line 4: naming nits:
- Not sure we need the _REQUEST on each of these. Will likely be easier for users to remember in their configs without it
- Maybe RUN -> EXEC/EXECUTE?

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserFacingAction.java
```diff
@@ -0,0 +1,12 @@
...
+  BOUNCE_REQUEST,
+  PAUSE_REQUEST,
+  SCALE_REQUEST,
+  KILL_TASK,
+  BOUNCE_TASK,
```
ssalinas, line 9: Is bounce task any different from KILL_TASK? I don't see it used anywhere

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityGroupsScopesAuthorizer.java
```diff
@@ -315,17 +270,50 @@ private void checkForbiddenForGroups(
...
     Set<String> allowedReadGroups = new HashSet<>(
       authConfiguration.getGlobalReadOnlyGroups()
     );
     allowedReadGroups.addAll(authConfiguration.getGlobalReadWriteGroups());
-    if (!request.getReadOnlyGroups().isPresent()) {
-      allowedReadGroups.addAll(authConfiguration.getDefaultReadOnlyGroups());
+
+    if (request.getActionPermissions().isPresent()) {
+      request
+        .getActionPermissions()
```
ssalinas, line 296: is this needed here still? I think the action permissions are all write-based (and also not SingularityAuthorizationScope objects)

ssalinas (commented):
Overall I like the actions strategy, seems a lot cleaner. One open question is if a write-jita should allow a user to bypass not being the proper action group. I think the answer should be yes there, but open to opinions. Definitely +1 for unit tests here. It's much easier to tell that all the authorizer bits are working with those than by reading the code alone

SingularityBase/src/main/java/com/hubspot/singularity/SingularityUserFacingAction.java
pschoenfelder: Agreed on dropping _REQUEST.

I feel like RUN matches up with the ui ("Run Now"), where EXEC/EXECUTE isn't a term users typically use, so it may be easier to remember.

SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityGroupsScopesAuthorizer.java
pschoenfelder: Good catch - missed this one

pschoenfelder:
re: write-jita — Could we disallow jita action bypass by default, and leave it up to the user to add the jita group to the action permissions if that should be allowed? It's less obvious for users, but more protective and explicit. If not, then I would assume that write-jita should allow bypassing the action permissions.

ssalinas:
I think the fact that we have the audit events that Aditya added should be enough on the jita front to just allow jita through to everything. One of the main requirements on this was that any actions normally not allowed would at least be logged/audited

pschoenfelder:
Added tests that validate JITA support for actions

ssalinas:
🚢

PR #2150, opened 2020-10-19
sentry cleanup in s3uploader

ssalinas:
Short circuiting a few cases that we know can never succeed, but will constantly retry and spam exceptions in logs/into sentry


ajammala:
🚢

PR #2151, opened 2020-10-26
Fix env var inheritance in docker

pschoenfelder:
Env vars need to be gotten while running the `docker.sh` (templated out as `runner.sh`), because they may not be available from the executor process.

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/DockerEnvironmentContext.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/DockerEnvironmentContext.java
@@ -1,5 +1,6 @@
 package com.hubspot.singularity.executor.models;
 
+import com.google.common.base.Joiner;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/models/DockerEnvironmentContext.java
```diff
@@ -20,16 +21,12 @@ public DockerEnvironmentContext(TaskInfo taskInfo, List<String> inheritVariables
...
-          if (val != null) {
-            variables.add(Variable.newBuilder().setName(v).setValue(val).build());
-            keys.add(v);
-          }
-        }
-      }
+    variables.add(
+      Variable
+        .newBuilder()
+        .setName("INHERIT_ENV_VARS")
```
ssalinas, line 27: If we are going to loop over these, no need to make it another variable. Doing it this way will end up setting an extra variable that will be unused. Can add the list instead as another getter and loop over it in the template functions to avoid duplication/unused vars

ssalinas:
closing since this was actually a separate issue

PR #2152, opened 2020-11-28
Don't include history in leader cache query + more efficient task persister

ssalinas:
Think this is slowing us down when sql purge has not run in a while. We don't need the full history, only active


pschoenfelder:
🚢

PR #2153, opened 2020-12-01
Also check history if not in leader cache

ssalinas:
Follow up from #2152 . Some of the api endpoints on the leader can be a bit funky due to that. If we don't find the task in our leader cache, we should also check zk, now that the leader cache only has active tasks


pschoenfelder:
🚢

PR #2156, opened 2020-12-16
Fix accidental crash loop dedupe

ssalinas:
This fixes an issue where, if a crash loop of the same type had just resolved, a new crash loop of that type could be missed. The previouslyActive.contains check was eating these due to the implementation of equals/hashCode on the CrashLoopInfo object. Changed that and the check to determine which are no longer active as well as adding start/end time to the equals and hash methods


pschoenfelder:
🚢

PR #2158, opened 2020-12-21
Stability improvements for sql/zk usage

ssalinas:
Few things here after we experienced some hiccups with persisting:
- Sort the task persisting by child node count, to take care of more expensive/troublesome parent nodes first
- Enable spinning up a separate instance that only serves api calls (i.e. a deployment pattern to isolate the scheduler)
- Optimize the active tasks for request id check
- make a cheaper HEAD request method to check if a request exists without having to fetch full request + active task data for the request parent object


pschoenfelder:
🚢  looks good to me

mikegajda:
:ship: LGTM

PR #2159, opened 2020-12-22
Fix condition for evaluating crash loops

ssalinas:
Looking into some cases where this triggered too slow for something that should have caught the startup failure loop. Trying to remember why this was a ! ...


pschoenfelder:
🚢

PR #2160, opened 2021-01-04
More efficient active tasks for request listing

ssalinas:
Untangled a bit of code from when the active task parent used to be consolidated instead of split by request id. The very I thought would be more stable actually scaled zk call count with number of requests instead of number of tasks for the request being queried. Since the active tasks are split unlike the old setup a year+ ago, this can actually be always a single zk call for active tasks for a request


pschoenfelder:
🚢

PR #2165, opened 2021-01-14
Additional counters and logging in publish subject failures

ssalinas:
cc @ajammala @pschoenfelder finally found where to do this. Will wire through to a metric


pschoenfelder:
🚢

PR #2168, opened 2021-01-21
Add ability to re-add recovered tasks to the load balancer

ssalinas:
Previously our handling of this was to just clean up everything if it was already removed from the load balancer. This PR adds a few things to the reregistration flow:

- Always enqueue a pending request to reevaluate state/scale so that we don't end up with a multiple instance 1s kind of state and get stuck there
- If the task is recoverable in zk (not persisted yet) but has already been removed from the lb. Instead of cleaning it up, treat it as if it were a new healthy task that just finished passing healthchecks. This way we can recover running tasks, especially in cases of large network partitions where relaunching that many new things could take some large amount of time

Updated the unit test for this to make sure the lb pending add is present, but would appreciate extra 👀 on it

cc @pschoenfelder @ajammala @rosalind210

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
@@ -16,6 +16,7 @@ import com.hubspot.singularity.LoadBalancerRequestType;
 import com.hubspot.singularity.RequestType;
 import com.hubspot.singularity.Singularity;
 import com.hubspot.singularity.SingularityCreateResult;
+import com.hubspot.singularity.SingularityLoadBalancerUpdate;
 import com.hubspot.singularity.SingularityMainModule;
 import com.hubspot.singularity.SingularityManagedThreadPoolFactory;
 import com.hubspot.singularity.SingularityPendingDeploy;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -346,14 +347,57 @@ private StatusUpdateResult unsafeProcessStatusUpdate(
...
+            if (
+              maybeTask.isPresent() &&
+              maybeRequest.isPresent() &&
+              maybeRequest.get().getState().isRunnable()
+            ) {
+              LOG.info(
+                "Task {} can be recovered. Clearing LB state and enqueuing check as new task",
+                taskId
+              );
+              taskManager.clearLoadBalancerHistory(taskIdObj);
```
pschoenfelder, line 383: Is there any chance we get into a weird state if we don't clear this before attempting to reactivate the task? I'm thinking new LB stuff put into zk between the two calls?

ssalinas: My initial thought is that, if we have gotten here, we already processed a task_lost update for the task and considered it terminal. However, there is the possibility that a removal might still be in progress if the removal was slow 🤔 , good point

ssalinas: Added handling for possible in-progress LB removal now ✅

pschoenfelder:
🚢

PR #2169, opened 2021-01-22
Initialize last persister AtomicLongs to current time

rosalind210:
Initializing each last persister success AtomicLongs to current time so we don't get alerts. When the metric is first grabbed and initialized to zero, the metric is read as last success in 1970, which triggers an alert.


pschoenfelder:
🚢

PR #2170, opened 2021-01-25
Default to user id when no email available in notifications

ssalinas:
/fixes https://github.com/HubSpot/Singularity/issues/2145

cc @yuriy-filatov


rosalind210 (approved):

rosalind210:
🚢

PR #2172, opened 2021-01-27
Added logging decorator for CuratorFramework to Audit ZK Endpoints

rosalind210:
Created a decorator for the CuratorFramework to handle logging calls to ZooKeeper. It holds a maps that connects the thread ID, classname, and method  nameto the number of zookeeper calls it has made over certain interval of time. 

The most important updates: 
 - `LoggingCuratorFramework` created that keeps track of call counts and logs when that amount 
 - New framework is set as CuratorFramework in the `SingularityCuratorProvider`, if a boolean `useLoggingCuratorFramework` is true in the `SingularityConfiguration`
 - logging curator's `clear()` performed by a timer, clears every 30 milliseconds.

[1/28 Update] Added timer to LoggingCuratorFramework that performs a clear (logs and empties everything in the map) every 3 minutes.

[1/29] Logs tested in `iad03-test`, output:
```
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 49: TaskManager's getTasks called ZK 1 times in 180000 milliseconds
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 45: TaskManager's getTaskIdsForRequest called ZK 6 times in 180000 milliseconds
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 49: TaskManager's filterInactiveTaskIds called ZK 12 times in 180000 milliseconds
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 47: TaskManager's getTaskHistoryUpdates called ZK 12 times in 180000 milliseconds
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 43: StateManager's getState called ZK 4 times in 180000 milliseconds
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 49: TaskManager's getTaskMetadata called ZK 2 times in 180000 milliseconds
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 49: TaskManager's getHealthcheckResults called ZK 2 times in 180000 milliseconds
INFO  [2021-01-29 21:22:58,168] com.hubspot.singularity.data.LoggingCuratorFramework: 49: DeployManager's getRequestDeployState called ZK 2 times in 180000 milliseconds
```

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
@@ -115,9 +115,9 @@ public class LoggingCuratorFramework implements CuratorFramework {
         callCount.get(),
         INTERVAL
       );
-    }
 
-    COUNTERS.clear();
+      COUNTERS.remove(caller);
+    }
   }
 
   @Override
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
@@ -3,6 +3,8 @@ package com.hubspot.singularity.data;
 import com.google.inject.Inject;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Timer;
+import java.util.TimerTask;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
@@ -39,7 +41,26 @@ import org.slf4j.LoggerFactory;
 
 public class LoggingCuratorFramework implements CuratorFramework {
   private final CuratorFramework curator;
-  private final Map<String, AtomicLong> counters;
+
+  private static final Map<String, AtomicLong> COUNTERS;
+  private static final Timer TIMER;
+  private static final Long INTERVAL = 30L;
+
+  static {
+    COUNTERS = new HashMap<>();
+    TIMER = new Timer();
+    TIMER.schedule(
+      new TimerTask() {
+
+        @Override
+        public void run() {
+          LoggingCuratorFramework.clear();
+        }
+      },
+      INTERVAL,
+      INTERVAL
+    );
+  }
 
   private static final Logger LOG = LoggerFactory.getLogger(
     LoggingCuratorFramework.class
@@ -48,50 +69,55 @@ public class LoggingCuratorFramework implements CuratorFramework {
   @Inject
   public LoggingCuratorFramework(CuratorFramework curator) {
     this.curator = curator;
-    this.counters = new HashMap<>();
   }
 
-  private String getCaller() {
+  private static String getCaller() {
     StackTraceElement[] stackTraceElements = Thread.currentThread().getStackTrace();
+    long threadId = Thread.currentThread().getId();
 
+    int levelInStack = 1;
     if (
       stackTraceElements[1].getClassName()
         .equals("com.hubspot.singularity.data.CuratorManager") ||
       stackTraceElements[1].getClassName()
         .equals("com.hubspot.singularity.data.CuratorAsyncManager")
     ) {
-      return (
-        stackTraceElements[2].getClassName() +
-        "\'s " +
-        stackTraceElements[2].getMethodName()
-      );
-    } else {
-      return (
-        stackTraceElements[1].getClassName() +
-        "\'s " +
-        stackTraceElements[1].getMethodName()
-      );
+      levelInStack = 2;
     }
+
+    String longClassName = stackTraceElements[levelInStack].getClassName();
+    String className = longClassName.substring(longClassName.lastIndexOf("."));
+
+    return (
+      threadId +
+      ": " +
+      className +
+      "\'s " +
+      stackTraceElements[levelInStack].getMethodName()
+    );
   }
 
   private void setCounter() {
     String caller = getCaller();
-    AtomicLong counter = this.counters.containsKey(caller)
-      ? this.counters.get(caller)
+    AtomicLong counter = COUNTERS.containsKey(caller)
+      ? COUNTERS.get(caller)
       : new AtomicLong(0);
     counter.getAndIncrement();
-    this.counters.put(caller, counter);
+    COUNTERS.put(caller, counter);
   }
 
-  public void clear() {
-    String caller = getCaller();
-    AtomicLong counter = this.counters.containsKey(caller)
-      ? this.counters.get(caller)
-      : new AtomicLong(0);
+  public static void clear() {
+    for (String caller : COUNTERS.keySet()) {
+      AtomicLong callCount = COUNTERS.get(caller);
+      LOG.info(
+        "{} called ZK {} times in {} milliseconds",
+        caller,
+        callCount.get(),
+        INTERVAL
+      );
+    }
 
-    LOG.info("{} called ZK {} times", caller, counter.get());
-
-    this.counters.remove(caller);
+    COUNTERS.clear();
   }
 
   @Override
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
@@ -1,11 +1,11 @@
 package com.hubspot.singularity.data;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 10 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,312 @@
...
+import org.apache.zookeeper.server.quorum.flexible.QuorumVerifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingCuratorFramework implements CuratorFramework {
+  private final CuratorFramework curator;
+
+  private final ConcurrentMap<String, Integer> counters;
+
+  private final Long interval = 60000L * 3; // 3 minutes
```
rosalind210, line 46: is this a good amount of time in between flushing data to logs?

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,312 @@
...
+        LOG.info("{} called ZK {} times in {} milliseconds", caller, count, interval);
+
+        counters.remove(caller);
+      }
+    );
+  }
+
+  @Override
+  public void start() {
+    curator.start();
```
rosalind210, line 119: should `start()` be logged?

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  @Override
+  public void close() {
+    curator.close();
+  }
+
+  @Override
+  public CuratorFrameworkState getState() {
+    setCounter();
```
rosalind210, line 129: should `getState()` be logged or not?

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
pschoenfelder: Seems reasonable to me 👍 . Can always adjust if needed. For clarity, could also do `interval = TimeUnit.MINUTES.toMillis(3);`

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,312 @@
...
+      className +
+      "\'s " +
+      stackTraceElements[levelInStack].getMethodName()
+    );
+  }
+
+  private void setCounter() {
+    String caller = getCaller();
+    int counter = counters.getOrDefault(caller, 0);
+    counters.put(caller, ++counter);
```
pschoenfelder, line 104: getting and setting in separate operations won't be atomic, something like `counters.compute(key, (k, v) -> v == null ? 1 : v+1)` should do the trick.

pschoenfelder: nit — may want to change `setCounter` with `incCounter` or similar for clarity

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+  }
+
+  @Inject
+  public LoggingCuratorFramework(CuratorFramework curator) {
+    this.curator = curator;
+    counters = new ConcurrentHashMap<>();
+    Timer timer = new Timer();
+
+    timer.schedule(new ClearCounter(), interval, interval);
```
pschoenfelder, line 66: We should use a ScheduledExecutorService instead of a Timer here (better thread control, exception handling).

There's a good example in the SingularityHealthchecker https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityHealthchecker.java#L93
https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityHealthchecker.java#L277

pschoenfelder:
Looks solid overall 👍  Couple minor things but the core is all there

I'm curious — will these logs mostly be read by humans or machines (presumably aggregating for further analysis)? If the latter, may want to make the output format more easily parsable, but otherwise looks good.

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
pschoenfelder: I think we can probably leave this (and `getState`) out since they should be fairly infrequent.

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,319 @@
...
+
+  public void logAndClear() {
+    counters.forEach(
+      (caller, count) -> {
+        LOG.info("{} called ZK {} times in {} minutes", caller, count, interval);
+
+        counters.remove(caller);
+      }
+    );
+  }
```
pschoenfelder, line 124: Think I missed this before — doing `counters.put(caller, 0)` instead of `remove` would prevent any iterator invalidation stemming from removal (or additions from other threads) during iteration.

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,319 @@
...
+    LoggingCuratorFramework.class
+  );
+
+  private final ScheduledExecutorService executorService;
+
+  private class ClearCounter implements Runnable {
+
+    @Override
+    public void run() {
+      logAndClear();
```
pschoenfelder, line 58: Can we wrap this in a try catch and log out any exceptions? If an exception were ever to escape `run()`, all subsequent executions would get cancelled.

pschoenfelder:
🚢

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,323 @@
...
+import org.apache.zookeeper.server.quorum.flexible.QuorumVerifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingCuratorFramework implements CuratorFramework {
+  private final CuratorFramework curator;
+
+  private final ConcurrentMap<String, Integer> counters;
+
+  private final Long interval = 3L;
```
ssalinas, line 46: style thing. This can be `static` which also means it should have a variable named in upper case like `INTERVAL`. Also a nit, but something specific like `FLUSH_INTERVAL` might give a better idea of what it's for

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,323 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(
```
ssalinas, line 48: more really nit, we generally put all the static things at the top of the class. e.g.

class {
private static final stuff
private final stuff
private stuff
constructor
}

SingularityService/src/main/java/com/hubspot/singularity/data/LoggingCuratorFramework.java
```diff
@@ -0,0 +1,323 @@
...
+  public LoggingCuratorFramework(
+    CuratorFramework curator,
+    SingularityManagedScheduledExecutorServiceFactory executorServiceFactory
+  ) {
+    this.curator = curator;
+    counters = new ConcurrentHashMap<>();
+
+    this.executorService = executorServiceFactory.get("logging-curator-framework");
+    this.executorService.scheduleAtFixedRate(
+        new ClearCounter(),
```
ssalinas, line 76: slightly different way of doing this in functional form. This can be just a private method instead of a class that extends runnable and you can pass the method reference to the executor service call.
edit: and add the try/catch in the private method then as well

ssalinas:
🚢

Issue #2171, opened 2021-01-27
Some of 'Disabled actions' are not working after rename


yuriy-filatov:
When trying to add new disabled action in UI at Admin -> Disasters -> Disabled actions, any action with %agent% is throwing an error:
`An error occurred while accessing /disasters/disabled-actions/DECOMMISSION_AGENT
path param action must be one of [BOUNCE_REQUEST, SCALE_REQUEST, REMOVE_REQUEST, CREATE_REQUEST, UPDATE_REQUEST, VIEW_REQUEST, PAUSE_REQUEST, KILL_TASK, BOUNCE_TASK, RUN_SHELL_COMMAND, ADD_METADATA, DEPLOY, CANCEL_DEPLOY, ADD_WEBHOOK, REMOVE_WEBHOOK, VIEW_WEBHOOKS, TASK_RECONCILIATION, STARTUP_TASK_RECONCILIATION, RUN_HEALTH_CHECKS, ADD_DISASTER, REMOVE_DISASTER, DISABLE_ACTION, ENABLE_ACTION, VIEW_DISASTERS, FREEZE_SLAVE, ACTIVATE_SLAVE, DECOMMISSION_SLAVE, VIEW_SLAVES, FREEZE_RACK, ACTIVATE_RACK, DECOMMISSION_RACK, VIEW_RACKS, SEND_EMAIL, PROCESS_OFFERS, CACHE_OFFERS, EXPENSIVE_API_CALLS, RUN_CLEANUP_POLLER, RUN_DEPLOY_POLLER, RUN_SCHEDULER_POLLER, RUN_EXPIRING_ACTION_POLLER, RUN_UPSTREAM_POLLER, TASK_SHUFFLE]`

Suppose this should be renamed as well (or am I doing something wrong?)
Currently I'm checking version 1.4.0

PR #2173, opened 2021-01-28
Remap names in disabled actions list

ssalinas:
/fixes https://github.com/HubSpot/Singularity/issues/2171

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/DisasterManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/DisasterManager.java
@@ -68,20 +68,21 @@ public class DisasterManager extends CuratorAsyncManager {
       new SingularityDisabledAction(
         action,
         String.format(MESSAGE_FORMAT, action, DEFAULT_MESSAGE),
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/DisasterManager.java
```diff
@@ -68,20 +68,21 @@ public SingularityDisabledAction getDisabledAction(SingularityAction action) {
       new SingularityDisabledAction(
         action,
         String.format(MESSAGE_FORMAT, action, DEFAULT_MESSAGE),
-        Optional.<String>empty(),
+        Optional.empty(),
```
rosalind210, line 71: Clarification for me, why are you removing the <String> and <Long>?

ssalinas: general cleanup. They were needed in older java versions, but not in newer ones. Been cleaning up files as I happen to edit them for other reasons

rosalind210:
🚢

PR #2175, opened 2021-02-04
Throw exception to resubscribe to mesos

rosalind210:
Sometimes there's an issue connecting to Mesos master and the alert that went off was a Mesos4xxException, stating that the framework is no longer subscribed to master. When that happens, we want to throw an exception to be picked up by the `SingularityMesosSchedulerImpl` to pause and reconnect. This is done on a separate thread so as to not disrupt the current one.

Unit tests check that the new function checkAndReconnect only attempts a reconnect when the exception is the correct exception.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
@@ -24,6 +24,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.mesos.v1.Protos.AgentID;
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerClient.java
```diff
@@ -358,18 +368,36 @@ public void sendCall(Call call) {
...
+    );
+
+    String message = t.getMessage();
+
+    if (
+      message != null &&
+      message.equals(
+        "Error while trying to send request. Status: 403 Message: \'Framework is not subscribed\'"
+      )
+    ) {
```
ssalinas, line 391: I think we can make this a bit more flexible so we aren't at the mercy of the exception message. I think this is a valid flow for any time we get a `Mesos4xxException` exception with a 403 in the message (or if the exception itself has a status code in it that would be even better to compare to the int vs a string.contains)

rosalind210: I can update to check that the message contains 403 but I removed the Mesos4xxException check earlier because I can't mock it because it's a final class

PR #2176, opened 2021-02-08
Opt in mode for Singularity emails

rosalind210:
This PR updates emailing to be opt-in rather than opt-out based on the optInEmailMode value being true.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/NotificationsManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/NotificationsManager.java
@@ -41,19 +41,13 @@ public class NotificationsManager extends CuratorManager {
   }
 
   public void unsubscribe(String email) {
-    if (configuration.isOptInEmailMode()) {
-      removeFromAllowlist(email);
-    } else {
-      addToBlocklist(email);
-    }
+    removeFromAllowlist(email);
+    addToBlocklist(email);
   }
 
   public void subscribe(String email) {
-    if (configuration.isOptInEmailMode()) {
-      addToAllowlist(email);
-    } else {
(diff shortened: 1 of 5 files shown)


ssalinas:
cc @ajammala , gonna have you be the reviewer here once this PR is ready

SingularityService/src/main/java/com/hubspot/singularity/data/NotificationsManager.java
```diff
@@ -39,21 +40,49 @@ public NotificationsManager(
         );
   }
 
+  public void unsubscribe(String email) {
+    removeFromAllowlist(email);
```
rosalind210, line 44: originally used a conditional based on the `optInEmailMode` but updated to perform both operations so that not much is lost if we revert back to opt-out, but I also couldn't find many documents on how expensive this approach is

ajammala: :+1: - agree with the approach.

ajammala: Should not be bad - if we decide to revert - the block list should still contain all the original information.

SingularityService/src/main/java/com/hubspot/singularity/resources/NotificationsResource.java
```diff
@@ -49,6 +49,13 @@ public NotificationsResource(
     return notificationsManager.getBlocklist();
   }
 
+  @GET
+  @Path("/allowlist")
```
ajammala, line 53: Not very important  - can probably use /subscribe for the path

pschoenfelder:
🚢

PR #2177, opened 2021-02-10
Clean up lifecycle ordering on shutdown

ssalinas:
Realized there were a few cases in here where we could possibly still have blips of multiple leaders or no leaders at a time. In particular the following set of events:

- Old leader gets a TERM signal
- Old leader fails to properly wait for all scheduled pollers to exit
- Old leader relinquishes leadership
- New leader gains leadership and starts loading leader cache
- Old leader pollers finish their run, possibly flush more state to zk
- state flushed by old leader is now not in leader cache on new leader

Need to test this a bit, but this adds:
- A specific pre-jetty-shutdown hook so that the leader cleanly moves from the leading instance to a read-only/standby before actually fully shutting down (so that we don't have weird blips in serving writes when leader changes). All pollers should stop in this step
- poll for leadership change to make sure it has propagated
- better/longer wait on running pollers
- log when pollers aren't shut down cleanly

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedThreadPoolFactory.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedThreadPoolFactory.java
@@ -93,7 +93,7 @@ public class SingularityManagedThreadPoolFactory {
           final long start = System.currentTimeMillis();
 
           if (!service.awaitTermination(timeoutLeftInMillis, TimeUnit.MILLISECONDS)) {
-            LOG.warn("Eeecutor service tasks did not exit in time");
+            LOG.warn("Executor service tasks did not exit in time");
             continue;
           }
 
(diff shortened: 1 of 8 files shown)


SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedThreadPoolFactory.java
```diff
@@ -88,7 +93,8 @@ public void stop() throws Exception {
           final long start = System.currentTimeMillis();
 
           if (!service.awaitTermination(timeoutLeftInMillis, TimeUnit.MILLISECONDS)) {
-            return;
+            LOG.warn("Eeecutor service tasks did not exit in time");
```
rosalind210, line 96: nit: typo on `executor`

rosalind210:
🚢

PR #2185, opened 2021-03-04
Block reconciliation if queue nearly full

rosalind210:
If the queue is nearly full, adding a number of tasks for reconciliation can put us into a loop of lag as the scheduler tries to go through tasks and more reconciliation attempts.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
@@ -44,6 +44,7 @@ public class SingularityTaskReconciliation {
   private static final Logger LOG = LoggerFactory.getLogger(
     SingularityTaskReconciliation.class
   );
+  private static final double NEARLY_FULL = 0.9;
 
   private final TaskManager taskManager;
   private final String serverId;
@@ -94,6 +95,17 @@ public class SingularityTaskReconciliation {
   }
 
   public ReconciliationState startReconciliation() {
+    if (
+      (float) taskManager.getNumScheduledTasks() /
+      configuration.getOfferCacheSize() >=
+      NEARLY_FULL
+    ) {
+      LOG.info("Queue is nearly full, not starting reconciliation");
+      return isRunningReconciliation.get()
+        ? ReconciliationState.ALREADY_RUNNING
+        : ReconciliationState.NO_DRIVER;
+    }
+
     final long taskReconciliationStartedAt = System.currentTimeMillis();
 
     if (!isRunningReconciliation.compareAndSet(false, true)) {


SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
```diff
@@ -44,6 +44,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(
     SingularityTaskReconciliation.class
   );
+  private static final double NEARLY_FULL = 0.9;
```
rosalind210, line 47: is this a good number to consider nearly full?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
```diff
@@ -94,6 +95,17 @@ boolean isReconciliationRunning() {
   }
 
   public ReconciliationState startReconciliation() {
+    if (
+      (float) taskManager.getNumScheduledTasks() /
+      configuration.getOfferCacheSize() >=
```
rosalind210, line 100: this was the closest value I could find for the denominator based on [these docs](http://getsingularity.com/Docs/reference/configuration.html#operational)

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
```diff
@@ -94,6 +95,17 @@ boolean isReconciliationRunning() {
...
+      NEARLY_FULL
+    ) {
+      LOG.info("Queue is nearly full, not starting reconciliation");
+      return isRunningReconciliation.get()
```
rosalind210, line 104: based on the issue we saw, we were already in a state of reconciliation when we entered the endless reconciliation attempt loop, so I wanted to base the return off of that possibility

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
ssalinas: I think we could always make this a field in SingularityConfiguration so we can edit it later if needed. As a note on size though, I think I'd go slightly lower maybe? Especially for something with 1000s of tasks, we'd want a good amount of overhead space. Maybe more like .75 or .8?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
ssalinas: Not quite the right thing here. We may actually need to instrument an executor to get this. Status updates are put in queue [here](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosSchedulerImpl.java#L742-L747) and put into the queue for [this executor service](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java#L647). Essentially what we want to inspect is the size of queued work on the executor service used in the status update handler

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
ssalinas: I think we also check this a few lines down

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
rosalind210: oh good point to put it in the SingularityConfiguration -- I'll move it. And yeah, that makes sense to lower it. when we discussed as a team, it sounded like we were already running at 80%, or was I misunderstanding?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
rosalind210: but this is short circuiting all of that logic, should I not short circuit?

ssalinas: I think the `if (!isRunningReconciliation.compareAndSet(false, true)) {` should accomplish the same short circuit. Which still makes me wonder about how we got multiple of those loops in there 🤔

rosalind210: hmm you're right, that is odd. maybe there is a different bug?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
rosalind210: Updated this to use the SingularityMesosStatusUpdateHandler: I added queue fullness function, but I wasn't sure if I should continue calculating fullness in SingularityTaskReconciliation with `configuration.getMesosConfiguration().getMaxStatusUpdateQueueSize()`

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityTaskReconciliation.java
rosalind210: Added statusQueueNearlyFull value to SingularityConfiguration and set it to 0.8 for now.

ssalinas:
let's test in staging, but looks good 👍 . We can do something like shrink the queue size really small + make an example request and scale it to something ridiculous like 300 to hopefully trigger and see it in action

rosalind210:
I tried adding unit tests, but the status queue was still processing them faster than I could run a reconciliation check so I wasn't able to hit the blocking conditional; and if I tried to create too many tasks, I was blocked by testing resources.

ssalinas:
🚢

ssalinas:
🚢

PR #2186, opened 2021-03-12
Create thread pool that blocks when full

ssalinas:
So, the theory here, is that it will be much better to just block the producing thread (main rxjava thread carrying mesos events) when the status update queue is full than to throw a RejectedExecutionException and abort/restart/bai//etc. If we are falling behind on status updates it's nearly always the case that those have to catch up first before we can do anything else.

Java-y bits added here are a custom ThreadPoolQueue which, because it delegates to put, will block on the put to the underlying queue when full instead of throwing the reject exception


rosalind210:
🚢

PR #2187, opened 2021-03-15
Add status queue size accessor for status update handler

rosalind210:
Added an accessor to get the status queue size from the `statusUpdatesExecutor` to gather metrics.


ssalinas:
🚢

PR #2188, opened 2021-03-17
Thread and stream clean up

ssalinas:
- Don't start all the threads at once on sized pools (non-scheduler instances were sitting with hundreds of them open and unused)
- Clean up open streams on failed download calls. It looked like we possibly had a leak previously, keeping these open. Unclear if that was the case or not and need to test


rosalind210:
🚢

PR #2189, opened 2021-03-18
Run a final logrotate on hourly and size based files as well

ssalinas:
cc @baconmania @pschoenfelder these need to trigger in order for glob matchers on the rotated file in the uploader to pick them up


ssalinas:
Folded into #2195 to fix conflicts

PR #2194, opened 2021-04-04
Wire up the ability for SingularityExecutor to compress arbitrary sandbox files

baconmania:
An example SingularityExecutor config that leverages this feature might look like:
```
executor:
  compressAdditionalFiles:
  - directory: logs
    filenameGlob: some.rotated.log.*
    compressionType: GZIP
```


ssalinas:
If we want the uploader to upload compressed, would it make more sense just to give the uploader a compressBeforeUpload: true option? Otherwise have to wait for executor cleanup cron after any failed/oom'd task before logs would show up since uncompressed won't match the uploader glob

baconmania:
I thought about compressing in the uploader, but I didn't want to I introduce any additional complexity or CPU-boundedness to S3Uploader. Keeping it as simple as it is now feels like a feature of the component.

That said, it's a good point that failed tasks would introduce a delay for the final compress + upload cycle. 🤔

ssalinas:
closing for #2195

PR #2195, opened 2021-04-05
Optionally compress files prior to uploading them in SingularityS3Uploader.

baconmania:


SingularityS3Uploader/src/main/java/com/hubspot/singularity/s3uploader/SingularityUploader.java
```diff
@@ -255,7 +257,38 @@ private AtomicInteger handleFile(Path path, boolean isFinished, List<Path> toUpl
 
     found.incrementAndGet();
 
-    toUpload.add(path);
+    if (uploadMetadata.isCompressBeforeUpload() && !path.toString().endsWith(".gz")) {
```
ssalinas, line 260: if we are checking for 'isAlreadyGzipped' here, should we also be checking that in the file glob section? E.g. if something else already zipped it, it should still match the toUpload glob as well

ssalinas: nevermind, stared down the section above a bit more, we are catching those other cases

ssalinas:
🚢

tpetr:
:ship:

PR #2196, opened 2021-04-06
Also check if a task is a killed task in rack placements

ssalinas:
We would previously wait until tasks that were cleaned -> killed -> still shutting down exited before actually scheduling a new pending task due to rack placement issues. This also removes tasks that are already killed from the counts fo rack placement as well as those in cleaning state


pschoenfelder:
🚢

PR #2198, opened 2021-04-08
S3Uploader aws client usage efficiency

ssalinas:
Previously we spun up one client per S3Uploader object, since they often have separate creds. However, many are shared, and duplicates become a problem when we start hitting thousands of uploader objects.

Open questions:
- Should there be some sort of cool down period before I junk a client so we aren't recreating too often?
- Keying on the string concat of creds feels dirty, but BasicAwsCredentials doesn't implement hashCode, thoughts? It's not logged anywhere, and the BasicAwsCreds are already in memory like that elsewhere obviously, just felt weird to write 🤷 

cc @baconmania


baconmania:
🚢

PR #2199, opened 2021-04-08
Add endpoint to get Singularity config values

rosalind210:
This PR exposes Singularity's `maxDecommissioningAgents`, defined in the config file.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityLimits.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityLimits.java
@@ -1,0 +1,13 @@
+package com.hubspot.singularity;
+
+public class SingularityLimits {
+  private final int maxDecommissioningAgents;
+
+  public SingularityLimits(int maxDecommissioningAgents) {
+    this.maxDecommissioningAgents = maxDecommissioningAgents;
+  }
+
+  public int getMaxDecommissioningAgents() {
+    return maxDecommissioningAgents;
+  }
+}
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityLimits.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityLimits.java
@@ -1,12 +1,21 @@
 package com.hubspot.singularity;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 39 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/AgentResource.java
```diff
@@ -173,6 +173,13 @@ public void decommissionAgent(
     );
   }
 
+  @GET
+  @Path("/agent/max-decommissioning-count")
```
pschoenfelder, line 177: This isn't really agent specific, and unfortunately doesn't fit neatly with any of our other resources either. Probably worth creating a separate "config" resource class to expose this.

rosalind210: haha I was considering doing that but then thought a new resource would be overkill -- I'll make a new one

SingularityService/src/main/java/com/hubspot/singularity/resources/SingularityConfigurationResource.java
```diff
@@ -0,0 +1,28 @@
...
+public class SingularityConfigurationResource {
+  private final SingularityConfiguration config;
+
+  @Inject
+  public SingularityConfigurationResource(SingularityConfiguration config) {
+    this.config = config;
+  }
+
+  @GET
+  @Path("/singularity-limits-format")
```
pschoenfelder, line 23: Nitpick — can we change this to just `/singularity-limits`?

SingularityBase/src/main/java/com/hubspot/singularity/SingularityLimits.java
```diff
@@ -0,0 +1,13 @@
+package com.hubspot.singularity;
+
+public class SingularityLimits {
+  private final int maxDecommissioningAgents;
+
+  public SingularityLimits(int maxDecommissioningAgents) {
+    this.maxDecommissioningAgents = maxDecommissioningAgents;
+  }
```
ssalinas, line 8: Since we don't have immutables set up in Singularity, this needs a `@JsonCreator` annotation on the constructor and @JsonProperty("maxDecommissioningAgents")` on the field. See other things in the Base module for examples

pschoenfelder:
🚢

pschoenfelder:
🚢

PR #2200, opened 2021-04-09
One additional null check in uploader client return

ssalinas:
Not impacting performance, just avoiding the err messages


rosalind210:
🚢

PR #2201, opened 2021-04-12
Update s3 client builders

ssalinas:
Getting off the deprecated methods so these pull in default regions from the environment correctly


ssalinas:
didn't end up being a fix for anything, but some deprecation fixes we should clean up at some point still

PR #2205, opened 2021-04-27
Refactor the LoadBalancerClient

ssalinas:
Take all of the baragon-specific bits that have leaked out over time and put them all back into the BaragonLoadBalancerClientImpl. This will allow for a cleaner secondary/outside implementation since it only requires interaction with singularity classes and not also baragon ones


ssalinas:
PR also ended up having some unit test tuning since the flaky builds were getting annoying

ajammala:
🚢

PR #2207, opened 2021-05-26
Add extension to agent decom

rosalind210:
We have been experiencing decommissioned agents coming back online too early. This endpoint will allow us to extend the decommission time if the agent isn't ready.

### DIFF
--- a/SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
+++ b/SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
@@ -136,6 +136,8 @@ public class SingularityClient {
   private static final String AGENT_DETAIL_FORMAT = AGENTS_FORMAT + "/agent/%s/details";
   private static final String AGENTS_DECOMISSION_FORMAT =
     AGENTS_FORMAT + "/agent/%s/decommission";
+  private static final String EXTEND_AGENTS_DECOMISSION_FORMAT =
+    AGENTS_FORMAT + "/agent/%s/decommission/extend";
   private static final String AGENTS_FREEZE_FORMAT = AGENTS_FORMAT + "/agent/%s/freeze";
   private static final String AGENTS_ACTIVATE_FORMAT =
     AGENTS_FORMAT + "/agent/%s/activate";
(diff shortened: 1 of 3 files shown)


SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
```diff
@@ -136,6 +136,8 @@
   private static final String AGENT_DETAIL_FORMAT = AGENTS_FORMAT + "/agent/%s/details";
   private static final String AGENTS_DECOMISSION_FORMAT =
     AGENTS_FORMAT + "/agent/%s/decommission";
+  private static final String EXTEND_AGENTS_DECOMISSION_FORMAT =
+    AGENTS_FORMAT + "/agent/%s/decommission/extend";
```
pschoenfelder, line 140: Probably want to change language from `extend` to `update` since it's a more generic operation.

SingularityService/src/main/java/com/hubspot/singularity/resources/AgentResource.java
```diff
@@ -173,6 +173,47 @@ public void decommissionAgent(
...
+      }
+    );
+  }
+
+  public void extendDecommissionAgent(
+    SingularityUser user,
+    String agentId,
+    SingularityMachineChangeRequest changeRequest
+  ) {
+    authorizationHelper.checkAdminAuthorization(user);
```
pschoenfelder, line 206: Someone could effectively reactivate or permanently decom an agent with this update, so we should also check `validator.checkActionEnabled(action)` for actions `SingularityAction.ACTIVATE_AGENT` and `SingularityAction.DECOMMISSION_AGENT` just to be safe here.

rosalind210: Ok, I can do that. The validator is private to AbstractMachineResource, do you think I should change it to protected or move this function to  AbstractMachineResource?

pschoenfelder: I'd follow whatever pattern the base decom method uses (I think move into AbstractMachineResource)

pschoenfelder:
🚢

PR #2208, opened 2021-06-02
Add taskId to loadbalancer upstream

jonathanwgoodwin:
Allows partitioning loadbalancer upstreams per deploy.

@pschoenfelder

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/LoadBalancerUpstream.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/LoadBalancerUpstream.java
@@ -11,23 +11,27 @@ public class LoadBalancerUpstream {
   private final String upstream;
   private final String group;
   private final Optional<String> rackId;
+  private final Optional<String> taskId;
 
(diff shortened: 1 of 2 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/LoadBalancerUpstream.java
```diff
@@ -43,6 +47,10 @@ public String getGroup() {
     return rackId;
   }
 
+  public Optional<String> getTaskId() {
+    return taskId;
+  }
+
   @Override
   public boolean equals(Object o) {
```
jonathanwgoodwin, line 55: I didn't update HashCode / Equals since the absence / presence of this new data doesn't really impact the upstream itself; its just a reference we can use to group things by singularity deploy at the load balancing layer later on. 

Since this data is likely persisted in this form without the value present adding HC / Equals would result in a change to those comparisons and likely break stuff.

pschoenfelder:
🚢

PR #2210, opened 2021-06-03
Check against artifacts without signatures

pschoenfelder:
Adding an option to require that artifacts have signatures.

### DIFF
--- a/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
+++ b/SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
@@ -189,7 +189,7 @@ public class SingularityExecutorConfiguration extends BaseRunnerConfiguration {
   private boolean failTaskOnInvalidArtifactSignature = true;
 
   @JsonProperty
-  private boolean failTaskOnMissingArtifactSignature = false;
+  private boolean failOnArtifactWithNoMatchingSignature = false;
 
   private boolean failOnSignatureWithNoMatchingArtifact = false;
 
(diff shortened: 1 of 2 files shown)


SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -188,6 +188,11 @@
   @JsonProperty
   private boolean failTaskOnInvalidArtifactSignature = true;
 
+  @JsonProperty
+  private boolean failOnArtifactWithNoMatchingSignature = false;
+
+  private boolean failOnSignatureWithNoMatchingArtifact = false;
```
rosalind210, line 194: Why doesn't this field need the `@JsonProperty` annotation?

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
```diff
@@ -602,6 +605,16 @@ public void setFailTaskOnInvalidArtifactSignature(
     this.failTaskOnInvalidArtifactSignature = failTaskOnInvalidArtifactSignature;
   }
 
+  public boolean isFailOnArtifactWithNoMatchingSignature() {
+    return failOnArtifactWithNoMatchingSignature;
+  }
+
+  public void seFailOnArtifactWithNoMatchingSignature(
```
rosalind210, line 612: I think you meant `setFailOnArtifactWithNoMatchingSignature`

SingularityExecutor/src/main/java/com/hubspot/singularity/executor/config/SingularityExecutorConfiguration.java
pschoenfelder: Not sure, just moved that one up here to be with the others.

tpetr:
:ship:

PR #2214, opened 2021-06-24
Replace caffeine cache with reference to full map

rosalind210:
Originally had the CaffeineCache set up in the RequestResource by user and call parameters, but the return values couldn't be cached across all users because of differing authorization and settings. We ended up still experiencing 504s when SingularityClient's `getRequests` got hit hard again. We decided to split up the cache to cover the heaviest ZK calls in the `getRequest` method: `RequestManager`'s `fetchRequests` and `DeployManager`'s `fetchDeployStatesByRequestIds`, which don't have caching around client calls, but then there was a performance degradation in the `getRequest` call because of the cache expiry.

Now we are replacing caffeine cache altogether with an AtomicReference to the full map of requests and deploys that are updated every ~~second~~ five seconds by a different thread. The update to five seconds came after overloading the CuratorFramework with every second calls for request and deploy data.

cc: @jschlather


SingularityService/src/main/java/com/hubspot/singularity/data/ManagerCache.java
```diff
@@ -0,0 +1,40 @@
...
+    if (values != null) {
+      LOG.trace("Grabbed values for {} from cache", key);
+    } else {
+      LOG.trace("{} not in cache", key);
+    }
+
+    return values;
+  }
+
+  public void put(K key, V value) {
```
jschlather, line 36: I think we want to use a loader method. If we do it this way, then we won't get any debouncing from the cache.

SingularityService/src/main/java/com/hubspot/singularity/data/ManagerCache.java
```diff
@@ -0,0 +1,58 @@
...
+    return values;
+  }
+
+  public Map<K, V> getAll(@Nonnull Iterable<? extends K> keys) {
+    Map<K, V> values = cache.getAll(keys);
+    if (values.isEmpty()) {
+      LOG.trace("Grabbed mapped values for {} from cache", keys);
+    }
+
+    return values.isEmpty() ? null : values;
```
pschoenfelder, line 52: I think it's more idiomatic to return an empty map instead of null

SingularityService/src/main/java/com/hubspot/singularity/data/DeployManager.java
```diff
@@ -99,6 +102,17 @@ public DeployManager(
...
+    this.deployCache =
+      new ManagerCache<>(
+        configuration,
+        requestId ->
+          getAsync(
+              "getRequestDeployStatesByRequestIds",
+              Collections.singletonList(getRequestDeployStatePath(requestId)),
+              requestDeployStateTranscoder
+            )
+            .get(0)
```
pschoenfelder, line 114: What happens if this throws an IndexOutOfBoundsException?

SingularityService/src/main/java/com/hubspot/singularity/data/ManagerCache.java
rosalind210: Both caches are now LoadingCaches.

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -2084,11 +2085,19 @@ public void setUseCaffeineCache(boolean useCaffeineCache) {
...
 
-  public int getCaffeineCacheTtl() {
-    return caffeineCacheTtl;
+  public int getDeployCaffeineCacheTtl() {
+    return deployCaffeineCacheTtl;
   }
 
-  public void setCaffeineCacheTtl(int caffeineCacheTtl) {
-    this.caffeineCacheTtl = caffeineCacheTtl;
+  public void setDeployCaffeineCacheTtl(int deployCaffeineCacheTtl) {
```
pschoenfelder, line 2092: Nit — can we remove `caffeine` from the terminology here? Users don't need to be aware of cache choice/impl, and we could change caches in the future if we wanted to without any awkwardness

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -2084,11 +2085,19 @@ public void setUseCaffeineCache(boolean useCaffeineCache) {
...
+    this.deployCaffeineCacheTtl = deployCaffeineCacheTtl;
+  }
+
+  public int getRequestCaffeineCacheTtl() {
+    return requestCaffeineCacheTtl;
```
pschoenfelder, line 2097: One last nit — can we append the time unit to this so users don't have to search code for it?

SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
rosalind210: I've also updated the `useCaffeineCache` to `useZKFastCache`

jschlather:
LGTM

SingularityService/src/main/java/com/hubspot/singularity/data/ManagerCache.java
```diff
@@ -0,0 +1,52 @@
...
+import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.google.inject.Inject;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ManagerCache<K, V> {
```
ssalinas, line 13: Check out the ZkCache class. It's basically the same thing as this and might be better suited (at least for the deployManager part of things)

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -632,11 +639,20 @@ public void activateLeaderCache() {
     if (useWebCache && webCache.useCachedRequests()) {
       return webCache.getRequests();
     }
+
+    if (requestsCache.isEnabled()) {
+      List<SingularityRequestWithState> requests = requestsCache.get("all");
+      if (requests != null) {
+        return requests;
+      }
+    }
```
ssalinas, line 648: I wonder if we should just not use the web cache here if this is the case. Seems overkill to have the leaderCache -> web cache -> zk cache all here. Lots of duplicate memory usage to store it all 3 times

SingularityService/src/main/java/com/hubspot/singularity/data/ManagerCache.java
rosalind210: The ZkCache just uses a Guava cache and we were hoping to get benefits from the LoadingCache created by CaffeineCache for debouncing requests because we were also getting timeouts at the DeployManager's request, do you still think that I should update the DeployManager's usage?

ssalinas: Ah ok, thought the Zk one was more similar. Can keep the separate one then

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
rosalind210: As discussed offline, to have ease in revert, the removal of web cache will be in a separate PR

SingularityService/src/main/java/com/hubspot/singularity/data/ApiCache.java
```diff
@@ -0,0 +1,77 @@
...
+    int cacheTtl,
+    Supplier<Map<K, V>> supplyMap,
+    ScheduledExecutorService executor
+  ) {
+    this.isEnabled = isEnabled;
+    this.supplyMap = supplyMap;
+    this.zkValues = new AtomicReference<>(new HashMap<>());
+
+    if (this.isEnabled) {
+      executor.scheduleAtFixedRate(this::reloadZkValues, 2, cacheTtl, TimeUnit.SECONDS);
```
rosalind210, line 35: Put in a 2 second delay because the reloading thread kept starting before the CuratorFrameworkImpl started

ssalinas: would it make more sense to try and do the first call to these inside some of the Managed classses that run during startup. Then we can:
- Make sure that the first fetch of this runs after curator has started
- Avoid writing zk calls in a constructor context vs some other method on an already instantiated class
- Have the cached filled before the service actually starts listening for requests
Thinking something like a start() method that we can call somewhere in SingularityLifecycleManaged

SingularityService/src/main/java/com/hubspot/singularity/data/DeployManager.java
```diff
@@ -99,6 +103,13 @@ public DeployManager(
     this.updateRequestTranscoder = updateRequestTranscoder;
     this.deploysCache = deploysCache;
     this.leaderCache = leaderCache;
+    this.deployCache =
+      new ApiCache<>(
+        configuration.useApiCacheInDeployManager(),
+        configuration.getDeployCacheTtl(),
+        this::fetchAllDeployStates,
+        executorServiceFactory.get("deploy-api-cache-reloader")
```
ssalinas, line 111: I _think_ there is a version of this that returns a single threaded executor, which is all we should need for this case

rosalind210: We don't have an option for a single threaded ScheduledExecutorService in SingularityManagedScheduledExecutorServiceFactory but Executors has newSingleThreadScheduledExecutor that I can add.

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -90,6 +93,7 @@
   );
 
   private final Map<Class<? extends SingularityExpiringRequestActionParent<? extends SingularityExpiringRequestParent>>, Transcoder<? extends SingularityExpiringRequestActionParent<? extends SingularityExpiringRequestParent>>> expiringTranscoderMap;
+  private final ApiCache<String, SingularityRequestWithState> requestsCache;
```
ssalinas, line 96: shoudl the `getRequests(List<String> requestIds)` and the singular `getRequest(String requestId)` also use the cache? Doesn't look like they are at the moment

SingularityService/src/main/java/com/hubspot/singularity/data/ApiCache.java
```diff
@@ -0,0 +1,94 @@
...
+    ScheduledExecutorService executor
+  ) {
+    this.isEnabled = isEnabled;
+    this.supplyMap = supplyMap;
+    this.zkValues = new AtomicReference<>(new HashMap<>());
+    this.cacheTtl = cacheTtl;
+    this.executor = executor;
+  }
+
+  public void startReloader() {
```
ssalinas, line 43: should this also synchronously perform the first fetch I wonder? Would avoid the case where getAll was incorrectly empty or get(K key) was null

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
rosalind210: I didn't update elsewhere yet because I wasn't sure if that's what we wanted to do since we were most concerned by the endpoint to get all requests

ssalinas: since it all pulls form the same place, and we are constnatly updating everything, I think it'd be worth it to update. I believe that the individual request endpoint was also pretty high up on the usage. Can always be a follow up PR if we want to check how effective this is first too

rosalind210: I added the cache to the getRequests(List<String> requestIds) and the singular getRequest(String requestId), and the singular request has a non-cache flag now for Orion usage.

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -668,6 +736,13 @@ public void activateLeaderCache() {
       return webCache.getRequest(requestId);
     }
 
+    if (requestsCache.isEnabled() && !skipApiCache) {
+      SingularityRequestWithState request = requestsCache.get(requestId);
+      if (request != null) {
+        return Optional.of(request);
```
rosalind210, line 742: I didn't make this `Optional.ofNullable(...)` in case it was a false null, aka the request came in between the 5 second reload, but this means that we will go to ZK for true nulls (the request was removed).

SingularityService/src/main/java/com/hubspot/singularity/data/RequestManager.java
```diff
@@ -544,6 +567,15 @@ public SingularityDeleteResult markDeleted(
       }
     }
 
+    if (requestsCache.isEnabled()) {
+      List<SingularityRequestWithState> requests = new ArrayList<>(
+        (requestsCache.getAll()).values()
```
ssalinas, line 572: nit, extra parens here?

ssalinas (commented):
One overall comment for zk efficiency. Do we want to have the reloader short circuit if the current instance is the leader (i.e. will have the leader cache instead)? I don't think there are any cases where we would hit this over the leader cache, so would likely want to limit the calls flooding the leader CuratorFramework

SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
```diff
@@ -408,6 +410,11 @@ public SingularityClient(
         .build();
   }
 
+  @Deprecated
+  public void setSkipApiCache(boolean skipApiCache) {
+    this.skipApiCache = skipApiCache;
+  }
```
ssalinas, line 416: for structure/usability of the client. Is there a reason you have this at the class level instead of just being an arg on the relevant method(s)?

SingularityService/src/main/java/com/hubspot/singularity/data/ApiCache.java
```diff
@@ -0,0 +1,127 @@
...
+
+    return value;
+  }
+
+  public Map<K, V> getAll() {
+    Map<K, V> allValues = this.zkValues.get();
+    if (allValues.isEmpty()) {
+      LOG.debug("ApiCache getAll returned empty");
+    } else {
+      LOG.debug("getAll returned {} values", allValues.size());
```
ssalinas, line 99: for this and the debug statement above, maybe these are more like TRACE level lines? Would get pretty noisy given that we could acall these multiple times a second

SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
rosalind210: I looked through the request call usages in Orion and talked to Suruu and we decided it would be easier on Deploy's side to have a class level argument rather than updating all usages. It also makes clean up easier for when we've solved the underlying CuratorFramework issue

SingularityService/src/main/java/com/hubspot/singularity/SingularityManagedScheduledExecutorServiceFactory.java
```diff
@@ -66,6 +66,29 @@ public synchronized ScheduledExecutorService get(
     return service;
   }
 
+  public synchronized ScheduledExecutorService getSingleThreaded(String name) {
+    return getSingleThreaded(name, false);
+  }
+
+  public synchronized ScheduledExecutorService getSingleThreaded(
```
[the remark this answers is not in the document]
rosalind210: I was following the pattern in the SingularityManagedThreadPoolFactory for `getSingleThreaded` but I'll update to not overload that method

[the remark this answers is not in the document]
ssalinas: 🤦 just realzied the one above calls it. Nevermind didn't even see that

SingularityService/src/main/java/com/hubspot/singularity/data/ApiCache.java
rosalind210: I made all of the get calls' logs trace level.

ssalinas:
🚢

PR #2220, opened 2021-07-20
Add request deploy history endpoint to SingularityClient

pschoenfelder:


SingularityClient/src/main/java/com/hubspot/singularity/client/SingularityClient.java
```diff
@@ -2109,6 +2112,31 @@ public void clearInactiveAgent(String host) {
     );
   }
 
+  public Collection<SingularityDeployHistory> getRequestDeployHistory(
```
rosalind210, line 2115: nit: can you make this function name plural so that it's clear that it returns multiple?

rosalind210:
🚢

PR #2224, opened 2021-08-12
Smarter timeouts in offer loop

ssalinas:
stop if the total time is too long or the time processing a single request is too long

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
@@ -545,11 +545,17 @@ public class SingularityMesosOfferScheduler {
                         long startRequest = System.currentTimeMillis();
                         int evaluated = 0;
                         for (SingularityTaskRequestHolder taskRequestHolder : entry.getValue()) {
+                          long now = System.currentTimeMillis();
                           if (
-                            System.currentTimeMillis() -
-                            startRequest >
-                            TimeUnit.SECONDS.toMillis(30) && // TODO - make configurable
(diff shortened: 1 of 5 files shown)


SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -999,7 +1018,11 @@ private double calculateScore(
     return score;
   }
 
-  private SingularityMesosTaskHolder acceptTask(
+  // This method is synchronized to avoid resource reuse within a single offer.
+  // Decisions about resources to use, specifically ports are made during the buildTask call, however
+  // these resources aren't subtracted from the offer until the addMatchedTask call, making this method
+  // not thread safe
+  private synchronized void acceptTask(
```
WH-77, line 1025: Would it be possible/preferable to synchronize on the offerHolder instead of the method, to avoid blocking acceptTask calls that are using different offers?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosOfferScheduler.java
```diff
@@ -541,7 +542,29 @@ private boolean isValidOffer(Offer offer) {
...
+                          if (
+                            now -
+                            startCheck >
+                            mesosConfiguration.getOfferLoopTimeoutMillis() ||
+                            (
+                              now -
+                              startRequest >
+                              mesosConfiguration.getOfferLoopRequestTimeoutMillis() &&
+                              evaluated > 1
+                            )
```
WH-77, line 558: nit - would be more readable with variables for each condition

PR #2226, opened 2021-08-19
Configurably skip the file open check in S3Uploader.

baconmania:
This would be exercised by adding a `checkIfOpen: false` to an `s3UploaderAdditionalFiles` entry in the `SingularityS3UploaderConfiguration`.

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityS3UploaderFile.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityS3UploaderFile.java
@@ -18,11 +18,13 @@ public class SingularityS3UploaderFile {
   private final Optional<Long> applyS3StorageClassAfterBytes;
   private final boolean checkSubdirectories;
   private final boolean compressBeforeUpload;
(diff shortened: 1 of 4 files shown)


SingularityBase/src/main/java/com/hubspot/singularity/SingularityS3UploaderFile.java
```diff
@@ -57,6 +60,7 @@ public SingularityS3UploaderFile(
     this.applyS3StorageClassAfterBytes = applyS3StorageClassAfterBytes;
     this.checkSubdirectories = checkSubdirectories.orElse(false);
     this.compressBeforeUpload = compressBeforeUpload.orElse(false);
+    this.checkIfOpen = checkIfOpen.orElse(false);
```
ssalinas, line 63: should this be `true` if the default is true?

baconmania: woof, good catch

ssalinas:
🚢

PR #2228, opened 2021-08-26
Support global overrides of request placement strategy/rack sensitivity

WH77:
This PR exists to speed up response/recovery time in the event of a rack outage, via new API endpoints that will set global configuration overrides of request placement strategy and/or rack sensitivity on the current leader.


SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -2063,7 +2063,7 @@ public void setUseLoggingCuratorFramework(boolean useLoggingCuratorFramework) {
   }
 
   public boolean isOptInEmailMode() {
-    return optInEmailMode;
+    return true;
```
pschoenfelder, line 2066: Why this change?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityMesosStatusUpdateHandler.java
```diff
@@ -308,7 +308,7 @@ private StatusUpdateResult unsafeProcessStatusUpdate(
     long now = System.currentTimeMillis();
     long delta = now - timestamp;
 
-    LOG.debug(
+    LOG.info(
```
pschoenfelder, line 311: How much additional logging is this going to be? With the number of tasks running could be overwhelming to have at info level. Same below

pschoenfelder:
🚢

PR #2229, opened 2021-09-10
Allow skipping file open check during S3 upload

pschoenfelder:


pschoenfelder:
Closing because duplicate

PR #2230, opened 2021-09-21
Fix possible port collisions across tasks again

WH77:
Attempts to address the same issue as https://github.com/HubSpot/Singularity/pull/2215.

Current logging hasn't uncovered a smoking gun, but there appears to a potential source of duplicated ports when offers are combined in SingularityOfferHolder. The most likely alternative cause would be stale offers in the cache, which would require a much heavier fix. For now, going to put this out there with additional logging around resource offers to help determine whether the cache or resource duplication is the root cause.


pschoenfelder:
Looks solid. How important are those TODO's in the unit tests?

WH77:
I don't believe they're the cause of the port allocation problem, they're more of a note that MesosUtils expects to be given valid resources to subtract.

PR #2233, opened 2021-09-22
Better handling of errors in scheduler poller

WH77:
Currently if the scheduler poller encounters an unexpected error it causes Singularity as a whole to abort. This PR changes the scheduler poller to try again in the event of unexpected errors, as well as modifying the scheduled task poll action to log errors for specific failed tasks instead of failing the entire batch.

This approach may swallow too much, but there are metrics that should still inform us if something is wrong with the scheduler as a whole.

cc - @pschoenfelder

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
@@ -359,7 +359,7 @@ public class SingularityScheduler {
             () ->
               lock.runWithRequestLock(
                 () ->
-                  handlePendingRequestsForDeployKey(
+                  handlePendingRequestsForDeployKeySafe(
                     obsoleteRequests,
                     heldForScheduledActiveTask,
                     totalNewScheduledTasks,
@@ -382,6 +382,26 @@ public class SingularityScheduler {
       heldForScheduledActiveTask.get(),
       JavaUtils.duration(start)
     );
+  }
+
+  private void handlePendingRequestsForDeployKeySafe(
+    AtomicInteger obsoleteRequests,
+    AtomicInteger heldForScheduledActiveTask,
+    AtomicInteger totalNewScheduledTasks,
+    SingularityDeployKey deployKey,
+    List<SingularityPendingRequest> pendingRequestsForDeploy
+  ) {
+    try {
+      handlePendingRequestsForDeployKey(
+        obsoleteRequests,
+        heldForScheduledActiveTask,
+        totalNewScheduledTasks,
+        deployKey,
+        pendingRequestsForDeploy
+      );
+    } catch (Exception e) {
+      LOG.error("Error handling pending requests for {}, skipping", deployKey, e);
+    }
   }
 
   private void handlePendingRequestsForDeployKey(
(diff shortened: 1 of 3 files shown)


SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -3841,6 +3842,12 @@ public void testCronScheduleChanges() throws Exception {
     );
   }
 
+  @Test
+  public void testBrokenQuartzCron() throws Exception {
+    String quartz = "0 0/60 * * * ?";
+    new CronExpression(quartz);
+  }
+
```
pschoenfelder, line 3850: Not sure how useful this test is in the sense that this is technically not a valid expression, but library maintainers have made it work in the past and since decided to change that behavior. Maybe add a note about the quartz upgrade for background here in case this starts failing on a future upgrade?

pschoenfelder:
Have you tested this yet?

WH77:
No, figured I'd wait until after we decided whether to move forward with this approach.

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularitySchedulerPoller.java
```diff
@@ -72,4 +72,9 @@ public void runActionOnPoll() {
       "SingularitySchedulerPoller"
     );
   }
+
+  @Override
+  protected boolean abortsOnError() {
+    return false;
```
tpetr, line 78: why set this to `false`?

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -3424,6 +3424,61 @@ public void testRequestedPorts() {
     Assertions.assertEquals(1, taskManager.getActiveTaskIds().size());
   }
 
+  @Test
+  public void testPortUsage() {
```
tpetr, line 3428: this test doesnt seem related to the PR -- can we move it to a separate one if it's still necessary?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularitySchedulerPoller.java
WH77: To avoid killing the entire leader if the scheduler poller runs into an unexpected exception during a poll - currently it aborts at https://github.com/HubSpot/Singularity/blob/1415dea99de62cd0e39f4b3196d74b1feff5e6c7/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityLeaderOnlyPoller.java#L150

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityScheduler.java
```diff
@@ -384,6 +390,27 @@ public void drainPendingQueue() {
...
+      handlePendingRequestsForDeployKey(
+        obsoleteRequests,
+        heldForScheduledActiveTask,
+        totalNewScheduledTasks,
+        deployKey,
+        pendingRequestsForDeploy
+      );
+    } catch (Exception e) {
+      LOG.error("Error handling pending requests for {}, skipping", deployKey, e);
+      unscheduledTasksMetric.mark();
```
WH77, line 410: Debated just reusing the lost tasks meter here, could be misleading (same task would get "lost" repeatedly) but would be fairly correct in spirit - @pschoenfelder you have any preference here?

PR #2235, opened 2021-09-23
Avoid null next value in zk curator load distributor

WH77:
cc - @pschoenfelder

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/SingularityCuratorProvider.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/SingularityCuratorProvider.java
@@ -50,11 +50,15 @@ public class SingularityCuratorProvider implements Provider<CuratorFramework> {
     ZooKeeperConfiguration zookeeperConfig = configuration.getZooKeeperConfiguration();
 
     if (configuration.isReadOnlyInstance()) {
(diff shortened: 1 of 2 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/curator/ZkClientsLoadDistributor.java
```diff
@@ -41,7 +40,8 @@ public void close() {
   }
 
   public CuratorFramework getCuratorFramework() {
-    return iterator.next();
+    int ci = curatorIndex.getAndUpdate(i -> (i + 1) % curatorFrameworks.size());
```
WH77, line 43: might've been able to get away without using an AtomicInteger here, but don't think it makes a huge difference either way

SingularityService/src/main/java/com/hubspot/singularity/SingularityCuratorProvider.java
```diff
@@ -51,7 +51,10 @@ private CuratorFramework getCuratorFrameworkForSingularityInstanceType(
 
     if (configuration.isReadOnlyInstance()) {
       LOG.trace("Creating multiple logging curator frameworks for read-only instance");
```
pschoenfelder, line 53: Might want to update this to reflect the actual value used in the line below

PR #2237, opened 2021-10-18
Limit # of crash loop nodes for requests without requiring new deploys

WH77:
Currently, crash loop nodes for requests that repeatedly create/resolve crash loops are not cleaned up until the request has a new deploy., which causes very slow but unbounded growth in ZooKeeper. This PR ensures that old resolved crash loop nodes are deleted instead of being skipped.

cc - @pschoenfelder


pschoenfelder:
Looks good to me 🚢

PR #2239, opened 2021-10-28
Improve Offer Loop Usage of Request Locks

WH77:
This PR attempts to reduce the consequences (ie scheduling lag) of attempting to schedule a task whose request lock is locked elsewhere for a long period of time (ie history persisters).

Changes:
- Actually make offer scoring parallel while waiting on request locks. The offer scoring thread pool uses an unbounded queue, so it would never grow past the core pool size of 1 / was still single threaded.
- Limit the amount of time that the offer loop will wait for a request lock - less relevant now that the thread pool can have more than 1 thread, but would still help if the thread pool was saturated with locked requests.

Took a very quick/lazy approach to the tryRunWithLock in the offer loop, so please send feedback - should it throw an error, log whether it was able to run, etc.

Considered attempting a larger scale refactor (poll from a single priority queue of pending tasks, remove request locks, etc), but that seemed very likely to introduce new and exciting bugs. The fact that there's separate `offer-scheduler` and `SingularitySchedulerPoller` threads that schedule tasks slightly differently (drainPendingQueue) is annoying, but doesn't seem to be the main source of problems.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
@@ -264,6 +264,11 @@ public class DeployResource extends AbstractRequestResource {
       // This can cause a conflict if run outside the lock, causing the pending deploy to be checked before deploy data is saved
       schedulerLock.runWithRequestLock(
         () -> {
+          deployManager.createDeployIfNotExists(
+            updatedRequest,
+            deployMarker,
+            validatedDeploy
+          );
           deployAlreadyInProgress.set(
             deployManager.createPendingDeploy(pendingDeployObj) ==
             SingularityCreateResult.EXISTED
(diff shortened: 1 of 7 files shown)


SingularityService/src/main/java/com/hubspot/singularity/resources/DeployResource.java
```diff
@@ -264,6 +264,11 @@ public SingularityRequestParent deploy(
       // This can cause a conflict if run outside the lock, causing the pending deploy to be checked before deploy data is saved
       schedulerLock.runWithRequestLock(
         () -> {
+          deployManager.createDeployIfNotExists(
+            updatedRequest,
+            deployMarker,
+            validatedDeploy
+          );
```
ssalinas, line 271: Why do this before the deploy-already-in-progress check? We would already save this data right afterwards in saveDeploy anyway

ssalinas: nevermind, I see what this is doing now, pending deploy gets created slightly before the actual deploy object. without it

ssalinas (approved):

ssalinas:
🚢

PR #2240, opened 2021-11-09
Cleanup hosts marked as missing on startup

ashdza:
This PR modifies the `SingularityAgentReconciliationPoller` to delete hosts that are marked as missing on startup. Previously, only hosts marked as dead were deleted by the poller, causing inactive hosts marked as missing on startup to still appear in Singularity.

Hosts that are dead or missing on startup are deleted from both the `AgentManager` and `InactiveAgentManager` lists to cleanup inactive hosts in the Singularity UI.

https://git.hubteam.com/HubSpot/PaaS-Run/issues/1525

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityAgentReconciliationPoller.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityAgentReconciliationPoller.java
@@ -55,7 +55,7 @@ public class SingularityAgentReconciliationPoller extends SingularityLeaderOnlyP
   @Override
   public void runActionOnPoll() {
     refereshSlavesAndRacks();
-    checkDeadSlaves();
+    checkInactiveSlaves();
     inactiveAgentManager.cleanInactiveAgentsList(
       System.currentTimeMillis() -
       TimeUnit.HOURS.toMillis(configuration.getCleanInactiveHostListEveryHours())
@@ -79,15 +79,19 @@ public class SingularityAgentReconciliationPoller extends SingularityLeaderOnlyP
     }
   }
 
-  private void checkDeadSlaves() {
+  private void checkInactiveSlaves() {
     final long start = System.currentTimeMillis();
 
-    final List<SingularityAgent> deadSlaves = agentManager.getObjectsFiltered(
+    final List<SingularityAgent> inactiveSlaves = agentManager.getObjectsFiltered(
       MachineState.DEAD
     );
 
-    if (deadSlaves.isEmpty()) {
-      LOG.trace("No dead agents");
+    inactiveSlaves.addAll(
+      agentManager.getObjectsFiltered(MachineState.MISSING_ON_STARTUP)
+    );
+
+    if (inactiveSlaves.isEmpty()) {
+      LOG.trace("No inactive agents");
       return;
     }
 
@@ -96,20 +100,18 @@ public class SingularityAgentReconciliationPoller extends SingularityLeaderOnlyP
       configuration.getDeleteDeadAgentsAfterHours()
     );
 
-    for (SingularityAgent deadSlave : agentManager.getObjectsFiltered(
-      MachineState.DEAD
-    )) {
+    for (SingularityAgent inactiveSlave : inactiveSlaves) {
       final long duration =
-        System.currentTimeMillis() - deadSlave.getCurrentState().getTimestamp();
+        System.currentTimeMillis() - inactiveSlave.getCurrentState().getTimestamp();
 
       if (duration > maxDuration) {
-        SingularityDeleteResult result = agentManager.deleteObject(deadSlave.getId());
+        SingularityDeleteResult result = agentManager.deleteObject(inactiveSlave.getId());
 
         deleted++;
 
         LOG.info(
-          "Removing dead agent {} ({}) after {} (max {})",
-          deadSlave.getId(),
+          "Removing inactive agent {} ({}) after {} (max {})",
+          inactiveSlave.getId(),
           result,
           JavaUtils.durationFromMillis(duration),
           JavaUtils.durationFromMillis(maxDuration)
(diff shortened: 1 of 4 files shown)


SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityMachinesTest.java
```diff
@@ -117,6 +116,74 @@ public void testDeadSlavesArePurged() {
     Assertions.assertEquals(0, agentManager.getObjectsFiltered(MachineState.DEAD).size());
   }
 
+  @Test
+  public void testMissingSlavesArePurged() {
```
ashdza, line 120: The addition of this test causes `testDeadSlavesArePurged` to fail when run after it but I'm not clear on why

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityAgentReconciliationPoller.java
```diff
@@ -55,7 +55,7 @@
   @Override
   public void runActionOnPoll() {
     refereshSlavesAndRacks();
-    checkDeadSlaves();
+    checkInactiveSlaves();
```
ssalinas, line 58: totally missed this in earlier PRs. Can you do a slaves -> agents rename as you go through this PR please. Really thought I had caught most of those already

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityMachinesTest.java
WH77: Decent chance it's because the agent manager is shared between the two tests, so the overlapping agent ids + reconciliations cause the failures.

Could try changing the ids for this test to not overlap, or else try refactoring this test class to always use a fresh agent manager for each test.

ssalinas:
PR looks like it will fix those. There is one addition we can make to the cleanup as well. A little context first:
- AgentManager -> DEAD/MISSING... contains the active listing of mesos agents (i.e. those used and considered in scheduling/offers/resources/etc and that show up in the top sections of the SingularityUI agents page)
- InactiveAgentManager - maintains a list of agent hostnames/IPs that are marked as 'should not come back'. These are different from the list above. While AgentsManager is initially populated by data from the mesos master as we get new offers (or on startup), InactiveAgentsManager's list is populated by the user (or our automation in this case). It is targetd at catching cases where we have a 'flappy' host. For example, a known bad host may be marked as DEAD by mesos, only to come back -> ACTIVE -> die again -> repeat, causing failed tasks and churn. If we add that hostname to the inactive agents list, it is automatically marked as DECOMMISSIONED when we see it again (rather than ACTIVE)

So, I think we ca, in addition, clean up anything in InactiveAgentManager's list that is 1) being deleted by the checks you already have above or 2) older than some amount of time (likely longer than the AgentManager threshold, more like a few days). This will serve to also clean up the large Inactive Agents list you see as the very bottom table in the Singularity UI agents page

WH77 (commented):
Looks reasonable to me, assuming a host that's missing on startup and cleaned up can be safely reintroduced later on.

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularityMachinesTest.java
ssalinas: If you look into the base tests there is some `@After` annotated code that should be clearing out the in-memory zk instance used for testing after each test. The manager itself does not hold data, it just accesses it. The data is held in zk, which for tests is an in-memory curator. +1 to changing IDs to make it not overlap though still, since these bugs are often hard to track down/not worht the time if something can fix itmore easily

ashdza: I did actually try changing the ids such that they don't overlap and the failure was still occurring :\ I'll try refreshing the agent manager. Thanks Will

ashdza: The failure persists after changing the agent ids to not overlap between tests. I see the `@AfterEach` code in the base tests that clears out the leader cache and invalidates all in the zk task and deploy caches, though...will continue investigating

ssalinas: Think I know where this might be. That test that now fails sets `configuration.setDeleteDeadAgentsAfterHours(1);` in the middle of the test. The SingularityConfiguration object is not currently reset between runs of tests that I can see. So it's possible that setting is remaining in effect and messing up subsequent tests. Might be able to put either something in the AfterEach to reset it, or just a finally block in your own test that will put it back to the previous default if you are changing it

ashdza: Thanks Steve, `configuration.deleteDeadAgentsAfterHours` is in fact not being reset between tests so resetting it fixed the problem

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityAgentReconciliationPoller.java
```diff
@@ -54,16 +54,16 @@
 
   @Override
   public void runActionOnPoll() {
-    refereshSlavesAndRacks();
-    checkDeadSlaves();
+    refereshAgentsAndRacks();
+    checkInactiveAgents();
     inactiveAgentManager.cleanInactiveAgentsList(
       System.currentTimeMillis() -
       TimeUnit.HOURS.toMillis(configuration.getCleanInactiveHostListEveryHours())
```
ashdza, line 61: >clean up anything in InactiveAgentManager's list that is 1) being deleted by the checks you already have above or 2) older than some amount of time (likely longer than the AgentManager threshold, more like a few days)

@ssalinas should this be addressing point 2 or are they different mechanisms?

ssalinas: Ah, this looks like it should be doing the 'anything old than X hours' part already

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityAgentReconciliationPoller.java
ashdza: May do more of this in a separate PR; upon renaming tests, the order that the tests run changes and causes some more test failures likely due to interference between tests

PR #2241, opened 2021-11-10
Add fire alarm

pschoenfelder:
Fairly basic fire alarm capability for Singularity. 

The idea is that in an emergency, when users taking certain actions could be destructive (e.g. bouncing while agent capacity is low due to an outage) but we don't want to outright disallow then, a banner can be displayed warning users about the potential consequences of certain actions. 

This allows setting, updating, and deleting a firealarm. Can easily make it more sophisticated if necessary. 

This does not include UI updates. When it's time to implement that, the banner should probably also check the disasters resource to check what actions are currently explicitly disallowed.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/resources/DisastersResource.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/resources/DisastersResource.java
@@ -224,8 +224,7 @@ public class DisastersResource extends AbstractLeaderAwareResource {
   @Path("/firealarm")
   @Operation(summary = "Get a firealarm warning in singularity")
   public Optional<FireAlarm> getFireAlarm(
-    @Parameter(hidden = true) @Auth SingularityUser user,
-    FireAlarm fireAlarm
+    @Parameter(hidden = true) @Auth SingularityUser user
   ) {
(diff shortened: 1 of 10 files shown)


ssalinas:
Backend part looks fine. As for UI, I'm wondering if we should include this data in some version of the state json (or some other thing the UI already fetches on every page). Just trying to be careful about adding yet another api call made on every UI page

pschoenfelder:
Added to state. Should just get picked up without any kind of migration since it's an optional.

SingularityService/src/main/java/com/hubspot/singularity/resources/DisastersResource.java
```diff
@@ -218,4 +219,33 @@ private Response runFailover(SingularityUser user) {
     );
     return Response.ok().build();
   }
+
+  @GET
+  @Path("/firealarm")
+  @Operation(summary = "Set a firealarm warning in singularity")
```
rosalind210, line 225: nit: this is the same summary as the post endpoint

rosalind210:
🚢

WH77 (approved):
🚢

PR #2243, opened 2021-11-15
Clean up more zk parent nodes for stale requests

ssalinas:
Realized these weren't being cleaned up properly. Some had 40k+ children


WH77 (approved):

WH77:
🚢

PR #2244, opened 2021-11-16
Avoid Request Locks With Task Lag

WH77:
This PR prevents request locks from being acquired by lower priority possibly long running components while the request has late pending tasks, to allow higher priority components like the offer scheduler to acquire them. It does not preempt lower priority components, so lag for a request may still build up while they're holding the lock.

List of components using the request lock:
- [ ] SingularityMesosStatusUpdateHandler 
- [ ] SingularityStartup
- [ ] DeployResource
- [x] SingularityAutoScaleSpreadAllPoller
- [ ] SingularityCleaner (lb/requests/tasks)
- [x] SingularityCrashLoopChecker
- [x] SingularityExpiringUserActionPoller
- [ ] SingularityScheduler
- [ ] SingularitySchedulerPoller
- [ ] SingularityUpstreamChecker
- [x] SingularityDeployHistoryPersister
- [x] SingularityRequestHistoryPersister
- [ ] SingularityMesosOfferScheduler

Somewhat concerned about this change introducing difficult bugs, because previously lock acquisition would wait for the lock forever instead of short circuiting. Not sure if it would be better to throw an error instead of logging+returning.


SingularityService/src/main/java/com/hubspot/singularity/helpers/TaskLagGuardrail.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public boolean isLagged(String requestId) {
+    updateLateTasksByRequestId();
+    return lateTasksByRequestId.containsKey(requestId);
+  }
+
+  private void updateLateTasksByRequestId() {
+    long now = System.currentTimeMillis();
+    if (now - lastUpdate > 1000L * configuration.getRequestCacheTtl()) {
```
ssalinas, line 36: Do we need to care about request cache ttl at all here? Shouldn't this be running on the leader only?

SingularityService/src/main/java/com/hubspot/singularity/helpers/TaskLagGuardrail.java
```diff
@@ -0,0 +1,57 @@
...
+  ) {
+    this.configuration = configuration;
+    this.taskManager = taskManager;
+    this.lateTasksByRequestId = new ConcurrentHashMap<>();
+    this.lastUpdate = 0;
+  }
+
+  public boolean isLagged(String requestId) {
+    updateLateTasksByRequestId();
+    return lateTasksByRequestId.containsKey(requestId);
```
ssalinas, line 31: I'm a bit worried about this being called in a super tight loop. I know it's mostly going to no-op on the if statement. But, does it make sense to be updating this on a thread that is trying to grab a lock vs just doing it in the background? Then we can be sure every check is just a map.containsKey and never has to worry about heavier calculations on an active path

ssalinas (commented):
I'm with you on being worried about these pollers getting stuck for too long. What if each poller also kept track of how long it had been since it last ran, and if too long -> grab high priority lock instead. Too long here could be in the 10m + range for things like persisters, maybe less for some of the others

SingularityService/src/main/java/com/hubspot/singularity/helpers/TaskLagGuardrail.java
WH77: This was mostly me not wanting to add another configuration option, and thinking that request cache ttl was an appropriate value for this.

SingularityService/src/main/java/com/hubspot/singularity/helpers/TaskLagGuardrail.java
WH77: My original thinking was that a scheduled thread would be overkill for this, but you're right that this is going to be a hot method and that I should add a background thread specifically for this.

WH77:
@ssalinas - I opted to allow low priority lock requests to go through if the request lock isn't currently locked to avoid pollers getting stuck due to task lag that is unrelated to request lock usage.

Do you think that I should still force them to become high priority after enough delay, or is this change enough?

ssalinas (commented):
🚢

PR #2245, opened 2021-11-16
Request Priority API

WH77:
Adds an API for modifying request scheduling priority, similar to the request scale API, for cases where specific requests need to be prioritized/deprioritized temporarily or permanently.

Skipped email support, because this isn't intended to be a common action.

Can add UI support similar to the request scale API to this PR, if desired.


SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequestHistory.java
```diff
@@ -31,7 +31,9 @@
     DEPLOYED_TO_UNPAUSE,
     BOUNCED,
     SCALED,
-    SCALE_REVERTED
+    SCALE_REVERTED,
+    PRIORITIZED,
+    PRIORITY_REVERTED
```
ssalinas, line 36: are we able to maybe fold these into UPDATED instead somehow? these fields come through on things that a lot of other systems pull in and the roll out can be a pain with the deployer/etc having to know about all the new fields first. If the purpose is just to indicate the action more easily, we can always have the messge text reflect that, which will also show in the UI

ssalinas:
🚢

PR #2246, opened 2021-11-23
Add request-level maximum scale option 

ashdza:
This adds a maximum scale option at the request level for users to specify the maximum number of instances allowed for a particular request. Previously, there only existed a global limit in `MesosConfiguration` to prevent overscales that applied to all requests.

https://git.hubteam.com/HubSpot/PaaS-Run/issues/1546

### DIFF
--- a/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
+++ b/SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
@@ -14,13 +14,14 @@ import java.util.Optional;
 import java.util.Set;
 
 @Schema(
-  description = "Settings that apply to all tasks and deploys assocaited with this request"
+  description = "Settings that apply to all tasks and deploys associated with this request"
 )
 public class SingularityRequest {
   private final String id;
   private final RequestType requestType;
   private final Optional<List<String>> owners;
   private final Optional<Integer> numRetriesOnFailure;
+  private final Optional<Integer> maxScale;
   private final Optional<String> schedule;
   private final Optional<String> quartzSchedule;
   private final Optional<ScheduleType> scheduleType;
@@ -61,6 +62,7 @@ public class SingularityRequest {
     @JsonProperty("requestType") RequestType requestType,
     @JsonProperty("owners") Optional<List<String>> owners,
     @JsonProperty("numRetriesOnFailure") Optional<Integer> numRetriesOnFailure,
+    @JsonProperty("maxScale") Optional<Integer> maxScale,
     @JsonProperty("schedule") Optional<String> schedule,
     @JsonProperty("instances") Optional<Integer> instances,
     @JsonProperty("rackSensitive") Optional<Boolean> rackSensitive,
@@ -129,6 +131,7 @@ public class SingularityRequest {
     this.id = checkNotNull(id, "id cannot be null");
     this.owners = owners;
     this.numRetriesOnFailure = numRetriesOnFailure;
+    this.maxScale = maxScale;
     this.schedule = schedule;
     this.rackSensitive = rackSensitive;
     this.instances = instances;
@@ -183,6 +186,7 @@ public class SingularityRequest {
       .setLoadBalanced(loadBalanced)
       .setInstances(instances)
       .setNumRetriesOnFailure(numRetriesOnFailure)
+      .setMaxScale(maxScale)
       .setOwners(copyOfList(owners))
       .setRackSensitive(rackSensitive)
       .setSchedule(schedule)
@@ -241,6 +245,10 @@ public class SingularityRequest {
   )
   public Optional<Integer> getNumRetriesOnFailure() {
     return numRetriesOnFailure;
+  }
+
+  public Optional<Integer> getMaxScale() {
+    return maxScale;
   }
 
   @Schema(nullable = true, description = "A schedule in cron, RFC5545, or quartz format")
@@ -586,6 +594,7 @@ public class SingularityRequest {
       requestType == that.requestType &&
       Objects.equals(owners, that.owners) &&
       Objects.equals(numRetriesOnFailure, that.numRetriesOnFailure) &&
+      Objects.equals(maxScale, that.maxScale) &&
       Objects.equals(schedule, that.schedule) &&
       Objects.equals(quartzSchedule, that.quartzSchedule) &&
       Objects.equals(scheduleType, that.scheduleType) &&
@@ -639,6 +648,7 @@ public class SingularityRequest {
       requestType,
       owners,
       numRetriesOnFailure,
+      maxScale,
       schedule,
       quartzSchedule,
       scheduleType,
@@ -686,6 +696,8 @@ public class SingularityRequest {
       owners +
       ", numRetriesOnFailure=" +
       numRetriesOnFailure +
+      ", maxScale=" +
+      maxScale +
       ", schedule=" +
       schedule +
       ", quartzSchedule=" +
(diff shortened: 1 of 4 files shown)


ashdza:
I'm not entirely clear on how this new option would interact with autoscale

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
```diff
@@ -14,13 +14,14 @@
...
 @Schema(
-  description = "Settings that apply to all tasks and deploys assocaited with this request"
+  description = "Settings that apply to all tasks and deploys associated with this request"
 )
 public class SingularityRequest {
   private final String id;
   private final RequestType requestType;
   private final Optional<List<String>> owners;
   private final Optional<Integer> numRetriesOnFailure;
+  private final Optional<Integer> maxScale;
```
ashdza, line 24: I'm not entirely clear on how this new option would interact with autoscale

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -230,12 +230,22 @@ public SingularityRequest checkSingularityRequest(
...
-    checkBadRequest(
-      request.getInstancesSafe() <= maxInstancesPerRequest,
-      "Instances (%s) be greater than %s (maxInstancesPerRequest in mesos configuration)",
-      request.getInstancesSafe(),
-      maxInstancesPerRequest
-    );
+
+    if (request.getMaxScale().isPresent()) { // check if number of instances exceeds request-level max scale first
+      checkBadRequest(
+        request.getInstancesSafe() <= request.getMaxScale().get(),
```
rosalind210, line 236: you could simplify this to be one call if you save boolean/maxInstances
Ex:
```
int maxInstances = request.getMaxScale().isPresent() ? request.getMaxScale() : maxInstancesPerRequest;
boolean requestCountWithinBounds = request.getMaxScale().isPresent() ? request.getInstancesSafe() <= request.getMaxScale().get() : request.getInstancesSafe() <= maxInstancesPerRequest;

checkBadRequest(
   requestCountWithinBounds,
   request.getInstancesSafe(),
   maxInstances
);
```

rosalind210: or something with a conditional if you don't think ternaries are readable

SingularityBase/src/main/java/com/hubspot/singularity/SingularityRequest.java
rosalind210: it looks like AutoScale uses SingularityClient's `scaleSingularityRequest` ([code](https://git.hubteam.com/HubSpot/AutoScale/blob/master/AutoScaleData/src/main/java/com/hubspot/autoscale/data/externalservice/SingularityService.java#L410)), which hits this [resource](https://github.com/HubSpot/Singularity/blob/master/SingularityService/src/main/java/com/hubspot/singularity/resources/RequestResource.java#L1660)

rosalind210: That resource uses SingularityValidator's `checkScale`, so you should the `checkBadRequest` also to that method so that AutoScale is blocked

(or make a `checkBadScaleRequest` method for both validator methods to use)

ashdza: It looks like [`checkScale`](https://github.com/HubSpot/Singularity/blob/76c383f587c22cd1ce21b2c837046326c8b3b2d5/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java#L1183) doesn't currently check against the global max scale in `MesosConfig`, so AutoScale isn't blocked by that when scaling. Do we want it be blocked by the new request-level max scale option?

pschoenfelder: Good catch — we should check both max scale values. Since Singularity has been missing this, we should verify that no active requests would be caught by the current `maxNumInstancesPerRequest` value before this is rolled out.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -230,11 +230,17 @@ public SingularityRequest checkSingularityRequest(
       !request.getInstances().isPresent() || request.getInstances().get() > 0,
       "Instances must be greater than 0"
     );
+
+    // check if requested number of instances exceeds max scale
+    int maxInstances = request.getMaxScale().isPresent()
```
pschoenfelder, line 235: This should be the lower of the two config values (i.e. a user shouldn't be able to exceed our global max scale by specifying a request level max scale)

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -230,11 +230,17 @@ public SingularityRequest checkSingularityRequest(
...
+      ? request.getMaxScale().get()
+      : maxInstancesPerRequest;
+
     checkBadRequest(
-      request.getInstancesSafe() <= maxInstancesPerRequest,
-      "Instances (%s) be greater than %s (maxInstancesPerRequest in mesos configuration)",
+      request.getInstancesSafe() <= maxInstances,
+      "Instances (%s) cannot be greater than %s",
```
pschoenfelder, line 241: Could be helpful to users to state which configuration value they're being blocked by here

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
ashdza: To confirm, does this mean global does not automatically override request-level and vice versa, and we instead want to always use the lower of the two when validating scale?

pschoenfelder: Yes, we should always use the lowest. Users should be able to further restrict the global option with a request level option, but not the other way around.

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -230,12 +230,32 @@ public SingularityRequest checkSingularityRequest(
...
+        );
+      }
+    } else {
+      checkBadRequest(
+        request.getInstancesSafe() <= maxInstancesPerRequest,
+        "Instances (%s) cannot be greater than %s (maxInstancesPerRequest in mesos configuration)",
+        request.getInstancesSafe(),
+        maxInstancesPerRequest
+      );
+    }
```
ashdza, line 258: may have been staring at this for too long, but I couldn't figure out a cleaner way of implementing this logic such that users would be told specifically which max scale config they were violating without being hacky

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -377,6 +373,34 @@ public SingularityRequest checkSingularityRequest(
       .build();
   }
 
+  private void checkRequestAgainstMaxScale(SingularityRequest request) {
```
ashdza, line 376: I wasn't sure how to implement this logic more concisely such that users are still told specifically which max scale config they are violating, without doing anything hacky

pschoenfelder:
🚢

pschoenfelder:
🚢

PR #2247, opened 2021-12-06
Restore old constructor in singularity request parent

WH77:
Need to keep the old constructor for external code that creates request parent objects directly.

cc - @ssalinas


ssalinas (commented):
deprecated constructor doesn't need all the JsonProperty annotations, gtg besides that

ssalinas:
🚢

PR #2250, opened 2021-12-14
Pre-compile Pattern usages

PtrTeixeira:
Pre-compiles all of the Pattern uses (/ implicit Pattern uses in`String.replaceAll`) 
with static final pre-compiled `Pattern` uses.


pschoenfelder:
🚢

PR #2251, opened 2021-12-16
Guice bindings for custom authorizer implementations

pschoenfelder:
Allows overriding `getAuthModule` to inject custom implementations of `SingularityAuthorizer`


SingularityService/src/main/java/com/hubspot/singularity/auth/SingularityAuthorizer.java
```diff
@@ -91,6 +99,16 @@ protected abstract boolean isAuthorizedForRequest(
...
   );
 
+  public void checkForAuthorizedChanges(
+    SingularityRequest newRequest,
+    Optional<SingularityRequest> oldRequest,
+    SingularityUser user
+  ) {
+    if (oldRequest.isPresent()) {
+      checkForAuthorizedChanges(newRequest, oldRequest.get(), user);
+    } else {}
```
ssalinas, line 109: nit, don't need the empty else block

ssalinas:
just the one nit, good beyond that 👍

PR #2252, opened 2021-12-22
Disasters UI Improvements

WH77:
- Notifications for successful configuration override API calls.
- An explicit clear placement strategy override button.

cc - @ssalinas


ssalinas:
🚢

Issue #948, opened 2016-03-14
Integration with Mesos maintenance mode


stevenschlansker:
> On Mar 14, 2016, at 12:12 PM, Joseph Wu <joseph@mesosphere.io> wrote:
> 
> Managing maintenance is currently up to the operator (you, presumably).  If you have something to contribute (code, docs, or examples), that would be greatly appreciated :)
> 
> We haven't prioritized other integration (like the CLI or web UI) since maintenance primitives themselves need to be supported by frameworks using the V1 HTTP API; and frameworks developers have not all started/finished migrating to the new API yet.  (Without framework support, frameworks will not react to your maintenance windows.)
> 
> Here's the tracking ticket for adding maintenance info to the web UI (no work has been done on this yet):
> https://issues.apache.org/jira/browse/MESOS-2082

janisz:
Is anybody working on this? It require switch from native mesoslib to HTTP API first.

ssalinas:
Nobody is actively working on this at the moment. We have discussed the switch to using the HTTP API, but that work has not been started or been scheduled yet. Internally we use Singularity's Decommissioning and Freezing of slaves to provide maintenance windows for us.

stevenschlansker:
We're running experimental Jenkins and Spark frameworks on our Mesos / Singularity clusters.  We use Singularity decommissioning to take hosts out for maintenance.

This means that our other schedulers can still place tasks.  This is surprising and disappointing from an operator's perspective who hasn't watched the history of this issue (that it predates maintenance mode entirely).

yuriy-filatov:
Any updates here? We also need this badly.

ssalinas:
@yuriy-filatov it wasn't currently in our planned work. By 'need this badly' is the current decommission buggy or causing you issues, or is it just an inconvenience to also have to remember to set mesos maintenance mode manually? I don't have an operator api client in here yet, but if you'd like to PR this, the steps would be:

- Add a client for the operator endpoints on the mesos master (or at least just that one and we can fill out more later)
- Add a call to it on the slave decommission code path
- Add a call to undo this on the reactivate slave code path

yuriy-filatov:
@ssalinas thanks for your quick answer.
[...]

PR #2253, opened 2021-12-22
Cleanup old decommissioned agents

WH77:
Assuming that if an agent is decommissioned (not stuck decommissioning) for a long time, it should be safe to clean it up the same way as dead/missing agents.

cc - @ssalinas


ssalinas:
I don't know that this is actually safe. I think we also need the condition that we have not received offers from it in that amount of time as well. Currently, if an agent is still active with the mesos master, but decommissioned for us, we are continually receiving offers. So, if we were to delete it from our agents list -> get a new offer, we will actually end up adding it back as active. There is a maintenance mode in mesos master (old issue https://github.com/HubSpot/Singularity/issues/948) that we could integrate with to stop offers coming in when decomissioned

ssalinas:
🚢

PR #2258, opened 2022-01-13
Option to truncate too large of task Ids

rosalind210:


SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -451,6 +451,9 @@
   private int deployCacheTtlInSeconds = 5;
   private int requestCacheTtlInSeconds = 5;
 
+  private boolean truncateLargeTaskIds = false;
+  private int taskIdSize = 200;
```
ssalinas, line 455: this can just be a constant since it's part of the db migration already

ssalinas (commented):
One thing I hadn't thoguht of earlier... Does this then cause any of the things that return a List< SingularityTaskIdHistory> to fail if the task id is cut off in a bad spot 🤔 . Didn't realize until reading the PR that we don't just read it as a String most places, we read it as a SingularityTaskIdHistory

Will try to think of some other options as well

PR #2259, opened 2022-01-14
Task Persister skips too long task Ids

rosalind210:
We won't be able to save tasks whose task Ids are greater than the DB column setting, so we should skip them and delete them from ZK after some time (set to 1 week in PR).

Once we have a longterm solution for these task IDs, we shouldn't need to skip anymore.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityTaskHistoryPersister.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityTaskHistoryPersister.java
@@ -88,21 +88,40 @@ public class SingularityTaskHistoryPersister
           int forRequest = 0;
           int transferred = 0;
           for (SingularityTaskId taskId : taskIds) {
-            if (moveToHistoryOrCheckForPurge(taskId, forRequest)) {
-              LOG.debug("Transferred task {}", taskId);
-              transferred++;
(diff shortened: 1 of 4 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/history/SingularityTaskHistoryPersister.java
```diff
@@ -88,21 +88,40 @@ public void runActionOnPoll() {
...
+                taskId.getCreateTimestampForCalculatingHistoryAge() >
+                TimeUnit.DAYS.toMillis(7)
+              ) {
+                LOG.warn(
+                  "Deleting {} from ZK, could not persist in DB because of task ID length",
+                  taskId.getId()
+                );
+                purgeFromZk(taskId);
+              }
+              LOG.error("Task ID {} too long to pesist to DB, skipping", taskId.getId());
```
ssalinas, line 106: nit, can put that log line in an else so we don't get the double log line when it's actually deleted and not skipped

rosalind210: oh whoops, yes

ssalinas:
🚢

WH77:
🚢

PR #2260, opened 2022-01-21
Add endpoint and client call to retrieve scheduled tasks info

ashdza:
This PR adds an endpoint to `StateResource` and client call to `SingularityClient` that enable retrieval of the current `SingularityScheduledTasksInfo`. `SingularityScheduledTasksInfo` contains the number of late task and maximum task lag, both of which will be checked in Serenity before decommissioning agents in order to implement a slow drain and prevent decommissioning too quickly.

issue: https://git.hubteam.com/HubSpot/PaaS-Run/issues/1588
context: https://hubspot.slack.com/archives/CD13L0XCY/p1642617864005200

### DIFF
--- a/SingularityServiceIntegrationTests/src/test/java/com/hubspot/singularity/SingularityStateIT.java
+++ b/SingularityServiceIntegrationTests/src/test/java/com/hubspot/singularity/SingularityStateIT.java
@@ -24,14 +24,8 @@ public class SingularityStateIT {
   }
 
   @Test
-  public void testLateTasksEndpoint(SingularityClient singularityClient) {
-    final int lateTasks = singularityClient.getNumLateTasks();
-    // TODO: assertion
-  }
-
-  @Test
-  public void testTaskLagEndpoint(SingularityClient singularityClient) {
-    final long taskLag = singularityClient.getMaxTaskLag();
+  public void testScheduledTasksInfoEndpoint(SingularityClient singularityClient) {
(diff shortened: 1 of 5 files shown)


SingularityServiceIntegrationTests/src/test/java/com/hubspot/singularity/SingularityStateIT.java
```diff
@@ -22,4 +22,10 @@ public void testStateEndpoint(SingularityClient singularityClient) {
 
     assertEquals(3, state.getActiveAgents());
   }
+
+  @Test
+  public void testScheduledTasksInfoEndpoint(SingularityClient singularityClient) {
+    final SingularityScheduledTasksInfo scheduledTasksInfo = singularityClient.getScheduledTasksInfo();
+    // TODO: assertion
+  }
```
ssalinas, line 30: We don't actually have any real unit tests set up for the SingularityClient at the moment (yes... shame on us...). If you are trying to fit unit tests in for this PR, better bet would be to write one for the `stateManager.getScheduledTasksInfo();` method. The SingularityClient methods you'll likely end up testing in our staging environment instead

ashdza: Sounds good - I'll test the new client method in staging then and won't add the unit test here as `stateManager.getScheduledTasksInfo()` is not the functionality being added in this PR.

ssalinas (commented):
🚢

PR #2262, opened 2022-01-24
Fix npe for tasks with null hostnames from decommissioned host resources

WH77:
#2253 likely caused null hostnames / updates for tasks on decommissioned hosts, because agent resource fields could get overwritten with nulls from mesos master on startup. This PR preserves the cleanup logic, while removing the unneeded agent resource updates.

cc - @ssalinas


ssalinas:
🚢

PR #2266, opened 2022-02-08
Mutate request instances based on existing request

rosalind210:
We want to start allowing deploys with Singularity Requests that have an empty `instances` field and instead of failing to deploy, those requests will take on the instance count of the existing Singularity Request. 

The new code path behind the config value  `allowSettingRequestInstances` will check checkBadRequest for both the new request has empty instances and the existing request being empty/empty instances. If the existing request isn't empty && doesn't have an empty `instances` field and the current request has empty `instances`, then the current request is updated to have the existing request's instance count.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
@@ -230,10 +230,30 @@ public class SingularityValidator {
       request.getId().length(),
       request.getId()
     );
-    checkBadRequest(
-      !request.getInstances().isPresent() || request.getInstances().get() > 0,
-      "Instances must be greater than 0"
-    );
+
+    if (singularityConfiguration.allowSettingRequestInstances()) {
+      checkBadRequest(
+        (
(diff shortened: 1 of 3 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -230,10 +230,30 @@ public SingularityRequest checkSingularityRequest(
...
+          !request.getInstances().isPresent()
+        ) ||
+        request.getInstances().get() > 0,
+        "Instances must be greater than 0"
+      );
+
+      if (
+        existingRequest.flatMap(SingularityRequest::getInstances).isPresent() &&
+        !request.getInstances().isPresent()
+      ) {
```
ssalinas, line 247: quick thought, should this only apply to SERVICE/WORKER types? I think it is a valid operation to want to unset the instance count for an ON_DEMAND, effectively removing the instance count cap for that. With the validation as it stands we would no longer be able to do that

WH77 (approved):
🚢

ssalinas (approved):

PR #2269, opened 2022-02-22
Retry failed tasks with new deploys

WH77:
This PR should fix failed tasks not being retried due to newer deploys while the tasks were running.

The branch that skips the newer deploy short circuit should generally get to the current retry logic, but all the other checks / cooldowns still apply.

The log level changes are in case I was mistaken and that the real cause of these tasks not getting retried was obsolete requests all along.

cc - @ssalinas , @pschoenfelder

### DIFF
--- a/SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
+++ b/SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
@@ -1255,6 +1255,94 @@ public class SingularitySchedulerTest extends SingularitySchedulerTestBase {
     Assertions.assertTrue(taskManager.getActiveTaskIds().isEmpty());
   }
 
+  @Test
+  public void testRetriesWithNewDeploys() {
+    SingularityRequestBuilder bldr = new SingularityRequestBuilder(
+      requestId,
+      RequestType.RUN_ONCE
+    );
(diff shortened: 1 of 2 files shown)


ssalinas:
Could we wire through a unit test for this one? Curious if the pending task will 1) come all the way through and 2) have the old or the new deploy attached to it and 3) if the old/inactive deploy, will it actually launch the task correctly?

SingularityService/src/test/java/com/hubspot/singularity/scheduler/SingularitySchedulerTest.java
```diff
@@ -1255,6 +1255,98 @@ public void testRetriesWithOverrides() {
...
+    );
+
+    statusUpdate(taskManager.getActiveTasks().get(0), TaskState.TASK_LOST);
+
+    scheduler.drainPendingQueue();
+    deployChecker.checkDeploys();
+    scheduler.drainPendingQueue();
+    resourceOffers();
+
+    // TODO - new deploys have new deploy statistics, so we lose track of the # of retries
```
WH77, line 1345: Tasks could exceed numRetries if they deploy often enough, but cooldowns should keep the retries from becoming a problem.

Debated adding a new `retryIgnoreDeploy` field to requests or deploys, but figured that was overkill.

ssalinas (commented):
🚢

PR #2270, opened 2022-02-22
Run persisters in parallel

ssalinas:
This allows parallel execution of persister tasks within a type of persister (e.g. 5 parallel executions for the tsak history persister). The number of these is configurable as `historyPollerConcurrency`

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
@@ -261,6 +261,8 @@ public class SingularityConfiguration extends Configuration {
 
   private long persistHistoryEverySeconds = TimeUnit.MINUTES.toSeconds(2);
 
+  private int historyPollerConcurrency = 5;
+
   private int maxPendingImmediatePersists = 200;
 
   private long reconcileAgentsEveryMinutes = TimeUnit.HOURS.toMinutes(1);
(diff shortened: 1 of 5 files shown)


SingularityService/src/main/java/com/hubspot/singularity/config/SingularityConfiguration.java
```diff
@@ -261,6 +261,8 @@
 
   private long persistHistoryEverySeconds = TimeUnit.MINUTES.toSeconds(2);
 
+  private int historyPollerConcurrency = 5;
```
rosalind210, line 264: if each history persister is using the same executor, could one end up backing up the others if we only allow 5 concurrent items to be saved at a time?

ssalinas: No, each runs behind the persister lock already

rosalind210:
🚢

tcampbell:
Hey @rosalind210, it looks like you _Commented_ instead of _Approved_ this PR. No big deal here, but in the future, please ensure that your :thumbsup: is reflected as an approval.

ssalinas:
@tcampbell the approval flow as far as I knew doesn't work for the open source projects. In the past they were only hooked up for a direct comment/ 🚢 . If that has changed we can change our workflow

tcampbell:
You're right, my bad.  sorry about that.  Please ignore.

PR #2274, opened 2022-03-17
Address missing task data

ssalinas:
We are seeing some NPEs come through here. Adding more logging to see what is missing on the task definition here

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
@@ -1394,10 +1394,9 @@ public class TaskManager extends CuratorAsyncManager {
       }
     }
 
+    AtomicBoolean hasErr = new AtomicBoolean(false);
     try {
       final String path = getTaskPath(task.getTaskId());
-
-      AtomicBoolean hasErr = new AtomicBoolean(false);
       curator
         .transaction()
         .forOperations(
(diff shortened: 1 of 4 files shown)


WH77 (approved):
🚢

ssalinas:
update here, found some additional pieces. Based on the logs and the fact that a regular task bounce (which uses getTask vs getTasks) succeeds. We can say the task data is in the zkCache, but not in zk itself. So, as a bandaid, this attempts to repair the task if we see that case, before continuing on to the methods that would throw exceptions. The overall flow should remain the same if the task is still not present

SingularityService/src/main/java/com/hubspot/singularity/data/TaskManager.java
```diff
@@ -1379,24 +1397,35 @@ private void createTaskAndDeletePendingTaskPrivate(SingularityTask task)
...
+              LOG.error("Error committing new task {} to zk {}", task.getTaskId(), r);
+              hasErr.set(true);
+            }
+          }
+        );
 
       // Not checking isActive here, already called within offer check flow
       leaderCache.putActiveTask(task.getTaskId());
       taskCache.set(path, task);
+      if (hasErr.get()) {
```
WH77, line 1425: should this be moved to a finally block, in case the transaction throws an exception?

SingularityService/src/main/java/com/hubspot/singularity/scheduler/SingularityDeployChecker.java
```diff
@@ -1495,9 +1495,33 @@ private SingularityDeployResult enqueueAndProcessLbRequest(
...
+      for (SingularityTaskId taskId : deployActiveTasks) {
+        if (!tasks.containsKey(taskId)) {
+          Optional<SingularityTask> maybeRepaired = taskManager.tryRepairTask(taskId);
+          if (maybeRepaired.isPresent()) {
+            LOG.info("Repaired and fetched definition for {}", taskId);
+            tasks.put(taskId, maybeRepaired.get());
+          }
+        }
+      }
+      for (SingularityTaskId taskId : toShutDown) {
```
WH77, line 1515: nit - duplicate for loop could be converted to a helper method

WH77 (approved):
🚢

WH77 (approved):
🚢

PR #2275, opened 2022-03-25
Prevent accidental large scale downs

pschoenfelder:
Definitely open to different logic, thresholds, etc. Just wanted to get something out here to start a discussion around.

### DIFF
--- a/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
+++ b/SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
@@ -1239,6 +1239,22 @@ public class SingularityValidator {
         );
       }
     }
+
+    if (previousScale.isPresent() && !request.getLargeScaleDownAcknowledged()) {
+      boolean scaleDownExceedsAbsoluteMax =
+        previousScale.get() - request.getInstancesSafe() > 10;
+      boolean scaleDownExceedsRelativeMax =
+        request.getInstancesSafe() < (previousScale.get() / 2);
+      checkBadRequest(
+        !scaleDownExceedsAbsoluteMax,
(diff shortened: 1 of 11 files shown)


SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -1239,6 +1239,22 @@ public void checkScale(SingularityRequest request, Optional<Integer> previousSca
...
+        request.getInstancesSafe() < (previousScale.get() / 2);
+      checkBadRequest(
+        !scaleDownExceedsAbsoluteMax,
+        "Cannot scale down by more than %s instances at a time",
+        10
+      );
+      checkBadRequest(
+        !(previousScale.get() > 10 && scaleDownExceedsRelativeMax),
+        "Cannot scale down by more than half of current instances."
+      );
```
ssalinas, line 1256: Can these err messages have '...without explicit acknowledgement' or something like that on them? Make it obvious what field has to be set

SingularityService/src/main/java/com/hubspot/singularity/data/SingularityValidator.java
```diff
@@ -1239,6 +1239,22 @@ public void checkScale(SingularityRequest request, Optional<Integer> previousSca
         );
       }
     }
+
+    if (previousScale.isPresent() && !request.getLargeScaleDownAcknowledged()) {
+      boolean scaleDownExceedsAbsoluteMax =
+        previousScale.get() - request.getInstancesSafe() > 10;
```
ssalinas, line 1245: Should we make the 10 configurable? Think that's my only other comment on this one. Easier to change yaml and redeploy than to rebuild

PR #2277, opened 2022-04-04
Add optional costs plugin and fix overwhelming err message size

ssalinas:
- Adds an optional table under task history in the case a SINGUALRITY_COSTS_API_URL_FORMAT is supplied. At the moment this is really for internal use but we don't have an internal-only UI build. So, I'm going to make this a configurable thing and it will be off by default
- Fix the fact that large errors can cause the Messenger pop up to take over the screen. Limit the err window and scroll for overflow


pschoenfelder:
:ship:

PR #2279, opened 2022-04-04
Better loading state for costs table

ssalinas:
Make sure we don't display the loading indicator when we don't need to. Also hide the section more completely if not configured to use it


pschoenfelder:
🚢

PR #2281, opened 2022-04-05
Various bug fixes and stability improvements

ssalinas:
- An LB delete that gets back `Invalid request...` will never be able to complete. Junk it and move on
- Clean up a number of deprecated things in CuratorAsyncManager and prefer more lambdas over old style
- In the load balanced curator, attempt to use the same curator client for the same thread where possible to avoid race conditions hitting possibly different zk instances
- For tasks that do not have any active task data when they hit status updates, clean them up. These will only cause issues if allowed to persist (and will never actually be added to LB)
- Fix a possible overflow in length for task usage column with task ids oover 200 chars. This fails gracefully and is fine to leave empty
- Rerun the scheduler check after marking as started during startup. this should fix a case where a status update came in but we did not schedule a new pending task for it yet because we were not marked as started


pschoenfelder:
🚢

PR #2282, opened 2022-04-06
Add large scale down flag to client

pschoenfelder:
Flag was added to endpoint [here](https://github.com/HubSpot/Singularity/pull/2275/files#diff-05dc8661deba9fc538b4bf1e7cc39260843cf5a0ca98070ef7cc5319b898000cR1785), but not to client.


rosalind210:
🚢

PR #2284, opened 2022-04-07
Handle tasks which no longer have data in zk

ssalinas:
Follow up to #2281 

Separate handling for tasks with missing data, before the isDuplicateOrIgnorable check. We canot actually recover these, so kill them before they become an issue. Creating a cleanup for them doesn't actually do anything because the cleaner short circuits on missing task data


pschoenfelder:
🚢

pschoenfelder:
🚢

PR #2285, opened 2022-04-08
Task history handling when missing data

ssalinas:
Two main things here:
- The history persister can race with task launches with selecting the list of active tasks to exclude from persisting
- We should more thoroughly clean up tasks that don't have task data in zk


pschoenfelder:
🚢

PR #2288, opened 2022-04-18
Add optional emissions data to costs table

ssalinas:
Basically what hte title says. One additional field in the table


WH77 (approved):
🚢

PR #2291, opened 2022-04-22
POC - queue-based task launches

ssalinas:
Trying to determine if this would be a good performance boost overal.

General Situation:
- As the number of offers and scheduled tasks gets larger, the offer scoring loop gets slower and slower. Generally not a great thing that it scales in complexity as {num pending tasks} x {num offers}. So, when we get behind, it puts us further behind
- Much time in the offer loop is actually spent waiting on locks, during which not much actual work is done
- A lot of the offer filtering/etc is done on each run, but could probably be kept somewhere else as a 'live' view of current available offers

Design thoughts:
- Offers flow directly into a revamped version of the offer cache
  - Any filtering or joining of offers with their associated agent usages can be done here
  - TODO - how to handle case where offer cache is turned off (shared cluster environment)
- Pending tasks is a PriorityBlockingQueue
  - A primary queue is used with more + lighter threads that simply wait on locks
  - Once locks are acquierd things are put on the actual pending task queue, which can assume it is behind a lock and do only meaningful work in a tight loop with all offers/agents/usages/etc cached
  - Tasks are launched immediately once a match is made instead of having to wait for the rest of the poll cycle to complete
 
Upsides:
- More responsive task launch timing
- Complexity scales as {num offers} within the tight loop since only a single pending task is being worked on at once

Downsides/Issues:
- Is a preliminary queue to wait on locks actually sustainable/a bad idea?
- Tricky to handle offers than may expire if we are trying to look at a consistent 'live' view of offers and have minimal locking for usage in a tight loop
- Also need to keep zk list of pending tasks in sync with in-memory list. Similar to leader cache I guess here, but split between upcoming pending tasks and 'due now' pending tasks


ssalinas:
Good amount of progress here. Over half of tests passing. Some of which just need to be changed to fit the new scheduler style. Big open TODOs:
- [ ] Behavior of offer cache in shared env (e.g. normally would have been with offer cache disabled). Maybe this is just a default shorter TTL
- [ ] Is there a way we could batch launches still without incurring too much overhead? e.g. use one offer for two tasks to improve the number of mesos master calls made
- [ ] Should offer matching be allowed to run concurrently or should it just be a single thread in a loop to make offer checkout/locking simpler
- [ ] Metrics to inspect state of the queue

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityOfferScoring.java
```diff
@@ -0,0 +1,288 @@
...
+import com.hubspot.singularity.data.usage.UsageManager;
+import com.hubspot.singularity.helpers.MesosUtils;
+import com.hubspot.singularity.mesos.SingularityAgentUsageWithCalculatedScores.MaxProbableUsage;
+import java.util.Map;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/*
+ * It is assumed that all code in this class is run behind a request-level lock
```
WH77, line 28: Is there a "checkRequestLockOwner" or similar method that could be used by this class to make violations of this assumption more explicit / easier to debug?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityPendingTaskQueueProcessor.java
```diff
@@ -0,0 +1,419 @@
...
+
+  public void start() {
+    if (lockFuture != null) {
+      lockFuture.cancel(true);
+    }
+    running.set(true);
+    lockFuture = queuePollExecutor.submit(this::runLockWait);
+  }
+
+  public void addPendingTaskIfNotInQueue(SingularityPendingTaskId pendingTask) {
```
WH77, line 142: Does the scheduler need to support deleting/tombstoning tasks, like if a new deploy overwrites a pending deploy?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityPendingTaskQueueProcessor.java
```diff
@@ -0,0 +1,419 @@
...
+    synchronized (handled) {
+      if (!handled.contains(pendingTask)) {
+        preLockQueue.put(pendingTask);
+        handled.add(pendingTask);
```
WH77, line 146: What does handled represent - the set of tasks that are currently being handled by this processor? If so, I think `processing` would be a more accurate name.

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityPendingTaskQueueProcessor.java
```diff
@@ -0,0 +1,419 @@
...
+      }
+    }
+  }
+
+  public Set<SingularityPendingTaskId> getHandledTasks() {
+    synchronized (handled) {
+      return new HashSet<>(handled);
```
WH77, line 153: Why use a HashSet with manual synchronization instead of a ConcurrentHashMap with stub values or ConcurrentSkipListSet?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityPendingTaskQueueProcessor.java
```diff
@@ -0,0 +1,419 @@
...
+          LOG.error("Could not find pending task for id {}", pendingTaskId);
+          return;
+        }
+        SingularityPendingTask toLaunch = maybePendingTask.get();
+        SingularityTaskRequestHolder taskRequestHolder = new SingularityTaskRequestHolder(
+          taskRequestManager.getTaskRequest(toLaunch),
+          defaultResources,
+          defaultCustomExecutorResources
+        );
+        List<SingularityOfferHolder> offers = offerManager.checkoutOffers();
```
WH77, line 277: Didn't see the offer lock in this class - is the checkout safe regardless?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityPendingTaskQueueProcessor.java
```diff
@@ -0,0 +1,419 @@
...
+          unused.stream().map(Offer::getId).forEach(offerManager::returnOffer);
+
+          synchronized (handled) {
+            handled.remove(pendingTaskId);
+          }
+        } else {
+          // Back on the queue and try again. Temporarily free up lock for other pollers first
+          preLockQueue.put(toLaunch.getPendingTaskId());
+        }
+        offers.remove(bestOffer);
```
WH77, line 309: Is it possible to match a big enough offer to multiple tasks? If so, is there a way to safely subtract resources from the offer and return it?

SingularityService/src/main/java/com/hubspot/singularity/mesos/SingularityPendingTaskQueueProcessor.java
```diff
@@ -0,0 +1,419 @@
...
+        lock.lock(pendingTaskId.getRequestId(), LOCK_NAME)
+      );
+      LOG.trace(
+        "Got request lock to launch pending task {} after {}",
+        pendingTaskId.getId(),
+        System.currentTimeMillis() - start
+      );
+      // Run this on the single thread responsible for offer matching and launches
+      CompletableFuture
+        .runAsync(() -> matchAndLaunch(start, pendingTaskId), taskLaunchExecutor)
```
WH77, line 244: Somewhat concerned about the performance of only matching/launching a single task at a time - not sure if it'd be worth modifying matchAndLaunch to pull all safely locked tasks for matching instead of explicitly passing task ids 1 at a time.

WH77 (commented):
Still trying to wrap my head around the internals, but the general idea seems reasonable enough and worth trying.

My understanding, please correct me if I'm wrong:
mesos offers -> offer cache
pending tasks -> locking queue -> locking threads -> launch thread + offer cache -> launched tasks

